PR #4, opened 2015-02-14
Eclipsearduino fixes

mumme74:
These fixes makes it possible to use EclipseArduino while working on the code.
As far as I can see it has no negative effects on ordinary Arduino IDE

### DIFF
--- a/comms.h
+++ b/comms.h
@@ -1,3 +1,6 @@
+#ifndef COMMS_H
+#define COMMS_H
+
 #define vePage    1
 #define ignPage   2
 #define afrPage   3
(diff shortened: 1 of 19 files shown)


noisymime:
Will test simetime over the next few days. Thanks!

noisymime:
Possibly a silly question, but what's the purpose of the static casts to int within the trigger function of speeduino.ino? Can they be removed if curGap is made an unsigned int?

mumme74:
Jepp they can go in that case, I didnt dare to change the datatype if there would be some case were values would be negative
Therefore I typecasted to silence a compiler warning.

noisymime:
Updated to remove the cast and merged.

Thanks for the patch!

PR #5, opened 2015-02-18
Sequential

mumme74:
I have been working on making all 4 channels working and to fire sequentialy
This pullrequest implements sequential injection and coil firing (coil on plug).

It works by first start in semisequential mode, mening 1-4 and 2-3 firing simultaneously so it squirts 1 time each crank revolution. Sparks is also firing 1-4 and 2-3 as a wasted spark ignition would normally do.

When engine has been running for a sec and idling is stable it cancels injection on cyl1 and looks for a missing powerstroke, if cyl misfires we are on the first revolution 0-360deg else we are on second revolution 360-720 deg. It does not (in its current form) use the camsensor input.

I have tested it this afternoon on a 4cyl engine at a teststand, although not tested it extensively.

What's left to do is to make sequential selectable via tunerstudio instead of a compilation setting.
It would also be nice to make it possible to configure the crankangles of each cyl from tunerstudio. 
ie. 4cyl could also be a boxer or a V4 or simply have another firing order. a V-twin also has some nonstard crankangles were each cyl is at TDC.


speeduino.ino
```diff
@@ -199,49 +261,49 @@ void setup()
...
+  channels.setCylCrankAngles(configPage1.nCylinders, angles);
 }
 
 void loop() 
   {
-      mainLoopCount++;    
-      //Check for any requets from serial. Serial operations are checked under 2 scenarios:
-      // 1) Every 64 loops (64 Is more than fast enough for TunerStudio). This function is equivalent to ((loopCount % 64) == 1) but is considerably faster due to not using the mod or division operations
-      // 2) If the amount of data in the serial buffer is greater than a set threhold (See globals.h). This is to avoid serial buffer overflow when large amounts of data is being sent
-      if ( ((mainLoopCount & 63) == 1) or (Serial.available() > SERIAL_BUFFER_THRESHOLD) )
```
noisymime: What is the intent of removing this line?

The serial buffer check is there to ensure that, if the serial receive buffer is starting to fill up, that the command() function will be called, even if it's earlier than the normal 64 loops. WIthout this check, I was seeing corruption of serial data when large blocks of data (ie whole tables) were being sent by TS all at once and were overflowing the serial buffer before the 64 loops were up.

noisymime:
Wow, this is terrific! Thanks for submitting. It's going to take me some time to go through this as there's a lot of changes here. How did it work out in hardware testing?

I added one comment on the change to the serial check as I am a little worried it might bring back a potential serial corruption issue that I'd seen, but other than that, I'll have to go through this in depth.

Thanks again!

speeduino.ino
mumme74: It not removed just restructered so that it makes one less call to Serial.available() in the case when Serial.available() > SERIAL_BUFFER_THRESHOLD

It is replaced with
```C++
     mainLoopCount++;
    //Check for any requets from serial. Serial operations are checked under 2 scenarios:
    // 1) Every 64 loops (64 Is more than fast enough for TunerStudio). This function is equivalent to ((loopCount % 64) == 1) but is considerably faster due to not using the mod or division operations
    // 2) If the amount of data in the serial buffer is greater than a set threhold (See globals.h). This is to avoid serial buffer overflow when large amounts of data is being sent
    if ( (((mainLoopCount & 63) == 1) and (Serial.available() > 0)) or
         (Serial.available() > SERIAL_BUFFER_THRESHOLD) )
    {
      command();
    }
```

It basically has 2 paths
1 mainLoopCount is 63 AND Serial.available() > 0 -> command()
2 (Serial.available() > SERIAL_BUFFER_THRESHOLD) -> command()

Before the change it path 2 was
2 (Serial.available() > SERIAL_BUFFER_THRESHOLD) -> (Serial.available() > 0) -> command()

mumme74:
>Wow, this is terrific! Thanks for submitting. It's going to take me some time to go through this as there's a lot of changes here. How did it work out in hardware testing?

No problem about the time, take your time. 
I spent more or less a day debugging and trying to make sense of whats happening with engine testing.
Turned out I have false sync triggering from CKP sensor, it was worse when low crank speeds it also seemed to be more frequent when engine was hotter.

Anyway it pushed a update which addresses some of this and some other bugs.

Tobbera:
This is a rather genial way of getting CAM sync. Did this ever make it into the Speeduino code?

Tobbera:
@ mumme74  did you manage to get this to work?

mumme74:
Sorry for the late response.

No I never got it to work reliably, although I later found out that was having bad cabling to my CKP sensor so it might have pickup up interference, messing with this algorithm.
This code has been dormant for several years.

noisymime:
Closing as this is all now largely in place.

PR #11, opened 2015-05-30
Updated for Cooling Fan controls

MoysieWRX:
Updated for Cooling Fan controls


noisymime:
I think there's something weird going on with the filename in this pull. I'm going to close this one now and make the change separately.

PR #15, opened 2015-05-30
Updated for Cooling Fan controls

MoysieWRX:
Updated for Cooling Fan controls

### DIFF
--- a/speeduino.ino
+++ b/speeduino.ino
@@ -81,6 +81,8 @@ unsigned long scheduleEnd;
 
 byte coilHIGH = HIGH;
 byte coilLOW = LOW;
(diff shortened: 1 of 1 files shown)


noisymime:
Pulling this one in (Thanks!), but I'll remove the mainloop check session as there is a nicer way to do this. Checkout the timer.ino file, it has sections in it for automatically doing things every x ms, including every 1000ms (or 1 second).

MoysieWRX:
Cool, i guess if someones processor is running a little slow too it may interfere with the operation of the fan.

noisymime:
Yeah, it's probably not a huge thing, but (in case you hadn't noticed) I'm a bit paranoid about anything that can possibly affect the main loop speed :)

PR #17, opened 2015-08-21
second O2 addition

Autohome2:
modified files include also a modified tuner studio ini , these alterations have been tested on stimulator board.
also added are variable and board definitions for v0.4 board as per the pcb layout and created new board layout entry for tuner studio for dazv6 .

darren


noisymime:
Hey Darren, thanks for this, it looks great!

Is it possible for you to alter your repo layout so that the files aren't all under a 'speeduino' subdirectory?

noisymime:
Committed as part of pull #19

PR #19, opened 2015-09-13
added second o2 support

Autohome2:
also added new board layout for a modifiied v04 layout and other extra pins


noisymime:
This looks nice and neat to me, well done!

Autohome2:
What's the best thing to do for the new vvt support?

PR #22, opened 2015-10-10
Basic Command Line Interpreter

ConnerMcLaughlin:
This enables users to access values from the Serial monitor much as the way Tuner Studio does.  It should not conflict with existing commands TS already uses.  By going into the Serial monitor a user should be able to type 'L' (case sensitive) and see the VE table.  Typing '?' will list all other commands.

### DIFF
--- a/comms.ino
+++ b/comms.ino
@@ -27,13 +27,24 @@ void command()
 
       case 'C': // test communications. This is used by Tunerstudio to see whether there is an ECU on a given serial port
         testComm();
+        break;     
+        
+      case 'L':
+        sendPage(true);
         break;
-
+                
+      case 'N':
+        Serial.println();
+        break;
+        
       case 'P': // set the current page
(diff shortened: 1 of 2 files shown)


comms.ino
```diff
@@ -588,7 +797,7 @@ void sendToothLog(bool useChar)
           Serial.write(lowByte(tempToothHistory[x]));
         }
       }
-      //Serial.flush();
+      Serial.flush();
```
noisymime: Any reason this has been reenabled? Forcing the serial flush can cause some large latency problems is all.

ConnerMcLaughlin:
Forgot to re-comment it back, sorry.  If you want we could meet up in the IRC today.

noisymime:
Yeah, happy to catch up on IRC. I've pushed a number of changes to this pull request into branch pr/22

PR #39, opened 2016-09-11
CLi ascii boost and vvt view

ConnerMcLaughlin:
Also with adjustments to the axisX and axisY in ascii view for all table views


ConnerMcLaughlin:
It got confusing when the previous files had all negs and the changed one had pos.  So might have to close this request and do again.  What a pain.  Let me know before you make an other commit.

PR #41, opened 2016-09-14
Fix for TS rom burn break

ConnerMcLaughlin:
This is a fix that's been tested for the issue with the boost and vvt tables not burning to the rom correctly


noisymime:
Has already been fixed in master. Will be changed anyway to be based on the table size

ConnerMcLaughlin:
thank you

PR #43, opened 2016-09-25
Added support for both NO and NC clutch switch on launch

ConnerMcLaughlin:
In many car designs a NC clutch switch acts as NO switch because the pressure plate is forcing the switch open when clutch is released and vise versa for a NO clutch switch.  But in almost all designs I've seen they are always grounding.  A hot switching option may be added later for exceptions to that design standard.


ConnerMcLaughlin:
Gave the user the option to use Pullup Resistor or not.  Also left room for David's (PSIG) pulldown resistor trick but without implementing it.

reference/speeduino.ini
```diff
@@ -380,7 +380,9 @@ page = 6
...
         boostKD      = scalar, U08,     59,             "%",        1.0,   0.0,  0.0,  200.0,      0 ; * (  1 byte)
-        unused6-60   = scalar, U08,      60,        "RPM",       100.0,       0.0,   100,     25500,    0
+	
+	lnchPullRes  =   bits, U08,      60         [0:1],      "Float"  , "Pullup", "INVALID", "INVALID"
+        unused6-60   =   bits, U08,      60,        [2:7],      "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID"
```
noisymime: Super small detail, but you can't have a bits entry with all INVALID options. Just make the first one a value of 'ONE' or something. TS chucks an error message when loading the project otherwise.

PR #44, opened 2016-09-27
Adds a selectable output stream to serial3 , for Can interface Use

Autohome2:
Option to enable the stream is in accessories within tuner studio.
Serial3 then responds to a small subset of commands specifically for use with the Can interface
The Can interface does all the packing and timing required for the CANBUS without loading the Speeduino processor or needing an spi pinset
The Can interface Firmware code will be available from https://github.com/Autohome2/Speeduino-can-interface


Autohome2:
revised version submitted

PR #50, opened 2016-10-18
Add some functions to speedup a little bit

VitorBoss:
Fix some compilation erros, variable name and remove unused variables from fastMap1023toX functin


noisymime:
OK firstly thanks a heap for submitting the pull request! It's great to have new contributors to the project.

There's a LOT going on in this pull and it really needs to be split out into separate requests, one for each change.

The float version of the table lookup in particular is troublesome. It cannot run on an AVR and I would suggest it's even questionable on other ARM based units that may not have an FPU (Eg Teensy 3.2). One issue is that having even a single declaration of a float in the code means that arduino IDE will pull in the whole float library, which is about 5kb. I'm not sure how much advantage there is in a floating point version of this either, though I could possibly be convinced if there is a real improvement somewhere. 

Have you done any testing to note the performance difference of using the DigitalWriteFast?

jakemoroni:
I just want to chime in and say that 'float' is supported by avr-gcc (and therefore Arduino), however, it's implemented (emulated) in software for target CPUs that do not have an FPU (which is most, if not all, of the AVR MCUs). I'm sure everyone here knows this, but I figured I should point it out for others who may be less familiar with programming in this kind of environment.

The end result is that seemingly simple operations (i.e., 1 + 2) can become very expensive when performed using floating point types. Tens of instructions will be generated for operations that could have otherwise been handled in a couple of instructions if fixed width types were used.

As for pulling in the entire float library as a result of a single 'float' declaration, I have not been able to reproduce this, and I don't understand why this would be happening. Floating point operations are handled by GCC transparently. It may generate calls to a libgcc function or something, but I wouldn't expect the final executable to contain code for functions which are unused and unreachable. I'm not 100% on this though.

Anyway, it's generally recommended to avoid floating point types in embedded environments because they're usually very expensive and less deterministic in terms of timing. However, I've personally made heavy use of them in my methanol injection controller project (which uses an Atmega328p). It really depends on the application and how complex the math is. Sometimes it becomes difficult to keep track of precision and overflows when implementing complicated math routines using fixed types. If you can't be 100% positive that your fixed types aren't going to overflow or lose unnecessary precision, then you should use floating point types. However, I guess one could argue that you should be 100% confident in your code before deploying it...

Sorry for the rant. Love the project, btw.

VitorBoss:
Noisymine I know the float operation isn't native and use an library, I'm new on Arduino world but testing the long and float version on a very simple PIC18F2550 witch is much less powerful than Mega2560 result on a faster or same speed of results. The writefast version I couldn't test yet, but reading about its results let me excited about it. It is easy to remove it or partially implement it, need only a few changes on digitalwritefast definition file. Unfortunately I do not have an oscilloscope here for test all.

VitorBoss:
I have realized a benchmark on standard and float version, this is the results:
Stadard code lookup 430760, Fload code lookup 1168884, Number of loops 5000

Resuts using micros() function

VitorBoss:
I have realized a benchmark on digitalWrite and digitalWriteFast version, this is the results:
Stadard digitalWrite 1588424 digitalWriteFast version 1588348  Number of loops 250000

Resuts using micros() function. With this results is better to ignore it.
All related to those functions should be removed.

noisymime:
Even I'm surprised that the difference on the digitalWriteFast was that small. 

You've probably seen in the code where some of the time critical digital operations are done at the lower level, which does make a worthwhile difference. I need to sit down and finish that off work just to make this a non-issue. 

Thanks so much for your testing though! Is there anything else from this PR that you want me to look at?

VitorBoss:
No, but I started here because it's potencial, I want to know your opinion abut this: http://speeduino.com/forum/viewtopic.php?f=17&t=800

I have started the thread, and have a lot of options there.

VitorBoss:
Today I have test the WriteFast again, using this code:
` #define pinNum 2
  #define nTests 25000  
    unsigned long InitTime = micros();
    for(int x = 0; x<nTests; x++)
    {
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
      digitalWrite(pinNum, HIGH);
      digitalWrite(pinNum, LOW);
    }
  FinishTime = micros() - InitTime - nTests;
  Serial.print(FinishTime);
`
Same code loop for both versions
The results was very exciting:
Stadard digitalWrite 5167960 digitalWriteFast version 88436  Number of loops 25000

VitorBoss:
I have reworked on code and found some things about digitalwritefast way of work. So I have put a limiter on Baro correction to don't screw up if the engine is already running and occurs a power cycle. Other minor changes on math declarations.

noisymime:
Branch pr/50 has been created for this as there are changes required before merger into master

PR #60, opened 2016-11-27
Change int to short

tausday:
Hello,

i have change every int to short, because teensy use int as 32bit and mega as 16bit.
So the Storage are mismatched/moved to other variable like trigger wheel and something else
http://speeduino.com/forum/viewtopic.php?f=11&t=889

On Mega is short and int the same
https://www.arduino.cc/en/Reference/Short
https://www.arduino.cc/en/Reference/Int


jakemoroni:
How come stdint isn't used? Then you can use types that aren't ambiguous (uint32_t, etc...). There are also "fast" types like "uint_fast32_t" which guarantee a minimum number of bits and may or may not use more depending on what's fastest for the target CPU. This seems like the most portable way to write code for various MCUs.

noisymime:
When the project was first started, one of the primary goals was to make this as simple for new developers as possible, including through use of as much of the Arduino syntax as was feasible.

This priority has fallen by the wayside a bit though and with the support of Teensy coming in I agree that it's reasonable to change the type declarations where needed. Some care will need to be taken however as there are places when exact sized types are required and use of a type that has an unknown number of bits (Eg the fast types that have a minimum rather than fixed number of bits) could lead to issues. This is primarily in the schedulers that rely on 16 bit values to match hardware timers. 

Whilst personally I prefer the use of the typedefs (Eg int, char, short etc) I agree that the proper way to do this is with the standard types (Eg int16_t, uint16_t) but would urge caution around the use of the fast types.

tausday:
don't merge it i don't know why but tunerstudio online works some seconds and then it will stop to go offline

noisymime:
Closing this as critical types have now been moved to the std definitions. Longer term everything will change to these, but for now it's suitable for running on other platforms

PR #63, opened 2016-12-27
Global injector close angle

ConnerMcLaughlin:
This makes it easier to set the injector close angle with one setting instead of four.


ConnerMcLaughlin:
Pull me code

ConnerMcLaughlin:
Maybe I don't know enough about github but seeing commits after a PR seems like the code isn't going to be pulled.  I was hoping to get comments at least.

noisymime:
Just a heads up, I will look to merge this early in Jan after the Dec release goes out. This will give me a little more time to test that there's no issues etc.

ConnerMcLaughlin:
Don't stress too much over it.  In the firmware code(not TS) there is literally two lines of code!  But of course it's a deviation from the feature requests.  I always value your feedback.  Go Speeduinees!

ConnerMcLaughlin:
January Bump

PR #68, opened 2017-02-15
Master

kierenblack:
added some features with TPS min and max values and also a fan and fuel pump indicator


noisymime:
Closing as stale. Most of these changes have now been included via other commits anyway

PR #73, opened 2017-03-10
hardware Test page

Autohome2:
This adds a hardware test page(enabled in project properties).
this allows direct testing of the outputs of the injector and spark channels when the engine is off


noisymime:
Thanks for the work!

I've pulled this into branch pr/73 as there are a few things I'd like to change on this before merging.

PR #77, opened 2017-04-08
8ch remote input over serial3

Autohome2:
Adds remote analog input over serial3 (from Canbus interface or direct from sensor device(s))
The data is displayed in Tunerstudio via gauges and logged to tunerstudio log files.
Speeduino polls for the data in the same way it does for local analog inputs in a periodic manner when activated from TS.
Although Speeduino currently doesnt use this external data , it is available when the main firmware allows it to be selected.
The protocol roughly follows j1939 methods .
A test program is available from https://github.com/Autohome2/Speeduino-can-interface in the examples folder  progam titled "simpleinput V1" . This demonstrates the basic transmission protocol. A PDF detiling the protocol further is also on that git page.


reference/speeduino.ini
```diff
@@ -524,47 +524,32 @@ page = 9
...
-        unused10_32 = scalar, U08,     32,        "",       1, 0, 0, 255, 0
-        unused10_33 = scalar, U08,     33,        "",       1, 0, 0, 255, 0
-        unused10_34 = scalar, U08,     34,        "",       1, 0, 0, 255, 0
-        unused10_35 = scalar, U08,     35,        "",       1, 0, 0, 255, 0
-        unused10_36 = scalar, U08,     36,        "",       1, 0, 0, 255, 0
-        unused10_37 = scalar, U08,     37,        "",       1, 0, 0, 255, 0
-        unused10_38 = scalar, U08,     38,        "",       1, 0, 0, 255, 0
-        unused10_39 = scalar, U08,     39,        "",       1, 0, 0, 255, 0
-        unused10_40 = scalar, U08,     40,        "",       1, 0, 0, 255, 0
+        enable_candata_in = bits,   U08,     0, [0:0], "Off", "On"
```
noisymime: If these are simple on/off bits, then they are being represented incorrectly in these ini lines. The memory address (0 in this case) should not increment for each bit as that are all part of the same overall byte. 

Have a look at the 'bits' entries on the other pages and see if that's how you were trying to use them.

Autohome2: These are simple on/off bits , but have been given an entry each to allow for other flags to be added to each input individually​ in a further development of the code

Autohome2:
The enable_candata_out  @ ini 50 is also spaced away from the input stuff to allow for expansion too. :-)

PR #96, opened 2017-08-04
Update for STM32

VitorBoss:


Autohome2:
The use of specific pin definition (eg pc3) in this version is much simpler and easier to follow previously.
If the useable frequency range of stm32 pwm is up to 5khz? Doesn't the TS ini need changing to allow correct selection?

VitorBoss:
The resolution is same of injectors(2uS), this give a minimum 8Hz(62500) and a maximum 5KHz(100) of ticks

PR #104, opened 2017-10-14
Antilag and Rotational Idle testing.

AKCore:
Basic anti-lag and rotational idle added. Both tested to work, Rotational idle is not cutting enough events though.


adbancroft:
This PR can no longer be merged. Please resubmit against the latest master.

PR #105, opened 2017-10-14
Launch Control TPS

ConnerMcLaughlin:
This is the enable/disable Launch Control and Flat Shift with throttle angle.
Enables at and above set percent.


ConnerMcLaughlin:
As per Stewart's suggestion, I removed the enable/disable by TPS from the Flat Shifting.  Thank You Mr. Stewart for pulling this code.

PR #112, opened 2017-11-13
Hardware testpage

Autohome2:
added 8 generic output pin hardware tests
50%dc at 1-65534ms pulse time on spark,fuel,and the 8 generic outputs.
inhibit added to stop starting the engine while in test mode
groundwork for stepper and iac tests in ts


dxControl:
Something I forgot to mention was the detection of default pins. TS will show a difference the first time you load after the detection, simply use the device version , it won't come up again unless you ever choose default pin again.the default pin in TS will now show the default for the selected board type not say default.

karniv00l:
I would very much like to see that feature merged. Is there anything I could help?

noisymime:
Closing as this is now mostly in place. Still need to add testing for non-fuel/ignition pins, but this can be done using the programmable IO now.

PR #113, opened 2017-11-15
2-Speed fan, turn on safety check, option for selecting analog pins in TS

joemper:
Fan don't start while engine is off, either when rpm is below cranking threshold. 
A9-A15 is selectable for fan pins. 
Pin mappings only added for V3 board, test with this board selected.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -97,7 +97,7 @@
 
    endianness          = little
    nPages              = 10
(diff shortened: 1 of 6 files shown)


Autohome2:
If the fan is on when engine stops will it continue to run ? It should be possible to have key on engine off with fan running incase you needed..

joemper:
No it will stop and start again when rpm is above cranking threshold.

PR #114, opened 2017-11-16
2-Speed Fan, Fan Don't Run While Cranking

joemper:
Added option in TS for selecting A9-A15 as fan outputs.
Pin mappings for other boards than V3 is yet to be added.


adbancroft:
This PR can no longer be merged. Please resubmit against the latest master.

PR #134, opened 2018-03-17
STM32 fast update

VitorBoss:
Changed the maximum number of injectors for STM32F1 boards
Better handling of VVT, Boot and Idle interrupts and setup
Fixed libmaple FastMath resolution, now is 12 bit
Added option to  change analog resolution on STM32GENERIC core using analogReadResolution(10);
Removed redundant CAN initialization code.
Added a variable currentVersion for EEPROM version to simplify the code reading and debugging.
Moved initialiseTimers(); because was freezing STM initialization.
Minor cosmetic changes.


Autohome2:
The CAN init for teensy and stm32 should be kept seperate as they both have different libs and methods?
 Is the ability for teensy and stm32 to use direct serial for serial3 Comms purposes still possible?

noisymime:
Thanks as always Vitor!

I've added a few comments in for review, but mostly this looks OK.
The one big problem with it is that it's failing to build currently, at least on PIO. Have a look at the Travis build output, it looks like there's a TIMER5 reference somewhere there shouldn't be: https://travis-ci.org/noisymime/speeduino/builds/354613279#L1518

VitorBoss:
@noisymime All fixed

noisymime:
Thanks Vitor!

noisymime:
I'm reverting the changes from updates.ino here. Per the comment above in that file, this is breaking all sorts of things as the new current version isn't getting updated at every step

Issue #131, opened 2018-02-26
Reverse option for TPS [$5]


PSIG-dev:
As previously discussed on Slack for a "reverse" checkbox in TS, and one code solution posted and tested here <https://speeduino.com/forum/viewtopic.php?f=19&t=1159#p18146>.

VitorBoss:
That was discussed almost a year ago. I had forgot about this

ConnerMcLaughlin:
Need some testers for this.
https://github.com/ConnerMcLaughlin/speeduino/tree/revtps

noisymime:
I've spent a bit of time playing with various options for this and have decided on an auto switching one. Due to the way the calibration works in TS, I cannot add a checkbox or dropdown to the calibration dialog, which is really the only place it makes sense to have it. 

In order to not confuse people by burying an option somewhere it may not be found, I think auto switching when the values are 'backwards' is best. The user will still get the warning, but the TPS readings will then be correct. 

Commit incoming to close this issue

PR #144, opened 2018-06-13
Invert TPS toggle option

ConnerMcLaughlin:
The vast majority of us physically swap the wires and/or terminals on our throttle position sensor to get the signal range moving in the right direction.  For the rest, there is this option which does it for you programmaticaly.


VitorBoss:
That is a fix for #131?

ConnerMcLaughlin:
Yes this is for #131 .  It would be awesome if you're testing it out.

ConnerMcLaughlin:
Stewart has a better solution without the need of a checkbox

PR #156, opened 2018-08-13
Fix Teensy IGN2 pin mapping

johnxx:
Correctly set pin mapping on Speeduino 0.4.x w/ Teensy Adapter.


noisymime:
Confirmed and merged. Thanks @johnxx !

PR #174, opened 2018-12-07
Added secondary fan control and main relay control

ByteVenom:
Also updated UA4C board definition to its default secondary fan control pin.


ByteVenom:
What does that fail mean? It looks like a file couldn't be found.

Autohome2:
The eeprom mapping changes are incomplete

ByteVenom:
> The eeprom mapping changes are incomplete

Changed the mapping in Storage.h, I'm not sure how I missed that. Thank you.

PR #181, opened 2019-01-06
Fix a few typos and comment formatting consistency

ConnorRigby:
Was skimming through the source and found a few typos and inconsistencies in comment formatting.


noisymime:
It's all good, the compile breakage isn't from your PR (It's something that has changed between stm32 4.4 and 4.6 libraries). 

Everything in here looks good, thanks for submitting.

Issue #132, opened 2018-02-26
Reverse option for Stepper IAC


PSIG-dev:
Same reverse option TS checkbox for correct stepper direction.

PR #182, opened 2019-01-08
Idle stepper direction changing

ric355:
I'd be grateful if this change could be integrated back into the mainline. It adds an "Inverted Y/N" setting to invert the direction of the stepper motor for situations where it is wired the opposite way around such that the forward direction in the software results in it being physically driven backwards, and vice versa.

I've tested it with open loop stepper mode on my vehicle.

I think I may even have seen an enhancement request for it some time in the past (#132 )


noisymime:
Looks good to me, thanks for the submission!

PR #186, opened 2019-01-14
Option to disable Thermo Fan Control if ENGINE OFF

cmjdev:
Added ability to only run thermo fan control when engine is running.

Added 8 bytes to configPage6 to account for the added fan variable needed.

Updated speeduino.ini to reflect change


cmjdev:
This does not remap EEPROM.. Didn't realize I needed to.  Will resubmit when fixed.

PR #188, opened 2019-01-16
Option to only run FAN when ENGINE ON

cmjdev:
Added flag to only run fan when engine is on.

Flag variable was added to configPage2 unused space.


noisymime:
This is great work and a solid quality first PR. Thanks for the submission!

cmjdev:
>           This is great work and a solid quality first PR. Thanks for the submission!

Thanks.  I've hacked on the code quite a bit for my own use, so I've got a few other features i'll try to organize to an acceptable degree to be included.  Great project!

PR #194, opened 2019-02-03
Ability to separate the tacho pulsewidth from ignition dwell

ric355:
I probably need to give some background to this one as I don't know if the post I made on the forum a few weeks ago was seen.  My car has a V8 and was originally distributor driven. The tacho takes a high voltage pulse from the coil negative as is pretty common.  My Speeduino conversion has its wasted spark ignition dwell set to 3ms and it uses a relay coil and transistor to generate an equivalent the HV pulse from the speeduino tacho output.  Currently, the firmware generates a tacho pulse for every ignition event (assuming it is not set at half speed), and the length of the pulse is equal to the dwell time. This means that by the time you get to around 3800rpm you're getting a sufficiently saturated output that the tacho can't discern the pulses anymore and the needle drops back to zero.

For example 3800rpm / 60s * 4ign/s * 4ms tacho pulse width = 1013ms, so more than the full second if you assume you need a 1ms gap between pulses minimum to keep the tacho happy (the 4ms is 3ms dwell and 1ms gap).

This code change separates the generation of the tacho pulse from the ignition dwell time. It does it using the following method:
- TS config to switch the feature on or off in the Tacho auxiliary dialog
- TS config to set the the maximum tacho pulsewidth on the same dialog. Max tacho pulsewidth is limited to (dwell - 0.5ms) and minimum is 0.1ms.
- The ignition ISRs have been modified to introduce an INTERMEDIATE stage which occurs between the PENDING and RUNNING states. The intermediate stage is timed to occur at the moment matching the configured tacho pulsewidth and it is here that the tacho pulse high call is made, instead of in the endcallback function. The RUNNING is reduced in length by the tacho pulsewidth. Due consideration has been paid to new ignition mode here as this has an absolute end time based on crank angle instead.
- If the tacho pulsewidth is set to the same or greater than the dwell time in use then the tacho pulse is generated at the end of the dwell period as it is now.

I have made the changes on both my October branch and on the mainline branch. This pull request contains the code rebased to the latest the mainline, and thus contains changes for all 8 ignition channels.

For testing, I have done various oscilloscope measurements to try and ensure that the ignition timing is not being affected, and tested on both new ignition mode and old mode. I have also of course had it on a running vehicle using my October branch and verified it addresses the issue. It doesn't appear to make any noticeable difference to loops per second.

I'm not sure what the appetite for this is as it makes a change to a key area but I would be grateful if you could consider integrating it so that I can continue to take the main release without needing modifications. I know NickZ and curtis12321 also had this same problem with their V8's.

Note I only have the capability to test wasted spark V8 so can't test the full 8 ignition channels.


noisymime:
So, I haven't been ignoring this one (I promise!) just wanted to knock a few other things off before looking at it more closely. 

I have a different method in mind for how to achieve this that is much less invasive and won't trigger any additional interrupts. It will allow the selection of the tacho pulse duration, but only in whole ms increments (Eg 1ms, 2ms, 3ms etc). This should be sufficient for most tacho systems though I believe. 

I'm going to close this PR off as I think there is a better way of solving the same problem, but I do greatly appreciate this submission and the time you've taken with it! I'll open a separate issue ticket to track this item

PR #197, opened 2019-02-11
AlphaMods - rough extra features for Arduino

ElDominio:
slew of extra features, which I have running on a few cars, but as such, is highly experimental and I take no responsibility to anything that happens if you so choose to run it.
Also, all my functions are hardcoded into firmware, since I don't know how to interface things with TunerStudio, and I have no real need for that since I can just change things as I go in Arduino IDE.

My firmware has these features:
1) Idle ignition timing compensation:
this function adds and removes timing to try and keep a stable idle RPM of 800. This can be changed by changing the "ZeroThrottleTiming" function in corrections.ino
This function also adds "pops and bangs" in a range from 3000RPM to 5000RPM since it sets timing to -5

2) WOT Dwell correction:
I've used this on some cars where they run a certain dwell while cruising, but need less dwell on WOT conditions (happened to me on Hyundai wasted spark coil packs), this usually needs to be lowered (or disabled) from what is in the repo.

3) A/C control:
This function expects a high input on pin 26 for A/C request, and a combined temperature are pressure signal that goes high on pin 28 (since it was coded for a Tiburon which works like this), that when both of these are high, it sets a flag that AC should be on. This then alters the idle duty one second before engaging the A/C compressor, as to stabilize idle at a higher load to compensate for the compressor itself. Compressor output is expected to be on pin 44

4) DFCO wait time:
This waits a few seconds to engage DFCO, instead of engaging immediately once conditions are met. the firmware has this set to 2 seconds.

5) VVL control:
This function, as stated, controls a VVL solenoid (on/off). it engages it at what used to be 5800, with a 200RPM hysterisis to 6000RPM

6) Check Engine Light control:
This function turns on a Check Engine light if any of the conditions in the function itself are true, which also means it is used as a shift light. The output is expected to be on pin 30 (this one I haven't had much time to play around with)

7) open loop boost assist:
This function holds boost wastegate duty to 90% when WOT until a certain kPa has been reached, then switches to open loop control.

8) upshift timing correction:
this is used on automatic cars, when rpmDOT is negative, timing is reduced by 10 degrees to assist with shifting and reduce wear on the transmission.

And a few other ones I will document later


ElDominio:
> FYI idleRPMtrg, min and max were 16 bits variables because idle RPM is likely to be greater than 255. Such a value would overflow in an 8-bit value, which would lead to errors.

I'm still figuring out GIT lol

Anywho, you mean as received from TunerStudio? I'm pretty sure TS has support for conversion factors, does it not? We had talked about using *10 on all RPM values to allow up to 2550

noisymime:
This was always too large to pull in a single PR and many of these features have now been implemented as their own PRs.

PR #201, opened 2019-02-19
Serial request 'a' handler

brunob45:
(Work in progress)
A feature discussed in [this forum thread](https://speeduino.com/forum/viewtopic.php?f=12&t=2602).
[RealDash](https://realdash.net/) is an iOS and Android app that can display realtime data through a sophisticated gauges and indicators. It is very much like MsDroid, minus the tuning capabilities.
The app already offers support for MegaSquirt ECUs, using the MS2 serial protocol.

This PR is proposing a handler for such protocol, which is not far from the MS3 protocol that Speeduino already supports.


noisymime:
Thanks for this, it looks like a good addition. I hadn't realised that RealDash was using the old, fixed legacy read out. 

I'll have a closer look at this tonight, but at a glance the only thing I can see that won't be right are the 2 temperature read outs (CLT and IAT). These will need to be sent as a **signed** value, taking into account the offset (`CALIBRATION_TEMPERATURE_OFFSET`). It also looks like these values are always expected to be in degrees F, so this will need to be taken into account if C is being used. 

On the whole though, this looks very clean, so should be a no brainer to merge :)

janimm:
RealDash uses the old legacy 'a' protocol when there is no INI file loaded for the connection. This was done to get 'something' working even without Megasquirt ini file, since we do not have access to ini file server and to our understanding we are not allowed to serve the ini files since they are proprietary of Megasquirt developers.

If you import the ini file in connection settings, then the new serial protocol is used, based on information parsed from ini file.

noisymime:
@janimm Thanks for the background info straight from the source :)

Whilst the ini is probably the preferred option for 'everything', I really like the idea of this working out of the box using the legacy system as it makes life for users a LOT simpler. I'll make a couple of tweaks to this PR, get it committed and then try it out. Would be great to have Speeduino supported on RealDash.

janimm:
@noisymime I agree. Forcing users find their ini file is a nuisance. Megasquirt is able to push data faster on new protocol, but that probably is not an issue with your system.

brunob45:
I've tried loading the megasquirt compatible speeduino.ini file into Real Dash, but it still used the legacy 'a' protocol.
@janimm great work with Real Dash, I'm really looking forward to using it this summer!

PR #204, opened 2019-02-23
Dev stm32 f407 ve

Tjeerdie:
Hi
A working version of the stm32F407VE boards, it is targeted at the black_stm32F407VET6 boards on ebay. The build works with the ST stm32duino core. But a few changes are necessary to make this work.  To make it work checkout the updated repository for the STM32duino core in my github as well. To compile use the arduino IDE (not tested with the platform IO IDE). 

Still TODO. The sequential setting does not seem to work correctly. Possible a bug that is also present in the atmega version.


VitorBoss:
Nice PR, for HardwareTimer I'm using this: https://github.com/huaweiwx/STM32GENERIC/tree/master/STM32/libraries/HardwareTimer/src
It supports up to 22 timers.

Tjeerdie:
@VitorBoss  The problem is that i do not use the STM32GENERIC core. I use the ST STM32duino core therefore i cannot use any other hardwaretimer libarary than the one provided by the ST STM32duino core. I see if i can change that one to be used in the ST stm32duino core.

noisymime:
So there's obviously a lot to go through on this PR. At a high level, I'm fine with the actual stm32 stuff, including the switch to the official core, but there are a couple of things that I'm a little less convinced on.

1. I'm not sure whether moving the pin mappings into the board files is the best way to go. Neither option is really terrific, but there is a whole heap of duplication by moving it into the boards. I'll have a bit more of a think about this and see which way is preferred.
2. Around the storage side of things, I've begun playing with a generic SPI storage library that is the equivalent of the existing EEPROM stuff in storage.ino. Given we both appear to be targeting the f407 black board with the extra onboard flash, I think that's going to be the preferred method for these boards. 

Otherwise though, terrific PR.

Tjeerdie:
Hi noisymime

My considerations for doing it the way I did.

1. The reasoning for me is that all defines in the init, wrapped in a case statement are not readable. For every -Driver board case statement- every -CPU board- has its own “if define” around it to make the pin mapping. It seems logical to move it to the –CPU- board definitions. I change it back if you are not happy with it ;-). 
2.	The trouble with the SPI storage is that is handles reads/writes/erases in blocks of >256bytes (or other page sizes depending on the EEPROM). The EEPROM equivalent in the AVR does this on a per byte basis. If the spi flash was writing per byte the flash would wear 256 times as quick :-D, and is 256 times as slow. The same page size issue holds true for the internal flash as EEPROM. I got around that problem quick and dirty using battery backed ram as no there is no wear. This is the easiest way to keep the existing code structure (writing per byte) for writing the settings to a storage location. I agree there needs to be a more permanent solution with the spi EEPROM but that takes a little more work by changing the storage implementation. Write a variable size page at the time, depending on the storage page size.

noisymime:
I'm going to pull this in now, but then do a little cleaning up to make the 2 cores (Official and stm32generic) work cleanly. I do think having the 2 split into different files is the better way to go.

Issue #191, opened 2019-01-28
Switching from Alt to Sim halves the pulse width


Old-Grey6:
Reqd Fuel - 10
PW 10ms
Squirts - 2
Alternating
4-stroke
4 cyl
4 INJs
Paired
I switch to Simultaneous, and PW halves(5ms).
They both have the same # of INJ pulses, just Alt offsets ch#2 by 180º, yet Sim halves the PW.
Test is here - https://youtu.be/sYio0XROS1I
And it produces this -
https://www.mediafire.com/convkey/3a0b/qnebtx4usft2hqu6g.jpg
Like PSIG says, this is a bit of a dangerous situation. Like one poster on the forum switched from Alt to Sim, but luckily didn't go for a drive after it started idling rough. He didn't know why, and I don't think he had O2, so it was potentially bad.
Thread - http://speeduino.com/forum/viewtopic.php?f=21&t=2556

Is there an undocumented convention, or is this a bug?

brunob45:
This is a convention.
2 squirt alternative is like : 
_*____/_*____/
____*_/____*_/
and 2 squirt sim is like :
_*__*_/_*__*_/
_*__*_/_*__*_/
where (*) is an injection event and / is the end of an engine cycle
As you can see, simultaneous make the injections happen twice as much, so we cut the PW to compensate

Old-Grey6:
Well then that's a FW bug, because Alt will continue to squirt 2 pulses when set on 2 squirts, or from the standpoint of Sim, Sim does not double the # of squirts.  Also there is not a documented warning in the Wiki telling you you MUST change to 1 squirt when using Alt, or double squirts in Sim. And even further it won't let you select 1 squirt in Alt anyway. 
    On Monday, 28 January 2019, 11:52:39 am AEDT, Bruno Bousquet <notifications@github.com> wrote:  
 
 
This is a convention.
2 squirt alternative is like :
*/*/
*/*/
and 2 squirt sim is like :
__/__/
__/__/
where (*) is an injection event and / is the end of an engine cycle
As you can see, simultaneous make the injections happen twice as much, so we cut the PW to compensate

Old-Grey6:
[...]

PR #211, opened 2019-03-08
Adjust semi-seq alternating injection pattern for 4 cylinders

brunob45:
As discussed in issue [#191](https://github.com/noisymime/speeduino/issues/191),
switching from Alternating to Simultaneous injection halves the pulse width.
I propose a simple solution for the 4-cylinder configuration.

Sequential injection, 1 squirt simultaneous and 2 squirts alternating all output the same pulse width.
2 squirts simultaneous and 4 alternating as well.

This should correlate with the expected behaviour.

To adapt this change to older tunes, one should only have to double the squirt number, if using alternating.


noisymime:
Thanks for the PR, but there's some problems with how this will work. Semi-Seq setups can't run over 720 degrees because you can't be sure that you will have a trigger pattern that supports it. 

**Eg** Semi-Sequential works fine on a missing tooth arrangement that doesn't have a cam pulse. If you try to run the injection timing over 720 degrees though, the angle will never go above 360 and so you'll never see the 2nd lot of pulses you're adding

brunob45:
Oh, I see...
Then, we would need to put injectors 1-4 and 2-3 on the same event (eg fuelSchedule1), then alternate between the two sets. Would that be an option?

brunob45:
Alternating injection now works with 360 degree triggers and still behave as expected (for 4 cylinders).

brunob45:
@noisymime I tested this fix with:
- Semi sequential injection, alternating and simultaneous, 
- Missing tooth at crank speed, without cam signal
- MAX_INJ_ANGLE = 720
- 4 strokes
And it worked.
Am I missing something?

PSIG-dev:
From the thread linked in #191: **_"TL;DR — the calculated and actual fuel flow should remain the same when changing either the mode or the number of injections."_**  This is important, and why I am bumping it.  Also bumping #191.  Thanks for the efforts to resolve this!

PSIG-dev:
This item (#211) has been deleted, not sure why.  Any updates?

brunob45:
Just did some other tests with the code in this PR, got these numbers.
I also added some basic graphs showing the behaviour of the injectors for each configuration (not to scale and made on paint, excuse my lack of scope).
All the tests were at 150 rpm, 15% ve and basic distributor trigger.
v1 0
And, _I think_, this has not been merged into master because it will break the tunes of people running alternating, 2 squirts or more.

PSIG-dev:
Cool. I am assuming this only represents timing, and the pulse widths in each are halved when they go to 2-per, or 25% each for 4, etc?  However for timing, my primary reaction is that the Alternating does not follow the sequence of 1-2-4 squirts, and instead shows settings of 2-per-cycle instead of 1, for the 1-squirt, like the other modes.  It may at-first seem minor, but was the primary error in function logic in "other" systems previously.  

I also don't think we should carry a basic error forward like this because it may 'break' older tunes — rather just tell users the ultra-simple **one-time** correction and we are past it.  Considering that users are having to deal with Req_fuel changes as it is now on every affected mode change, it would be a big improvement.  Awesome work and thanks for the visual.

brunob45:
For the pulse width, correct. As the value on the gauge (minus 1ms deadtime) is 50% when going 2 squirts, and 25% with 4.
As of the alternating settings, we cannot set 1 squirt alternating, as it is hardcoded in Tunerstudio. Trying to do so will throw an error:

We could think of the alternating settings as the number of ignition _events_ (there are two that alternate eachother).
As of now, my changes only works on 4 and 6 cylinders. 2 and 3 are not affected by this bug, I think. So I'm only missing 5 and 8 cylinder engines.

brunob45:
Just changed (yet again) the proposed solution.
Now, I have replaced the 'std_injection' panel in TunerStudio by custom fields, so there is no more, "number of squirts must be evenly divisible by the number of cylinders" errors : I replaced the "divider" field of the tune by a "nSquirts" field, which let us put whatever value we want.
Also, this fixes the issue discussed above.

PR #220, opened 2019-04-11
Coolant based ASE %, ASE duration, priming pulse width and Coolant based ignition adjustment

pazi88:
Speeduino was lacking coolant based After Start Enrichment, After Start Enrichment duration and Priming pulse widths. Which are mandatory features when using car in cold environments. Also the ignition timing needs to be adjusted based on coolant temperature in cold environments to maintain engine performance also with cold engine. So that feature is also added as Cold Advance.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -207,9 +207,9 @@
 page = 1
       unused2-1     = scalar, S08,       0,         "kPa",      1.0,       0.0,   -127,    127,       0
       unused2-2     = scalar, U08,       1,         "kPa",      1.0,       0.0,   0.0,    255,       0
-      asePct        = scalar, U08,       2,         "%",        1.0,       0.0,   0.0,    95.0,      0
-      aseCount      = scalar, U08,       3,         "s",        1.0,       0.0,   0.0,    255,       0
-      wueRates      = array,  U08,       4, [10],   "%",        1.0,       0.0,   0.0,    255,      0
+      unused2-3     = scalar, U08,       2,         "%",        1.0,       0.0,   0.0,    95.0,      0 ;This used to be asePct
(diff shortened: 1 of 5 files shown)


VitorBoss:
The table lookup could be perform just once a second if you add a variable and put lookup on speeduino.ino#L279, temperatures don't spikes on a second IMAO.

Good job.

pazi88:
@VitorBoss Yes you are right, but I wanted to make this simple as possible, without doing too many changes.

I have now tested all these changes so that they actually work. The third commit is because I needed to fix the priming pulse width.

noisymime:
I think there's an improvement to be made here with the coolant read on startup. I suspect you're probably not getting an accurate value due to it being the first read. The filter will be (or could be) dramatically distorting the true read. I'll add an option on the coolant sensor read to bypass the filter if desired.

pazi88:
Oh you are right. I didn't think about that. I tested this on my tune where I have filters set to 0, because that gives me stable readings in my setup. So the filters didn't distort readings. And yeah bypassing the filter for this would work fine, becsuse it doesn't matter if the reading would few degrees off, because possible noise.

lsdlsd88:
is it me or the new ASE dialogs need some GUI tweaks? here you can see the FONT SIZE is very different from the usual. adding default values in the base tune would also be more user friendly.

ase

pazi88:
The TS is very strange about that. The fonts looked strange on my PC also but I didn't find anywhere to adjust those. Not even after hours of trying to figure that out. But then after populating tables with sensible values, saving tune and closing the TS. The fonts were ok next time I opened TS. And yes some default values would be ok. I think josh already made change that for old tunes, the old ASE, priming etc values are used to populate those curves when updating FW version.

PR #231, opened 2019-06-07
Pin definition and Case 52 for Teensy 8x8

NickZ1969:
Pin definition and Case 52 for Teensy 8x8
added "Teensy 8x8" in to speeduino.ini
added Case 52 and pin definition into init.ino for Teensy 8x8


adbancroft:

PR #232, opened 2019-06-15
Pure Alpha-N changes

RenanBecker:
Some changes to allow live baro correction when the onboard map sensor is left unconnect in pure alpha-n setups and some minors changes in speeduino.ini to make the dashboard more useful


noisymime:
Replaced by #670

PR #237, opened 2019-07-19
All ignition and fuel timers run at same speed

brunob45:
Remove the need for slow versions of MAX_TIMER_PERIOD and uS_TO_TIMER_COMPARE() macros.
Enables the possibility to make the schedule code cleaner.

### DIFF
--- a/speeduino/board_avr2560.h
+++ b/speeduino/board_avr2560.h
@@ -37,7 +37,7 @@
   #define FUEL2_COUNTER TCNT3
   #define FUEL3_COUNTER TCNT3
   #define FUEL4_COUNTER TCNT4
-  #define FUEL5_COUNTER TCNT1
+  #define FUEL5_COUNTER TCNT4
   #define FUEL6_COUNTER TCNT4 //Replaces ignition 4
   #define FUEL7_COUNTER TCNT5 //Replaces ignition 3
   #define FUEL8_COUNTER TCNT5 //Replaces ignition 2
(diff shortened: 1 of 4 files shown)


brunob45:
> On lines 7-11 in board_avr2560.ino. You should used pre-processor constants instead of killing out 5 bytes of mem.

After compiling both options, they produce programs with the same size. I decided to follow your comment so it is more explicit

noisymime:
> After compiling both options, they produce programs with the same size. I decided to follow your comment so it is more explicit

The program size will be the same, but the amount of ram used when the firmware runs will be different. Using the `#defines`, nothing actually gets defined anywhere in memory (Other than on the registers when they're set). 

Just as a note though, you can remove the `#undefs` as they don't actually make any difference (And are also prohibited under MISRA).

brunob45:
> Other than on the registers when they're set

Well, the value they're set to must come from somewhere, right?

noisymime:
> > Other than on the registers when they're set
> 
> Well, the value they're set to must come from somewhere, right?
> [Edit] I compared the binary .hex files generated from the compilation of each variant, and they are exactly the same.

If they're variables, they'll be coming from program space, not from SRAM. Program space and SRAM are very different things on a Harvard architecture MCU and Speeduino has a heap of program space to work with, but not a lot of RAM. 

In this case given that their value is never changed, the compiler is almost certainly converting them to hard values anyway, hence why you're seeing identical hex files. My guess (And I haven't tried this) is that if you compiled it with `-Os` compile flag, there would be a difference, but it's just a guess.

brunob45:
I tested again the ignition, fuel and idle outputs with a scope and they're all working fine

PR #238, opened 2019-08-01
fix 'unsigned >= 0 always true' warnings

brunob45:


noisymime:
I actually had similar changes lined up locally that I hadn't pushed yet, so I'll close this one for now. Thanks hugely for submitting this though!

Just around the change in the `crankMaths.h` file, adding a `%` in there would cause a fairly significant performance hit, so it's better to address it with some relevant casts. Have a look in b6108b5 at how I've done this if you're interested.

brunob45:
That also works! I forgot that the modulo operator was that complicated to implement on an arduino.

noisymime:
Modulo uses 1 or 2 divide operations in the background and with no hardware divide (Something that is still unfortunately common on newer MCUs as well) it's best avoided if possible. The same is true even for some of the stm32 MCUs that are potential platforms for the firmware, so it's good practice to minimise their use.

PR #242, opened 2019-08-12
Add unit testing to the speeduino toolchain

brunob45:
I made an easy setup to do unit tests on the speeduino code.
To run the tests on Linux:
- make sure you have cmake, make, gcc and python installed 
- in a terminal, navigate to the tests directory 
- enter the command "make" to compile the tests
- enter the command "./tests" to run the tests

At the moment, only a few things from the setup() function are tested, but it's easy to add more.


brunob45:
I created a separate repo for unit testing, it will make things simpler : [speeduino-unittesting](https://github.com/brunob45/speeduino-unittesting)

PR #246, opened 2019-08-16
currentStatus struct reorder

ConnerMcLaughlin:
This is an attempt to conform to a rigid orthodoxy order of the struct currentStatus' members.  This being a prequel to unifying the multiple datasets that represent live sensor/engine data.


ConnerMcLaughlin:
currentStatus.RPM and currentStatus.MAP both make since to be unsigned because there are no negative RPMs and MAP is an absolute value.  currentStatus.vvtAngle on the other hand can be negative during calibration.  This makes a dilemma of using only one method(integerPID) for all.  The current solution is to type cast the pointers to unsigned variables to  int16_t * but later function overloading might be a better option.

noisymime:
Closing as stale

PR #248, opened 2019-08-20
Fix second revolution check for MAP minimum

tvoyle:
Second revolution is never picked up with MAP Sample Method Set to cycle minimum as  the +1 on the second check is in the wrong place. This causes issues with a single cylinder engine as there is only one MAP pulse every two cycles, but is sampled every cycle causing the MAP to fluctuate between two values


noisymime:
Good pickup! Thanks for the PR

PR #250, opened 2019-08-27
fix confusing engine constants window

brunob45:
This is my take at fixing the confusing Engine Constants window in TS.

With this change, we get:
* Injector Layout: (not changed)
  * Paired: Each channel controls 2 injectors
  * Semi-Sequential: Each channel controls 1 injector, but they fire in pairs
  * Sequential: Each channel controls 1 injectors, and they are all independent.
* Squirts per Engine Cycle: it's the number of times each injector fires per cycle (360° for 2-stroke, 720° for 4-stroke), no matter the other settings, and every value from 1 to 8 is valid.
* Injector Staging:
  * Simultaneous: All injectors are fired at once every "squirt cycle" (engine cycle / squirts per engine cycle)
  * Alternating: All channels are fired in sequence over the "squirt cycle"
* Req Fuel: same as before, but the calculator utility is gone. We could check with the TS devs if we could extract it from the default panel

Here's a preview of the new look:
engine_constants

The code implements the fixes, and comes with a fully functional update function.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -233,7 +233,7 @@ page = 1
       displayB2     = bits,   U08,      23, [4:7],  "RPM", "PW", "Advance", "VE", "GammaE", "TPS", "IAT", "CLT"
 
       reqFuel       = scalar, U08,      24,        "ms",        0.1,       0.0,   0.0,    25.5,      1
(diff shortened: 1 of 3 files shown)


noisymime:
Closing as stale. Some of the GUI layout work may still be of value, so please feel free to resubmit as their own PR

PR #252, opened 2019-09-05
Add initial support for reading fuel temp from Flex input

### LINKED ISSUE #165: Fuel Temp Input
Could you add fuel temperature as an input from a flex sensor? The sensor reads the temp anyways and it comes in as a duty cycle from what I remember, so I would think it's reasonably doable.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/63557374-fuel-temp-input?utm_campaign=plugin&utm_content=tracker%2F706250&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F706250&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ConnorRigby:
fixes #165

### DIFF
--- a/speeduino/sensors.ino
+++ b/speeduino/sensors.ino
@@ -435,7 +435,17 @@ void readBat()
  */
 void flexPulse()
 {
-  ++flexCounter;
+  flexPulseState = !flexPulseState;
+  if(flexPulseState) {
+    //if flexPulse is HIGH, increment the counter and calculate pulsewidth
+    ++flexCounter;
+    //convert microseconds to milliseconds
+    flexPulseWidth = micros_safe() - flexStartTime;
+    flexStartTime = 0;
+  } else {
+    //flexPulse is LOW, store micros for calculation later.
+    flexStartTime = micros_safe();
+  }
 }
 
 /*
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -3213,7 +3213,7 @@ cmdtestspk450dc = "E\x03\x0C"
    ; you change it.
 
    ochGetCommand    = "r\$tsCanId\x30%2o%2c"
-   ochBlockSize     =  98
+   ochBlockSize     =  99
 
    secl             = scalar, U08,  0, "sec",    1.000, 0.000
    status1          = scalar, U08,  1, "bits",   1.000, 0.000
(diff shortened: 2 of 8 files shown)


speeduino/sensors.ino
```diff
@@ -435,7 +435,17 @@ void readBat()
  */
 void flexPulse()
 {
-  ++flexCounter;
+  flexPulseState = !flexPulseState;
+  if(flexPulseState) {
+    //if flexPulse is HIGH, increment the counter and calculate pulsewidth
+    ++flexCounter;
+    //convert microseconds to milliseconds
+    flexPulseWidth = flexStartTime * US_IN_MS;
```
noisymime: Is this right? Shouldn't `flexPulseWidth = micros_safe() - flexStartTime`?

ConnorRigby:
@noisymime ah i think you're correct. Updated

iLeeeZi:
I tested this pull request in car with a continental 13577379 ethanol sensor. Temperature was stable but it was reading low and after a bit of research I found a formula for continental sensors:
```Degree Celsius = (41.25 * pulsewidth(ms)) - 81.25``` and therefore the scale would be 1-5ms = -40c - 125c. Not sure if it applies to all sensors so configurable tunerstudio fields might be needed. Another thing was that in range conversion there was negative numbers and decimals and I changed those to be offset by 40 degrees  to be able to send the values to tunerstudio. Now it reads correctly at least in my bench setup.

I changed this:
```
      //Continental sensor reads temp from 1ms=-40c and 5ms=151.25°C
      if((flexPulseWidth >= 1000) && (flexPulseWidth <= 5000)) {
        currentStatus.fuelTemp = fastMap(flexPulseWidth, 1000, 5000, -40, 151.25);
      }

```
To this:
```
      //Continental sensor reads temp from 1ms=-40c and 5ms=125°C
      if((flexPulseWidth >= 1000) && (flexPulseWidth <= 5000)) 
      {
        currentStatus.fuelTemp = fastMap(flexPulseWidth, 1000, 5000, 0, 165); // currentStatus.fuelTemp is offset by 40 degrees to allow negative values.
      }
```

Otherwise this is a good pull request and it would be a good addition as it allows fuel temperature correction in future.

PR #255, opened 2019-09-21
Added SPI flash as EEPROM for STM32

Tjeerdie:
Added the SPI flash onboard the black STM32F407VET6 boards from ebay as EEPROM. It uses 1,5MegaByte of the flash and has therefore a LOT of wear leveling. The EEPROM size is just over 4kb. 

Give it a try on your own board to test it out. CORE_STM32_OFFICIAL  and SPIFLASH_AS_EEPROM must be defined for it to be incorporated. It is tested with the official STM32 arduino core from ST no other core is tested.  

Also made some changes to the board pinout. That can change in the future but now there is an over view of all the pins of the board. And it is clear which pin goes to what function.


noisymime:
This looks like some great work. I'll have to test in detail over the next few days, but most of it seems nicely separated from other areas, so should avoid any issues if there are any. 

What is is the purpose of the `NewWriteConfig()` function? I'm only skimmed through it, but I couldn't see any differences from the existing `writeConfig()`

Tjeerdie:
Hi!

Shit i did leave that in it. That's an attempt to re write everything. For
now not needed indeed the original wil be sufficient.

Tomorrow i wil try and fix it.

I tested the implementation with two Black stm32 boards on my desk. Works
well so far.

Regards
Tjeerd.

speeduino/storage.ino
```diff
@@ -33,6 +39,7 @@ and writes them to EEPROM as per the layout defined in storage.h
 */
 void writeConfig(byte tableNum)
 {
+  // EEPROM.begin();
```
Tjeerdie: can be removed

speeduino/storage.ino
```diff
@@ -11,6 +11,12 @@ A full copy of the license may be found in the projects root directory
 #include EEPROM_LIB_H //This is defined in the board .h files
 #include "storage.h"
 
+void startstorage(){
```
Tjeerdie: Need to initialize the SPI flash. This function can be moved further back in the file to be with the rest of the helper functions

speeduino/comms.ino
```diff
@@ -1641,9 +1641,10 @@ void receiveCalibration(byte tableID)
   bool every2nd = true;
   int x;
   int counter = 0;
-  pinMode(LED_BUILTIN, OUTPUT); //pinMode(13, OUTPUT);
+  // stm32 board has buildin led used as fuel pump
```
Tjeerdie: Can this be removed from the code? I see no real benefit other than debugging. When sending the Thermistor calibration the STM32 just froze. It took me forever to find out it had todo with analogwrite(LED_BUILDIN ..) or digitalWrite(LED_BUILTIN, LOW).

VitorBoss: Yes, I had included that 2 years ago for debugging STM32 without st-link as the code frozen on some parts of code. With st-link debugging is pretty easy though

speeduino/storage.ino
VitorBoss: Move it to board specific files, I had done this for FRAM

speeduino/src/SPIAsEEPROM/winbondflash.cpp
```diff
@@ -46,7 +46,7 @@ typedef struct {
     winbondFlashClass::partNumber pn;
     uint16_t id;
     uint32_t bytes;
-    uint16_t pages;
+    uint32_t pages;
```
Tjeerdie: This was a bug in the orginal code. It build fine with the arduino IDE. But not with platformio. this is the fix

speeduino/utils.h
```diff
@@ -6,6 +6,7 @@ These are some utility functions and variables used through the main code
 
 #include <Arduino.h>
 
+#define CONCATS(s1, s2) (s1" " s2) //needed for some reason. not defined correctly because of utils.h file of speeduino (same name as one in arduino core)
```
Tjeerdie: When using the CDC usb serial parts of the stm32 core some weird things happen with the defines. It has to do something with the util.h in the core defining __UTILS_H  and speeduino defines something like UTILS_H. This breaks compilation with usb CDC serial enabled. This line fixes it for now.

platformio.ini
```diff
@@ -58,10 +58,10 @@ platform = ststm32
 framework = arduino
 ;board = genericSTM32F407VET6
 board = black_f407ve
-lib_deps = EEPROM
+;lib_deps = EEPROM
 board_build.core = stm32
 ;build_flags = -fpermissive -std=gnu++11 -UBOARD_NR_GPIO_PINS -DCORE_STM32_OFFICIAL -DSRAM_AS_EEPROM
-build_flags = -fpermissive -std=gnu++11 -UBOARD_NR_GPIO_PINS -DCORE_STM32_OFFICIAL -DSPI_AS_EEPROM -DMENU_USB_SERIAL
+build_flags = -fpermissive -std=gnu++11 -UBOARD_NR_GPIO_PINS -DCORE_STM32_OFFICIAL -DSPIFLASH_AS_EEPROM -DUSBCON -DUSBD_VID=0x0483 "-DUSB_MANUFACTURER=\"Unknown\"" "-DUSB_PRODUCT=\"BLACK_F407VE\"" -DHAL_PCD_MODULE_ENABLED -DUSBD_USE_CDC -DHAL_UART_MODULE_ENABLED
```
Tjeerdie: Added a lot of defines that are missing for usb CDC serial port usage in the platform io build. Now you can connect tunerstudio to the board using the usb cable

speeduino/comms.ino
Tjeerdie: Do you recommend any manuals to get debugging with st-link v2 working under platformIO? If i set a breakpoint i get the error the program stopped for debugging!

speeduino/storage.ino
Tjeerdie: Storage init needs to be called before board_init is called. Also the begin() function is part of the EEPROM instance. Best to keep all the EEPROM function calls in the storage.ino.

VitorBoss: I agree, but we need to keep ever board related code on their specifc files, see https://github.com/noisymime/speeduino/blob/f185b1e200b78d74f43d187cbef540a78ebf40b0/speeduino/board_stm32_generic.h#L27

With that line the FRAM lib is started.

speeduino/comms.ino
VitorBoss: I'm using VScode for a while, it is much faster than PIO, but if you like it VScode have a PIO plugin too. On VScode I have to include some lines manually and you need openocd v0.10 , older versions don't work

speeduino/storage.ino
Tjeerdie: The downside of this is that we get calls/definitions to EEPROM in multiple places (and not only in storage where you would expect it)

I will take a look if it is possible to call a function in that location. Your example is a declaration and definition (or instantiation) of the variable EEPROM. That is not the same as function call to EEPROM.begin(). I will take a look when i get home tonight how to do this.

speeduino/comms.ino
Tjeerdie: I got debugging working in the end. It was simple, change the default build to the black_stm32 environment. I use VScode also but with the PlatformIO plugin to build arduino code. To much work to set it up oterwise :-D.

speeduino/storage.ino
Tjeerdie: Hmmm i did not get it to work without the call to EEPROM.begin() just before reading the configuration. 

If i try to use the SPI class instance during class instantiation of SPIAsEEPROM it does not send/receive any thing over the SPI (As seen with logic analyzer). Therefor this seems the only way for now.

PR #256, opened 2019-09-24
Single dataset

ConnerMcLaughlin:
This is an attempt at ridding the firmware of another data set that is redundant to the currentStatus struct.  The primary benefit to this code is ease of maintenance.  There are also features to relieve the serial buffer from blocking which had been a waste of cpu cycles. This code is not expected to be pulled but is a premonition of future code that will.


adbancroft:

PR #260, opened 2019-10-09
Adjust semi-seq alternating injection pattern: take 2

brunob45:
[WIP]
Step 1. Done. Sequential 4-stroke special case moved outside of (nCylinders) switchofn init.ino, it was the same for every of the 7 cases.
Step 2. Discuss whether, for alternating injection, nSquirts or reqFuel should be divided by 2 (because as of now, one of those is two times too much).
Step 3. Edit line 291 of init.ino to add the condition selected at step 2.


brunob45:
Updated PR description.

PR #275, opened 2019-11-03
Add idle ignition timing correction feature

iLeeeZi:
I added a new setting for idle ignition timing correction. Idle advance correction is useful if you don't have a idle air control valve or if idle control valve doesn't respond quick enough to sudden load changes.  This has been possible earlier by tweaking the main spark table but this makes it easier and allows better overall control.

This feature uses closed loop iac target rpm table and adjusts ignition advance at idle based on difference between current rpm and target idle rpm if conditions are met. Conditions are pretty basic at this time: active below TPS and active below RPM. Some oem throttle bodies or throttle position sensors have a separate idle switch and I added possibility to use it instead of TPS value.

Idle advance correction can be added to main spark table values or switched so when idle state is detected main spark table is ignored and ignition timing is what you set in the idle advance correction curve.

idleadvancesettings

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -315,11 +315,11 @@ page = 1
 #else
       primeBins     = array,  U08,       87, [4],  "F", 1.8,    -22.23,    -40,    215,      0
 #endif
-      idleSwitchPin       = bits,  U08,  91, [0:5], "Board Default", "INVALID", "INVALID", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "22", "23", "24", "25", "26", "27", "28", "29", "30", "31", "32", "33", "34", "35", "36", "37", "38", "39", "40", "41", "42", "43", "44", "45", "46", "47", "48", "49", "50", "51", "52", "53", "54", "A8", "A9", "A10", "A11", "A12", "A13", "A14", "A15", "INVALID"
-      idleSwitchPolarity  = bits,  U08,  91, [6:6], "Normal", "Inverted"
-      idleSwitchEnabled   = bits,  U08,  91, [7:7], "Off", "On"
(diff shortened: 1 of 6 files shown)


pazi88:
This is really good feature. Hopefully josh adds this to the master code.

cmjdev:
I have not pulled this to check it out, but based on the image, would it be better to be an adder to the base ignition instead of a static value?  Also, the delta direction looks backwards compared to the timing advance?

iLeeeZi:
It has option for fixed curve as it is set in image and if you switch idle advance correction from switched to added then it uses base ignition table and advances/retards ignition as much as values are set in curve. For example then you should set added ignition timing to 0 degrees at 0 rpm delta.

I was also pretty confused with delta direction but now it uses basically same delta value as closed loop idle control delta gauge set in tunerstudio and it is also kinda backwards but I don't know if it is intended to work like that.

ric355:
It feels like you have overloaded the word "Switch" here in the settings. Having the idle mode "switched" (which I think  means the idle advance numbers in your new table are used instead of the spark table) and having the idle advance mode of "idle switch" (which I think means use a switch instead of using a TPS value to indicate when the engine is expected to idle) is just a bit confusing.

I suggest the idle advance correction should be "Absolute" to mean use the values in the new table directly ("Direct" would be an alternative description but I prefer absolute personally). "Added" is fine for the other one.

iLeeeZi:
Those were indeed too similar names. I changed "Idle Switch" related names to "Closed Throttle Position Sensor" for now and left Switched/Added as second fuel table uses same terms for similar function.

noisymime:
I'm going to merge this for now as I think it's a really great addition and some quick testing seems positive.

That said, I do also have some concerns about the interface and terminology that I will look to clean up. I think it might also need something like an afterstart (Eg only engage 2s after the engine starts etc) delay as I had some difficulties starting with this enabled, seemingly not during cranking but immediately after it catches. Need more testing here. 

Overall though, big thanks for submitting this as it's a great addition.

pazi88:
@noisymime Yes you are right about the delay after start. It's pretty much mandatory. The MS has start delay adjustment for this feature and also CLT threshold which also could be useful.

noisymime:
@pazi88 The CLT one I need to think about it a little more. I would've thought this would be quite useful at lower CLT values just as much as higher ones, but I could (easily) be wrong about that

pazi88:
@noisymime I haven't personally used the CLT limit. I have only used the start delay in MS. Don't know how useful the CLT limit is, but the MS seems to have that limit.

PR #276, opened 2019-11-06
Added small pulsewidth correction

muuris:
Added possibility to adjust actual injector low range (0..2,55 ms) pulsewidth in relation to calculated pulsewidth (fuel amount). This is to compensate for non-linear injector flow in relation to pulsewidth. This is basically as in Megasquirt 3 (other ecus have this feature as well).

For example a specific injector may need 1,5 ms pulsewidth for the fuel quantity that should flow already at 1,0 ms, but after 2,0 ms pulsewidths the flow is linear.

Some modifications done after initial commit, this is now based on 201911 release and has 5 bins instead of 4. The lookup is done after all fuel corrections are applied but before injector opentime is added. 

My use case example: I've measured my injectors and they provide a linear fuel flow between 1,8 ms and 10 ms pulsewidths, but the low pulsewidth area is problematic. In 1,2-1,4ms area the flow is almost constant despite pw changing and then the flow drops very quickly at pulsewidths less than 1,2 ms. So as Speeduino tries to pull a little fuel, at first nothing is happening, and suddenly there is almost no fuel at all.

On TS settings page I've drawn a green line to represent calculated pw without/before this correction.

bosch0280158123 nonlinearity

smallpw_correction


muuris:
This hasn't been updated in a while, @noisymime could you comment on this -- if this is found useful, I'm happy to update this to be compatible with current master.

ninjawarrior821:
I need this... My poor car is idling at 11.3 AFR with larger injectors.

muuris:
> I need this... My poor car is idling at 11.3 AFR with larger injectors. Anything under 1.44 ms gets bad.

If not already using sequential injection, it will most likely make a difference. The feature in this PR is only useful if you have measured pw vs actual flow rate data.

This is old PR is a long way behind current firmware. It hasn't received any input on whether or not it would be included in the firmware, so I haven't been updating it.

ninjawarrior821:
> If not already using sequential injection
Yeah, that's the next step. They're FF640s, so Bosch EV14, which I believe are well documented already.

5peedmanual:
excuse me, how are you getting such low PW values? The lowest I can get is 5,5ms and I could use lower PW because my idle is too high

ggurov:
> excuse me, how are you getting such low PW values? The lowest I can get is 5,5ms and I could use lower PW because my idle is too high

come visit discord, can probably get help there better the commit comments.

adbancroft:

PR #280, opened 2019-11-13
6 and 8 cylinder sequential injection

iLeeeZi:
Mostly updated/added some currently unused 6-8 cylinder sequential code. I tried to avoid touching 5 cylinder stuff so this shouldn't break anything. Some changes may not be neccessary but I tried to make the code uniform between channels/cylinders.

Visible things I added are injector channel 5-6 close angle settings and hardware output tests.


VitorBoss:
On lines 420 to 427 you are performing 6 identical calls, try to call it just once to reduce unnecessary work.

iLeeeZi:
Thanks for the heads up. I disabled 6 cylinder individual fuel trims for now as using only two trim tables doesn't work with many applications and I don't know if two more trim tables can be added in future as they use quite a bit of eeprom.

iLeeeZi:
This should now allow 6 and 8 cylinder semi-sequential and sequential injection with the INJ_CHANNELS and IGN_CHANNELS change in globals.h.
Also fixes 5 cylinder paired injection as it was firing only 3 channels before. Now it fires 4 channels in paired mode and 5 channels in semi-seq mode where channels 3 and 5 fire at the same time

pazi88:
Nice!

muuris:
Thanks to iLeezi and pazi88 I've been running sequential 6 cyl (BMW M50) for more than two months now. This definitely should be included in the official fw.

PR #283, opened 2019-11-16
Add H-Bridge DC Motor idle control method

iLeeeZi:
This pull request adds initial support for H-bridge controlled DC idle motor using VNH2SP30 H-Bridge driver board. This control method is used mostly in some VW/Audi cable throttle bodies which uses a DC motor to control throttle flap at idle range. For position control I used a pid control which uses throttle body idle range position sensor as input and open loop pwm tables for setpoint so it is kinda closed loop position control and open loop idle control for now.

The code also adds selectable pins for idle pwm outputs and option to have idle control running even if engine is off for open loop pwm and the new h-bridge control methods.

hbidle

I opened this pull request to show the current state of the code so if anyone has suggestions to do things other way I'm happy to modify it. I don't know how many is going to use this feature but for me it is necessary.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -1822,7 +1822,7 @@ menuDialog = main
         field = "Fan temperature SP",  		 fanSP, { fanEnable }
         field = "Fan hysteresis", 		 fanHyster, { fanEnable }
 
-    dialog = idle_settings,""
+    dialog = main_idle, ""
         field = "Idle control type",     iacAlgorithm
         field = "Enable before start",   idleEngOff,            { iacAlgorithm == 2 || iacAlgorithm == 6 }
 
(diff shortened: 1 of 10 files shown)


VitorBoss:
On idle.ino L398 you are comparing to loops instead of a fixed frequency, you could use the HZ bits instead, see boost pid calls.
When comparing target you miss the == condition when the target are met, both outputs have to be same level(on or off)to stop the motor or it could overheat the driver.
You don't need to set the kp, ki and kd on every time you need to compute the pid. 

Nice PR though.

iLeeeZi:
Good points! I need to change those. Do you mean with the == condition that I need to implement motor braking by setting both direction pins to high or low? just double checking that I understood that right

VitorBoss:
Yes, I meant motor breaking

adbancroft:

PR #284, opened 2019-11-18
small fix for the idle advance settings dialog

iLeeeZi:
allow to change ctps pin settings even if the idle advance is not in use to avoid leaving it accidentally on. ctps pin might be used for other functions in the future so its best to be able to change it all the time.


noisymime:
Closing this as there is now a check in `init.ino` that prevents the CTPS pin being used if the function is not enabled.

PR #288, opened 2019-11-21
add io summary and gauge limits dialogs in TS

brunob45:
Pin assignments are scattered across multiple dialogs, this will add a special dialog containing a read only summary of the pins usage.
Also, RPM and MAP gauge limits will vary from project to project, so I added a window to easily edit these.


muuris:
This is very useful!

noisymime:
Looks good! I've been meaning to get around to something like this for a while, so thanks for putting it together

PR #289, opened 2019-11-22
Prevent priming pulse if flood clear is active

iLeeeZi:
As a big priming pulsewidth is usually required with ethanol in cold climates, you might want to have an option to disable the priming pulse. For example in situations when you need to power cycle multiple times without starting. This pr disables the priming pulse if tps is over the flood clear value when you switch power on.


muuris:
I've also found the need for this.

PR #291, opened 2019-12-10
Replace table3D malloc by static initialization

brunob45:
I made a pair of pragmas to easily fill table3D structs without the need for mallocs.

The advantages are a better control and a better estimate of the RAM usage at compile time.
This works by:
- Creating a "container" struct for every table3D instance, with DECLARE_TABLE3D(name)
- Then linking the container's field to the table3D, with INIT_TABLE3D(name).

The container's name is automatically generated by appending an underscore before the table3D's name (e.g. fuelTable's container is _fuelTable)


noisymime:
A big belated thanks on this! I've been meaning to look at it for some time and have any just gotten around to seriously working through it. 

I'm not going to pull this in directly, but will pushing a commit shortly that uses the heap_alloc you've written, just with some changes to make the updating of the total heap size a little simpler.

brunob45:
Glad I could help in some way :)

PR #295, opened 2019-12-20
Pin mapping for BMW PnP boards

pazi88:
BMW PnP pin mapping to support 6cyl fully seqvential. Basically same as v0.4 pin mapping, but in addition there is correct pins for ign/inj 5-6.


pazi88:
I messed up the old branch for the pin mapping, so here's new one. This pin mapping has been already verified with actual HW.

PR #303, opened 2020-01-15
Add 'p' command to serial3 to enable downloading of pages outside of TS.

ric355:
This change just enables additional data to be downloaded over serial 3. I am using it at the moment for access to fuel, spark, and AFR target tables in my logger/display app.


noisymime:
Closing as this has been implemented separately.

PR #307, opened 2020-01-18
enable teensy canbus interface

Autohome2:
This adds the canbus library for Teensy 3.5 and 4.
input read and out writes activated for initial testing.
for testing only TPS adc is broadcast (hardcoded for test)


Autohome2:
This pr submission is ONLY for the inclusion of the library needed and to test the init code with a simple read and write out, with any necessary changes needed .
The second pr planned to follow after this, will add OBD2 style port to allow a data logger ,dash or diagnostic reader to connect and access the real-time variables available.

noisymime:
If I'm understanding how the Teensyduino libraries hang together, the FlexCAN_T4 library has been included in version 1.49 which was release on Jan 14th. This should therefore be included if you upgrade to that version of Teensyduino and at some point flow through in PlatformIO

noisymime:
Can confirm that this works with the latest Teensy package in both the Arduino IDE and PlatformIO without the need to include the source for FlexCAN_T4. If you pull that out from the src dir I'll commit the PR

Autohome2:
i have updated to teensyduino 1.49 and confirmed the firmware compiles in both arduino ide and platformio correctly

PR #309, opened 2020-01-24
replace BOARD_H macro by a real header file

brunob45:
`include BOARD_H` was kind of a hack, I propose to unhack it.


noisymime:
Thanks for this submission. 

I've given it a lot of thought though and will elect to stick with the current arrangement. The `BOARD_H` is kind of a hack, but it's a neat hack that I don't really see any downsides to. It keeps everything nicely segmented by platform and is flexible enough that everything specific to a platform can be included in one go.

PR #318, opened 2020-02-07
teensy35 FTM1 fixes for Boost and VVT controls

opelpanfan:
There was an issue with teensy35 board boot and vvt timing PWM controls using FTM1 times. Took idle control as an example and updated boost and vvt to act similarly. Tested on SCG_ECU board and can confirm boost solenoid can be controlled also vvt controlled using HC1 output successfully.


noisymime:
Looks good, thanks for submitting!

PR #319, opened 2020-02-07
Calculate the injector open time in tunerstudio

iLeeeZi:
Currently tunerstudio shows the base injector open time and the battery voltage correction values separately. I added an option which calculates a real injector open time from the base open time and the battery voltage correction and leaves the battery voltage correction to 100%.

I understand that this might be a matter of opinion so I left it selectable. For me the calculated value is better as with it the calculations.dash makes more sense. Also the fuel consumption calculations in tunerstudio aren't currently accurate as those use only the base injector open time and this option adds the battery voltage correction to the calculations.

the option disabled:
injOpenCalcOff

and enabled:
injOpenCalcOn


noisymime:
I'm not even sure I'd bother with the option to turn this off. It seems like it really should just be enabled all the time as it gives the actual correct value?

iLeeeZi:
I removed the option to turn it off. I didn't see any need for it myself but I thought that someone might want to use the old method even if it is wrong

PR #320, opened 2020-02-07
Remove variable declaration from header files

brunob45:
By renaming every .ino file to .cpp, the compiler highlights every problem with misplaced variable declaration. Doing so will make unit tesing possible on every part of the code. 
I fixed static and inline functions that couldn't be, like callback functions.
I also added back speeduino.ino, an empty file, so that the project compiles with the Arduino IDE.
I understand that this PR is quite heavy, but this is probably what we'll end up with if we're proactive with unit-testing.
I did not test the effect of these changes on the loop speed, but I'll be doing so this week


noisymime:
Just in case it helps, I do like the idea behind this PR and would certainly like to implement elements of this. If it's something you want to look at still, see how much it can be broken down into smaller PRs as that will make it easier to test and more likely to be merged.

Issue #290, opened 2019-12-07
Secondary serial issue (Mega2560 Serial 3)


muuris:
I'm developing my own incar controller, which need to have two-way communication with Speeduino. 

I've tried to opimize my controllers timing to have the least stress on Speeduino, which tries to read secondary serial rx every 64 loops and polls aux channels at 4Hz. I read the aux polls and queue them, so I can reply with specific intervals, not too quickly one after another. I poll statusdata with 'n' command with 150ms interval at minimum, I can then read loopsPerSec from statusdata and hence try to optimize timings to get my data comms in the middle of the 64 loop period.

The problem is, if I have both status polls and Speeduino AUX polls set up, Speeduino is crashing after some time, secl bounces to 0 and TS communication is lost. Secl goes to zero before byte limit also when TS is not connected (reading stats via serial 3). I thought this was USB or signal issue, but now I noticed that many times (but not every time) just before the Speeduino crashes freeRam shows > 45000 bytes both in TS and serial 3. I've examined the Speeduino code and haven't yet figured why this problem is occuring.

This is now tested with 201911 Speeduino firmware and two separate Mega2560s (one Speeeuino and one of my own).

Autohome2:
It sounds like your external code is not waiting for complete messages from speeduino before requesting more. This is a common issue that has been discussed before.
I'm not sure understand correctly but you are not only requesting data from speeduino but having it read external data too?
Without seeing code for how that is being handled it is very difficult to follow that error.
May I suggest looking at my gpio std firmware to give you some ideas on how to handle this ?

muuris:
Yes, my controller needs Speeduino realtime data (sends 'n' polls) and also acts as external IO (receives 'R' aux channel requests, queues them and replies with specified intervals, one channel request at a time). This problem doesn't come up if only polling realtime data or sending aux channel data.

I've checked the std gpio code, doesn't seem to work as is with 201911 firmware. I got mini gpio working and I think I managed to get the same problem. The gpio codes don't queue Speeduino aux polls but sends them instantly as the request is processed. I found this problematic in 2-way comms if using more than one aux channel, it worked better to queue requests and reply so that Speeduino receives only one channel request between the 64 loop period it checks secondary serial.
 
[...]

PR #321, opened 2020-02-09
Secondary serial requests are sometimes ignored

ric355:
This change fixes a problem where secondary serial requests are sometimes ignored when the command string is more than one byte in length.

The problem is caused by the fact that the secondary serial code makes an assumption that all of the bytes of a multi-byte command (e.g. the 'r' command) are sitting in the serial buffer as soon as CANSerial.available() returns a non-zero value.  This is of course not true because the completeness of the message is very timing dependent. Requesting at a rate of 2 requests per second it can only take around 30 requests or so before the serial buffer is only partially filled. The higher the request rate, the more frequently it happens. This fix delays processing of the command until the whole command string is available.

I suspect this bug is the cause of some problems people have had in the past getting reliable data out of the serial feed. Only people who use 'r' would notice this because the 'A' command is a single byte command, so it is either there in its entirety or not at all and thus does not suffer this issue.

I have tested the fix on the 'r' command and fixed others in the same way, but have not been able to test them all.

### DIFF
--- a/speeduino/cancomms.h
+++ b/speeduino/cancomms.h
@@ -20,6 +20,7 @@ uint8_t canlisten = 0;
 uint8_t Lbuffer[8];         //8 byte buffer to store incomng can data
 uint8_t Gdata[9];
 uint8_t Glow, Ghigh;
+bool canCmdPending = false;
 
 #if defined(__AVR_ATmega1280__) || defined(__AVR_ATmega2560__)
   HardwareSerial &CANSerial = Serial3;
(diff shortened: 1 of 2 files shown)


pazi88:
hmmn, maybe this fixes the problem I have been having with the 'r' command. I have been debugging my own code, but haven't checked if speeduino could be causing the problems.

muuris:
This has some relation to [issue 290](https://github.com/noisymime/speeduino/issues/290). I noticed that 2-way communication to and from Speeduino also suffers from not getting complete commands before processing, then trashing the command header and on next loop there might be the remaining data but without the header it's considered just trash.

PR #325, opened 2020-02-15
Loop timers improvement

VitorBoss:
Recently discussion on a timer check on idle.ino make a need to check LOOP_TIMER on other parts of code, now it is possible by clearing the mask bits on beginning of loop.


VitorBoss:
As this just bring no real change I'm closing it

PR #328, opened 2020-02-22
fix building with spi flash as config storage

Tjeerdie:
Hi All. I just fixed building speeduino with spi flash as storage. Still not working as it should but it builds. More fixes to come on this


VitorBoss:
Why you changed the USE_SPI_EEPROM?

On `begin(pb0)` uoy should use a `#if defined(CORE_STM32)` to avoid errors on other boards.
Lat changes on this was made by @noisymime to use on Teensy boards.

noisymime:
There's been a few different flags used for the SPI EEPROM stuff and we need to make sure we're all on the same page. The one to use going forward is `USE_SPI_EEPROM`

Tjeerdie:
I used the spi flash storage define because the spi eeprom is misleading. It is no spi eeprom chip used in this code. If in the future someone connects a spi eeprom chip this wil need to change again. This concept it is spi flash used as eeprom emulation. two different concepts. Please advise i will change back to spi eeprom for the black boards.

Tjeerdie:
> Why you changed the USE_SPI_EEPROM?
> 
> On `begin(pb0)` uoy should use a `#if defined(CORE_STM32)` to avoid errors on other boards.
> Lat changes on this was made by @noisymime to use on Teensy boards.

Because the pin now changed from PB0 to hardcoded use pin 6. That does not work in the black board. I would appreciate it if changes are made to make the CS pin more generic and to be used for other boards. But please do not change pins on the storage hard coded so it breaks the stm32 black boards.

VitorBoss:
That is because you missed to call begin() before, check https://github.com/noisymime/speeduino/commit/106b75509dc0edf9743cb969ed31261881d96eb8

Tjeerdie:
> That is because you missed to call begin() before, check [106b755](https://github.com/noisymime/speeduino/commit/106b75509dc0edf9743cb969ed31261881d96eb8)

yes i see what you guys did with the call to EEPROM.begin(CSPIN);. But this does not work for the STM32 because EEPROM is not instantiated yet when the call is made on the STM32 build. Apparently the order of building is different for STM32 and Teensy

VitorBoss:
Than you should use this instead:
```
    if(!SpiFlashAvialable){
    #if defined(ARDUINO_BLACK_F407VE)
      begin(PB0);
    #else
      begin();
   #endif
    }
```

Tjeerdie:
> > That is because you missed to call begin() before, check [106b755](https://github.com/noisymime/speeduino/commit/106b75509dc0edf9743cb969ed31261881d96eb8)
> 
> yes i see what you guys did with the call to EEPROM.begin(CSPIN);. But this does not work for the STM32 because EEPROM is not instantiated yet when the call is made on the STM32 build. Apparently the order of building is different for STM32 and Teensy

This is the exact reason i did call the begin() the first time the read function is called. I know this is a hack and not pretty. i changed this begin(); call in de read function to incorporate PB0 for that reason. Else it starts with the default pin 6 if no argument is given.

Tjeerdie:
> Than you should use this instead:
> 
> ```
>     if(!SpiFlashAvialable){
>     #if defined(ARDUINO_BLACK_F407VE)
>       begin(PB0);
>     #else
>       begin();
>    #endif
>     }
> ```

That still does not work unfortunately. If i change the define to read USE_SPI_EEPROM the code tahat breaks it becomes.
#ifdef USE_SPI_EEPROM
  //We need to send the flash CS (SS) pin if we're using SPI flash. It cannot read from globals.h
  EEPROM.begin(pinSPIFlash_CS);
#endif

Tjeerdie:
Hmm there is a problem i cannot solve at the moment. All the definitions of the Flash CS pin are in the board initialization function but before initBoard(); is called a call is made to loadConfig(); And  that needs the CS pin to be defined for succeeding. (what board is used is stored in the EEPROM). So this always fails and is a deadlock. 

@noisymime Please advise how to proceed and where to put the definitions. i can make something i think is ok but that does not mean you agree :-D.

PR #329, opened 2020-02-24
new pin definitions board Tjeerd and updated spi flash as eeprom CS pin selection to use 

Tjeerdie:
I updated the pin definitions for my board (Also updated the default pin configuration of the STM32F407VE). 

Added a way to change the chip select pin (CS pin) for the SPI flash using a flag set in the compile environment.


Tjeerdie:
@noisymime Do you have any interest in this?

PR #332, opened 2020-02-29
[AE] Introduce new method

VitorBoss:
Introduce ReqFuel % adder to the final PW calculation on Acceleration Enrichment.

~~Include some new timer flag definitions.~~


LinkBR:
Waiting merge

VitorBoss:
@noisymime Any luck on this?

Just to clarify, the actual mode is VE based and, depending on value it should not be enough. This PR add an option to do AE based on req_fuel and have a faster response time than actual.

PR #338, opened 2020-03-18
STM32F407 fixes

iLeeeZi:
These should fix 32bit timers missing/double pulses and SRAM backup write/read problems. Some board configurations did break because of the bluepill/blackpill "hacks" so those are now checked only with bluepill/blackpill boards.

### DIFF
--- a/speeduino/board_stm32_official.h
+++ b/speeduino/board_stm32_official.h
@@ -20,8 +20,8 @@
 */
 #define PORT_TYPE uint32_t
 #define PINMASK_TYPE uint32_t
-#define COMPARE_TYPE uint16_t
-#define COUNTER_TYPE uint16_t
+#define COMPARE_TYPE uint32_t
+#define COUNTER_TYPE uint32_t
 #define micros_safe() micros() //timer5 method is not used on anything but AVR, the micros_safe() macro is simply an alias for the normal micros()
 #if defined(SRAM_AS_EEPROM)
     #define EEPROM_LIB_H "src/BackupSram/BackupSramAsEEPROM.h"
--- a/speeduino/board_stm32_official.h
+++ b/speeduino/board_stm32_official.h
@@ -71,8 +71,8 @@ extern "C" char* sbrk(int incr);
 ***********************************************************************************************************
 * Schedules
 */
-#define MAX_TIMER_PERIOD 65535*2 //The longest period of time (in uS) that the timer can permit (IN this case it is 65535 * 2, as each timer tick is 2uS)
-#define uS_TO_TIMER_COMPARE(uS) (uS>>1) //Converts a given number of uS into the required number of timer ticks until that time has passed.
+#define MAX_TIMER_PERIOD 65535*4 //The longest period of time (in uS) that the timer can permit (IN this case it is 65535 * 4, as each timer tick is 4uS)
+#define uS_TO_TIMER_COMPARE(uS) (uS>>2) //Converts a given number of uS into the required number of timer ticks until that time has passed.
 
 #define FUEL1_COUNTER (TIM3)->CNT
 #define FUEL2_COUNTER (TIM3)->CNT
--- a/speeduino/board_stm32_official.h
+++ b/speeduino/board_stm32_official.h
@@ -71,8 +71,8 @@ extern "C" char* sbrk(int incr);
 ***********************************************************************************************************
 * Schedules
 */
-#define MAX_TIMER_PERIOD 65535*4 //The longest period of time (in uS) that the timer can permit (IN this case it is 65535 * 2, as each timer tick is 2uS)
-#define uS_TO_TIMER_COMPARE(uS) (uS>>2) //Converts a given number of uS into the required number of timer ticks until that time has passed.
+#define MAX_TIMER_PERIOD 65535*2 //The longest period of time (in uS) that the timer can permit (IN this case it is 65535 * 2, as each timer tick is 2uS)
+#define uS_TO_TIMER_COMPARE(uS) (uS>>1) //Converts a given number of uS into the required number of timer ticks until that time has passed.
 
 #define FUEL1_COUNTER (TIM3)->CNT
 #define FUEL2_COUNTER (TIM3)->CNT
--- a/speeduino/board_stm32_official.h
+++ b/speeduino/board_stm32_official.h
@@ -51,18 +51,20 @@ void initBoard();
 uint16_t freeRam();
 extern "C" char* sbrk(int incr);
 
-
-#ifndef PB11 //Hack for F4 BlackPills
-  #define PB11 PB10
-#endif
-//Hack to alow compile on small STM boards
-#ifndef A10
-  #define A10  PA0
-  #define A11  PA1
-  #define A12  PA2
-  #define A13  PA3
-  #define A14  PA4
-  #define A15  PA5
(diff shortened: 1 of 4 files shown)


VitorBoss:
Speeduino code uses timers in microseconds format, tick is timer source clock, the overflow value should be kept same on all timers to scale. Allowable values for overflow are any value from 0 to max

pazi88:
@VitorBoss problem is that with current implementation the injector and ignition pulses are randomly missing or too short and the stm32 is useless to control engine because of that. This PR fixes the random problems in inj/ign pulses.

iLeeeZi:
And it looks like this: timer3

There is something weird with it and I agree that my "fix" might not be the correct method but at least it works fine with it. This only happens with the 32 bit counter timers.

VitorBoss:
This problem is only on official core, on STM32GENERIC is just fine. I've forgot, the microseconds is set later when the timer clock is divided, the clock should be TICK, could you test with 16 bits please? Os STM32GENERIC is all 16bit and works just fine.

This misconception about timer mode is because HardwareTimer is different from the STM32GENERIC. Nice catch

iLeeeZi:
Yeah I forgot to add that these fixes are only for the official core. I tested with 16bit max overflow but it is same as in the screenshot above. HardwareTimer library says that it handles all timers as 16bit but there seems to be some differences.

VitorBoss:
The PWM pins output double of configured frequency, is just a copy-paste error, could you fix that by changing board_stm32_official.ino#L75 from 
`Timer1.setPrescaleFactor(((Timer1.getTimerClkFreq()/1000000) * 4)-1);   //4us resolution`
To
`Timer1.setPrescaleFactor(((Timer1.getTimerClkFreq()/1000000) * 2)-1);   //2us resolution`
Thanks

iLeeeZi:
Of course! Is there any benefits or disadvantages if injection and ignition timers are also changed to the 2us resolution?

VitorBoss:
With 2uS you get only a better accuracy, but for that change you have to change the `MAX_TIMER_PERIOD` and `uS_TO_TIMER_COMPARE(uS)` to match STM32GENERIC definitions

iLeeeZi:
Thanks! Changed those too.

Tjeerdie:
Hi ILeeeZi

I fixed those double pulse problems a couple of months ago on my black STM32F407VE boards. I tested the timing extensively with some digital analyzer and found to be correct. I will try find out what changed between my last commit and this to break the code.

I have some issues with this commit

1. It assumes timer 2 is always a 32 bit timer. This does not hold for every STM32F4 board I assume? Also did you check the result with this? As the overflow setting of 0xFFFF for a 32 bit or 16 bit timer should be equally correct. (If the timers are configured to reset to zero on overflow as this is supposed to happen)

I think there is another issue but i am not totally sure. The code assumes that the overflow always happens at 0XFFFF for every timer (16 bit). If the pulse end is set to a point that is over 0xFFFF it will subtract 0xFFFF from that number. example:
1. Pulse end wil be calculated to happen at 100us from now in code.  
2. Timer counter is now at 0X65500 
3. Pulse end interrupt is now set at  65500 + 100/2(usperpulse) = 65550 
4. Code will put pulse end at 65550 - 65535 = 15us (assuming 16 bit timers)
5. This will break pulses around the overflow (when start is before overflow and end after overflow) when timers overflow of 32bit timers is set at 0XFFFFFFFF 

I found this problem when using larger pulse widths and then the problem happens more frequently. Or is this problem mitigated in some other way?

Tjeerdie:
> ~Speeduino code uses timers in microseconds format,~ tick is timer source clock, the overflow value should be kept same on all timers to scale. Allowable values for overflow are any value from 0 to max

overflow in the official core is not set in register format (bits) it is set in useconds. Therefore not set to 0XFFFF but to max timer periode in us. I know it is strange and it took me quite a while to figure out why the double pulses occurred.

Tjeerdie:
> Of course! Is there any benefits or disadvantages if injection and ignition timers are also changed to the 2us resolution?

There is very nasty down side of the 2uS resolution. That will stick its ugly head up with very low RPM values. (slow running motors). Because there can never be a ign or inj pulse start scheduled longer than 0XFFFF*2 Us in advance. This can give trouble if during cranking to running the RPM < 457.77.

With 4uS resolution this trouble only starts with RPM < 228.88. And is therefore much more realistic number. 

So please do not change. There have been people with WV boxer motors that had trouble with this i think.

iLeeeZi:
Thanks for looking for the issues! I will revert those changes and check if I get it working.

iLeeeZi:
@Tjeerdie I did try your suggestions but double/missing pulses still happens on the TIM2 and TIM5. It seems to happen at every 65535us * timer resolution so with 4us resolution it happens at every 262140us. It is similar what you described about the 2us resolution downside but it is only happening on the 32 bit timers regardless of the RPM.

Tjeerdie:
> @Tjeerdie I did try your suggestions but double/missing pulses still happens on the TIM2 and TIM5. It seems to happen at every 65535us * timer resolution so with 4us resolution it happens at every 262140us. It is similar what you described about the 2us resolution downside but it is only happening on the 32 bit timers regardless of the RPM.

Ok, i will try and investigate this weekend, I should have the time now with all the social distancing going on :-D. 

Maybe i "fixed" this problem a couple of months ago by just not using any 32 bit timers and not making an PR for this. I do remember seeing the same thing but i cannot remember how and when i fixed it. I will take a look at my personal repository to see what i did try around that time.  

Setting the overflow to 32 bits (or 0xFFFFFFF in registers) should not work at all i assumed. The pulse scheduler expects an overflow on 16 bits. I do not understand what is going on and why it even works kind of with the over flow set to 65535us on 16 bit timers. And not on 32 bit timers.

Tjeerdie:
I only now see you changed the setOverflow type to TICK_FORMAT! Sorry man my mistake to overlook this. This is/should also be correct! Can you try one last thing? Setting ALL timers (also the 32bit timers TIM2 and TIM5) with:

TimerX.setOverflow(0xFFFF, TICK_FORMAT);

That should fix the double pulses on both 32bit and 16bit timers

Tjeerdie:
Hi all

Some small update on this.

TimerX.setOverflow(0xFFFF, TICK_FORMAT); works for 16 bit timers not for 32 bit timers
TimerX.setOverflow(MAX_TIMER_PERIOD, MICROSEC_FORMAT); works for 16 bit timers not for 32 bit timers

TimerX.setOverflow(0xFFFFFFF, TICK_FORMAT); Does not solve the problem for 32 bit timers. There are no more pulses on the channels for that timer because scheduler does not work. 

I have the impression that callbacks for channel 1..4 are also called on overflow for 32 bit timers. I have no clue why yet. Still investigating.

iLeeeZi:
Thanks for the update! I did check and the bug indeed seems to happen even with 0xFFFFFFF and it propably still happens with the 0xFFFFFFFF as you said so it just hides the issue.

Tjeerdie:
> Thanks for the update! I did check and the bug indeed seems to happen even with 0xFFFFFFF and it propably still happens with the 0xFFFFFFFF as you said so it just hides the issue.

Just re-read my previous post. I edited it some more. It is all correct but i did not understand why in the beginning, now i finally do.

iLeeeZi:
@Tjeerdie I appreciate your taking the time to check what really caused the problem. That explanation was good.

Looks like whole STM32F4 series has 32 bit TIM2 and TIM5 so I added a check for that and changed all timers back to 4us resolution. Later those could be changed to whatever works the best but for now I feel it is good enough.

PR #342, opened 2020-03-23
[ASE] Transition time to disable

VitorBoss:
Configurable time in 0.1S resolution


muuris:
Great addition! Especially when the ASE needs to have a big value, without this fade off-transition the engine easily either coughs/runs rough, or runs unnecessarily rich for longer than needed.

VitorBoss:
ASE

noisymime:
You've setup a configuration variable `ASEValue` but then appear to be using this to track a status value. `ASEValue` should be on `currentStatus` rather than as a map variable as it's not part of the tune itself.

noisymime:
Also:
`volatile int runSecsX10;`

Should be a uint16_t. You're checking for it to overflow at 65535, but have declared it as a signed int (Which will overflow at half that)

VitorBoss:
Hope didn't miss what you mean with this commit

Issue #343, opened 2020-03-24
O2 correction may be disabled when ignition count cycles over


bera-n:
There's a possibility that O2 corrections will stay disabled when ignition count rolls over to 0. This is because the target cycle is not checked or updated if o2corr is disabled by any engine condition (AFR, TPS temps etc)
Could be fixed by checking/updating the target cycle first and then checking engine conditions.

On a 4cyl engine at 2000rpm this might happen every ~16min.

bera-n:
Well my fix didn't work as expected as the correction was reset constantly. New merge request with a fix. Tested on running engine and seems to work.

PR #344, opened 2020-03-24
Check and update ignition count target before checking other conditio…

bera-n:
…ns to avoid peramently disabling ego correction in case ignition count rolls over while ego correction is disabled.


bera-n:
#343

noisymime:
Looks good. Just tried it out and it all seems to be working around the overflow point.

Thanks for the submission

bera-n:
@noisymime Realised this isn't working as it keeps resetting the corrections whenever the igncycle isn't due for an update. I've got an (untested, won't be able to in a few days) fix on the branch in my Github repo. How should I proceed? Make a new pull request? https://github.com/bera-n/Speeduino/tree/dont-miss-egocorr-at-igncount-rollover

Sorry for breaking the code :(

PR #345, opened 2020-03-25
Configurable CS pin flash chip

Tjeerdie:
A couple of changes in this PR. This should have been three different PR's sorry for that.

1. This PR makes it possible to configure the spi flash chip CS pin during build. It is set by using a macro. The macro is set to the pin number used as CS pin. Tested on an STM32F407 board. 
2. Changed some pins around for my current board design. This will most definitely change again in the future.
3. Fixed building for the STM32F407 with official core and the recent CAN bus changes. This needs proper fixing in the future. But that is something to consider when the CAN implementation is more mature.


noisymime:
I'm happy to pull this one in having reviewed it and tested, but from what I can see the main intent of it (Having a CS pin that is configurable at build time) has actually been disabled. 

Is that still right?

Tjeerdie:
Hi Josh

Thanks for looking at it. The last time i checkt it was working the way it
supposed to. I wil check this evening and let you know.

Tjeerd

noisymime:
It's entirely possible I was looking at this incorrectly. The line I was specifically looking at was in init.ino: https://github.com/noisymime/speeduino/pull/345/files#diff-32cb6304c5b6f1df24d90e9351789241R2237

Looking again that I can see that the `begin()` is called further down in another file anyway, so you're probably right.

Tjeerdie:
Hi Josh. begin() must be called before the configuration is read from the
EEPROM. This calls begin() once during the first EEPROM read. That should
always work no mater form what location you access the EEPROM first.. The
previous location was after the EEPROM was accessed and therefore did not
work correctly.


Op do 16 apr. 2020 om 09:05 schreef Josh Stewart <notifications@github.com>:

Tjeerdie:
So to confirm, this is correct indeed.

PR #347, opened 2020-03-29
[PW Math] Introduce fixed MAP multiply

VitorBoss:
This should allow former megasquirt users to use their VE on speeduino.


pazi88:
This addition would make life much easier. I have so many MS tunes I could use as starting point.

PR #350, opened 2020-04-02
Add MicroEFI pinout

NicoEFI:
Add MicroEFI pinout
add MPXH6250 sensor pressure
add "Not coonnected" sensor pressure if MAP is not used


adbancroft:
ClosClosing - PR is too old to merge. Please re-submit based on latest master.

PR #351, opened 2020-04-02
Launch modification

NicoEFI:
Launch modification
Can use a potentiometer for a variable launch.
Because the road or the track can be wet or dirty.
Soft launch limit is modified by a zone  (delta) under Hard launch limit.


VitorBoss:
Why it can't be a switch?

NicoEFI:
There is a switch AND a potentiometer.
With the potentiometer you can change the launch hard limit rpm.
If you do not want this to be variable, indicate the same rpm value in Launch hard ReV Mini and in Launch hard ReV Maxi

VitorBoss:
The fail is because LEDpixel variable.

NicoEFI:
Thanks VitorBoss

noisymime:
I would STRONGLY suggest looking at a way of doing this without changing the table size. Altering table sizes mean that all existing tunes need to be updated to reflect the change and it can cause serious problems if not done correctly. You also need to change the storage routines otherwise the additional values won't be kept in EEPROM correctly. 

Is there anyway of getting the variables you need in another table?

adbancroft:

PR #353, opened 2020-04-03
[DFCO] Add a minimum coolant

VitorBoss:
On pure ethanol and E85 the engine need a lot fuel to return from DFCO when cold.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -336,7 +336,12 @@ page = 1
       injAngRPM     = array,  U08,      95, [4],   "RPM"        100,       0.0,   100,    10000,    0
       idleTaperTime = scalar, U08,      99,    "S",         0.1,       0.0,   0.0,    25.5,      1
       dfcoDelay     = scalar, U08,      100,   "S",         0.1,       0.0,   0.0,    25.5,      1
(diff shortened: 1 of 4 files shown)


muuris:
And also higher cold start idle rpms require this to avoid humping.

noisymime:
Thanks for another great submission Vitor.

Just reviewing, in `corrections.ino` should the temperature check include the offset:

Eg: `( currentStatus.coolant >= configPage2.dfcoMinCLT - CALIBRATION_TEMPERATURE_OFFSET)`

?

VitorBoss:
Miss that, that offset should be on update.ino when I set the value? For °C and °F the offset changes right?

muuris:
Internally in Speeduino data is always °C, the °F scaling is done in TS end (TS always sends the byte value which should be the same for °C/°F of same temperature)

muuris:
I noticed dfcoMinCLT is defined as array (which it isn't) instead of scalar in .ini

noisymime:
@VitorBoss You've got the calibration offset adding rather than subtracting. It should be something like:

`(currentStatus.coolant > (int)(configPage2.dfcoMinCLT - CALIBRATION_TEMPERATURE_OFFSET))`

VitorBoss:
Fixed and squashed

PR #356, opened 2020-04-10
Add Hardware ID functions

VitorBoss:
Feature to use on future speedLoader


VitorBoss:
As you mentioned to do this in a different way, I'm closing this.

VitorBoss:
As you mentioned to do this in a different way, I'm closing this.

PR #358, opened 2020-04-11
Display IAC pwm actual value in TS

muuris:
IAC PWM (open and closed loop) value is now shown with 1 % precision instead of 2 %.


noisymime:
Looks great, thanks.

PR #359, opened 2020-04-14
Acceleration enrichment cold temperature adjustment

muuris:
This adds possibility to increase AE amount when engine is cold. Needed ie for E85 and stone cold engine. Cold modifier is tapered between two temperature points. Default values are set so this adjustment won't apply (so no change for those who won't notice this feature).

image


noisymime:
A really great example of a nice neat PR. Its function is specific and atomic, it's got a good description and screenshot and the icing on the cake is that you've put in the required sections in `updates.ino`. 

Thanks for the PR

PR #364, opened 2020-04-21
Cranking enrichment to ASE (or run) taper

muuris:
This adds possibility to taper cranking enrichment smoothly to ASE value instead of rapid drop.

With large cranking enrichment the taper allows to have less ASE without stalling the engine right after it starts. The functionality and use case of this goes hand in hand with IAC taper from crank to run.

Valid range is from 0 to 25,5 seconds. If ASE has less duration than this, then this will just taper to 100 %.

image

An example log of this from my car, the vertical line is on the moment cranking ends in both logs. First without taper, second with it.
image

### DIFF
--- a/speeduino/corrections.ino
+++ b/speeduino/corrections.ino
@@ -189,7 +189,7 @@ uint16_t correctionCranking()
   }
   
   //If we're not cranking, check if if cranking enrichment tapering to ASE should be done
-  else if ( (uint16_t) (currentStatus.runSecs*10) <= configPage10.crankingEnrichTaper)
+  else if ( (uint32_t) runSecsX10 <= configPage10.crankingEnrichTaper)
   {
     crankingValue = table2D_getValue(&crankingEnrichTable, currentStatus.coolant + CALIBRATION_TEMPERATURE_OFFSET);
     crankingValue = (uint16_t) crankingValue * 5; //multiplied by 5 to get range from 0% to 1275%
--- a/speeduino/corrections.ino
+++ b/speeduino/corrections.ino
@@ -181,11 +181,18 @@ Additional fuel % to be added when the engine is cranking
 uint16_t correctionCranking()
 {
   uint16_t crankingValue = 100;
-  //if ( BIT_CHECK(currentStatus.engine, BIT_ENGINE_CRANK) ) { crankingValue = 100 + configPage2.crankingPct; }
+  //Check if we are actually cranking
   if ( BIT_CHECK(currentStatus.engine, BIT_ENGINE_CRANK) )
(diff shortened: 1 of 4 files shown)


speeduino/corrections.ino
```diff
@@ -181,11 +181,21 @@ Additional fuel % to be added when the engine is cranking
...
   uint16_t crankingValue = 100;
-  //if ( BIT_CHECK(currentStatus.engine, BIT_ENGINE_CRANK) ) { crankingValue = 100 + configPage2.crankingPct; }
+  //Check if we are actually cranking
   if ( BIT_CHECK(currentStatus.engine, BIT_ENGINE_CRANK) )
   {
     crankingValue = table2D_getValue(&crankingEnrichTable, currentStatus.coolant + CALIBRATION_TEMPERATURE_OFFSET);
     crankingValue = (uint16_t) crankingValue * 5; //multiplied by 5 to get range from 0% to 1275%
+
+  //If we're not cranking, check if if cranking enrichment tapering to ASE should be done
+  } else if ( (uint16_t) (currentStatus.runSecs*10) < configPage10.crankingEnrichTaper)
```
VitorBoss, line 191: You could use `else if ( runSecsX10) < configPage10.crankingEnrichTaper )` as runSecsX10 is used above, it will speed up the line.

VitorBoss:
Nice addition.

muuris:
But runSecsX10 will roll over eventually and it would do the tapering again when it shouldn't. runSecs will stay maxed out.

VitorBoss:
But ro runSecsX10 overflow will take 6553.5 seconds or 18H12M+

pazi88:
Usefull addition once again 👍

speeduino/corrections.ino
muuris: As discussed, runSecsX10 isn't an option here as it will overflow after 109 minutes.

VitorBoss: Just checked here, the runSecsX10 will run until 65535 and wont overflow as the code is: 
`if ( (BIT_CHECK(currentStatus.engine, BIT_ENGINE_RUN) ) && (runSecsX10 < 65535) ) { runSecsX10++; }`

Or I'm missing something and did that code wrong?

muuris: The code is fine
```
if ( (BIT_CHECK(currentStatus.engine, BIT_ENGINE_RUN) ) && (runSecsX10 < 65535) ) { runSecsX10++; }
    else { runSecsX10 = 0; }
```
But at 65535 the if condition isn't true, so the else will put it into zero.

PR #367, opened 2020-04-24
STM32F407 pin mapping for BMW-PnP board

pazi88:
This PR adds STM32F407 pin mapping for BMW-PnP board definition when it's used with my F407 Arduino Mega replacement board. This Mega replacement F407 board can be used with v0.4 boards also so it can be used as upgrade in those. But I didn't touch the v0.4 board mapping at all. Only the BMW-PnP board.

There is also now Baro and VSS pins mapped, because I have those available in the BMW-PnP boards.


pazi88:
This isn't actually needed,  because the mega2560 pin mapping also works

PR #368, opened 2020-04-26
Added option for soft limit to also be able to retard by x-amount of degrees

pazi88:
The soft limiter has been absolute timing advance value, but that doesn't work very well. For example lets say you have have engine that at 100kpa  might need about 25 degrees timing at WOT at rev limiter. To get nice behavior of soft limiter, it will require 5 degrees of retard so 20 degrees of absolute timing. But if that engine is boosted the needed advance might be something like 5 degrees at 250kpa around rev limiter RPM. Now that you have the 20 degrees of advance set for soft limiter that works for off boost, but now you are hitting rev limiter at 1.5bar of boost and speedy suddenly uses 20 degrees timing at soft limiter and maximum that works is 5 degrees = broken engine. I have heard few engine failures caused by this exact thing. So it would be much safer to change this to retard by x-amount of degrees instead of absolute timing advance. Now when soft limiter retard is set to for example 5 degrees, it will retard 5 degrees no matter what is the current advance and behaves as nice and safe soft limiter. This PR does that with simple change, which shouldn't break anything in the code.

There has been also request to allow negative timing advances for soft limiter to make bangs and pops while your are at rev limiter. Even thought that's not the wisest thing to do, this change also allows that.


muuris:
This is a needed change indeed.

noisymime:
Looks good, thanks for the changes.

PR #370, opened 2020-04-30
Add "Half sync" indicator

pazi88:
This PR adds extra "Half Sync" indicator for situations where the trigger setup requires primary and secondary triggers, but only primary trigger is detected. Like missing tooth trigger in sequential setup. Previously it has been difficult to debug if there is both primary and secondary triggers missing or only the secondary trigger. With this PR, it's easy to see from the yellow "Half Sync" indicator that the primary trigger is ok, but secondary trigger is missing and that's the reason for no RPM. This half sync indicator isn't in use at all if the setup only requires primary trigger.

image

I added the half sync for three different trigger patterns. There might be others that this could be added too, but I don't know those well enough to do that.


muuris:
There is definitely use for this!

noisymime:
There might be some confusion around the 'half sync' terminology with this usage. In most other systems (at least as far as I understand it) 'half-sync' typically refers to it having crank angle, but not cam phase, which means it can run in wasted spark/paired mode but not sequential. 

This PR is very similar to that, but not exactly the same. For example, in a dual wheel setup if you have just the crank signal you still can't even run the engine. It's still no sync at all, even if the crank signal is working fine. 

That said, I do think there is value in showing this info, but maybe with different terminology. What about having 2 new indicators, one each for crank and cam? That way you're not conflating the presence of a signal with actually getting sync.

pazi88:
Well that was my overall goal to have it running wasted spark/paired mode when there is only crank signal. But I thought this could be used as debug tool for others too and separated this to smaller PR without that first. If you think it's better to have this only for triggers that can do wasted/paired from crank only, I can do that. Actually is there even others currently there that can do that other than missing tooth?

pazi88:
@noisymime I did now remove this from the other trigger patterns and left only for the missing tooth, which I'm going to add the actual Half Sync functionality in another PR later once I have verified that in car. Is that ok for you?

runesmdk:
Will it be possible to have a semi-seq startup that reverts to full sync when cam phase is detected? Just might help startup that little bit in some situations.

pazi88:
> Will it be possible to have a semi-seq startup that reverts to full sync when cam phase is detected? Just might help startup that little bit in some situations.

That will be next step and was the actual purpose of this PR. But I though doing this indicator first and adding the actual semi-seq/wasted spark functionality later when this is merged.

PR #372, opened 2020-05-05
EEPROM emulation on internal flash for STM32F407

Tjeerdie:
Hereby the PR to incorporate internal flash memory of the STM32F407xx as EEPROM emulation. Also some changes to the SPI flash as EEPROM to be able to flash the first configuration without problems.

It uses the last 512kb of the flash for EEPROM on STM32F407.
It uses the last 1024kb of the flash for EEPROM on SPI flash  
It incorporates wear leveling.  

The whole flash as EEPROM has been cleaned so that it now makes use of the same base code for both the internal and SPI flash implementation.


noisymime:
Looks good! I'll have to dig out my black board and play around, but this seems fairly harmless to the rest of the code, so happy to pull in.

PR #373, opened 2020-05-06
Add define pins for other functions

Tjeerdie:
Add new pinDNU definitions. (pin Do Not Use)

If a pin is defined in the pinDNU array it is not used for normal speeduino in and outputs but is also not available for other aux definitions. This is for pins like CAN, Chip select, RX, TX, USB pins etc. 

This is useful to not to brick/break your board because usb comms fail by selecting a AUX pin in TS on the USB port. This can then not be recovered easily. 

The pins are set in the board definition because it obviously dependents on what hardware is connected to a particular board.

Also i added all the relevant pinDNU definitions to my own experimental board.


Tjeerdie:
I assume this is never going to be looked at again, so i will close the PR next week if there is no further interest.

Tjeerdie:
I close this as there is no interest. I have seen multiple people have issues with boards being soft bricked by wrong pin settings that would have been prevented by this PR.

PR #374, opened 2020-05-06
STM32F407 32bit timer issue fix (new)

Tjeerdie:
Because of issues with 32bit timers on the STM32F407 (again) a different fix 

1. Reverted the overflow to 16 bit(0xFFFF) for all timers again.
2. Explicitly cast all writes to the Compare registers to uint16_t 
3. Set the battery voltage to sane value on initialization

This works on the STM32 and should not create (much) more overhead for the Arduino MEGA because normally this is already done implicitly when writing to the 16 bit compare registers. This is NOT tested on the MEGA very thoroughly. But in theory this should not in anyway influence performance or function. 

It also set the battery voltage to a default value when board starts. So before a sensible value is read from the ADC this is 0. The dwell correction is creating the ignition spurious pulses direct after reset of the  board for around 250ms when battery voltage is 0. (When "flying start" the engine)

### DIFF
--- a/speeduino/board_stm32_official.h
+++ b/speeduino/board_stm32_official.h
@@ -20,8 +20,8 @@
 */
 #define PORT_TYPE uint32_t
 #define PINMASK_TYPE uint32_t
-#define COMPARE_TYPE uint32_t
-#define COUNTER_TYPE uint32_t
+#define COMPARE_TYPE uint16_t
+#define COUNTER_TYPE uint16_t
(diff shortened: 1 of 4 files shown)


Tjeerdie:
Hmm ok, Without this PR 32bit timers on the STM32 are still not working the first 51 seconds after boot. Hopefully someone will find a better fix.

noisymime:
Tested OK on mega2560

PR #375, opened 2020-05-07
Bug fix: Pointers to Idle pins/port must always be initialized. 

Tjeerdie:
The pointers to IDLE pins must always be initialized even if the PWM CL or PWM OL are not selected. This is general good practice to not leave pointers uninitialized.  

This is most important on STM32F4 because when the IDLE interrupts run (they do run after a EEPROM reset, all 0xFF) it access the pins/port pointers uninitialized.   

Uninitialized pointer point to Internal flash memory and trying to write to these uninitialized will break the access to the internal flash as EEPROM emulation. Because writing the internal flash when not initiated properly will put flash access in error state.

### DIFF
--- a/speeduino/idle.ino
+++ b/speeduino/idle.ino
@@ -23,6 +23,12 @@ void initialiseIdle()
   //By default, turn off the PWM interrupt (It gets turned on below if needed)
   IDLE_TIMER_DISABLE();
 
+  //pins must always be initialized. 
+  idle_pin_port = portOutputRegister(digitalPinToPort(pinIdle1));
+  idle_pin_mask = digitalPinToBitMask(pinIdle1);
+  idle2_pin_port = portOutputRegister(digitalPinToPort(pinIdle2));
+  idle2_pin_mask = digitalPinToBitMask(pinIdle2);
+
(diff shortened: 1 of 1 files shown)


noisymime:
Purely out of interest more than anything, why do the interrupts fire when the EEPROM is reset? That seems like dangerous behaviour.

I'll also look at adding a check in the interrupt function that only modifies those pins if those modes are enabled.

Tjeerdie:
Hi

I am not too sure why it is running. It should obviously not but it still
is running at every ~250ms (overflow 0xFFFF with a 4us clock). I had found
the issue when doing work in the internal flash EEPROM emulation. The error
flag for reading/writing flash where set every ~250ms after resetting them.
So when i solved it i could go on with the task at hand, fixing the
internal EEPROM emulation. It is on my TODO list for things to check this
more in depth.

It might be as simple that the IDLE_TIMER_ENABLE(); is called somewhere, or
that the IDLE_TIMER_DISABLE(); is not working correctly. (And none of the
disable timer functions are working if that is the case)


Op vr 8 mei 2020 om 00:28 schreef Josh Stewart <notifications@github.com>:

PR #378, opened 2020-05-11
Add Idle up output

diy-efi:
Implement functionality to allow a defined output pin to toggle (normal or inverted) based on the idle up input pin.


lsdlsd88:
would be nice to allows to set a configurable delay from the moment idle-up pin changes state to the output activation? would make it easier on the engine to wait for the idle to stabilize on the higher value before switching on stuff like the AC compressor and FAN.

diy-efi:
> would be nice to allows to set a configurable delay from the moment idle-up pin changes state to the output activation? would make it easier on the engine to wait for the idle to stabilize on the higher value before switching on stuff like the AC compressor and FAN.

That is my intention for the future, I just wanted to get the initial code in that I had written already.
Another potential expansion is disable AC when TPS is over X%, so if you are WOT it can kill the AC fan and clutch.

PR #379, opened 2020-05-14
Antilag Functions

nicola-reat:
Hi everyone, I have finally completed the Antilag functionality. 
Now this new version of code has the possibility to enable the antilag functionality by means of a switch. If we want the antilag just enable the function on tunerstudio and then we can activate it when we want with the switch. On tunerstudio we have to configure 3 options: The maximum opening of the tps (When the antilag reaches this value it is automatically disabled), The advance / postponement of the ignition and the quantity of fuel to add when ALS is active. The system works only when the engine is running and not when it is in the cranking phase. Soon I will also add a led when the antilag is on. In addition to this I will add other features :D

P.S: The code was only tested in the laboratory using the simulator. As soon as the text on the car I write here: D. Before then use it with caution

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,13 @@
 ##### A low cost, DIY friendly Engine Management System (ECU) based on the Arduino framework
 </div>
 
+## ANTI-LAG FUNCTIONALITY
+I added the antilag to the speeduino code. On tunerstudio there is a modal where you can configure the anti lag. There are 3 options to configure:
+1) Max TPS before disabling
+2) Static Ignition Timing while active
+3) Amount of fuel to add while active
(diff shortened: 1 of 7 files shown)


pazi88:
I would really change the "BANG Off" and "BANG On" names to more sensible ones like "ALS on/off". Also I don't see anywhere the ignition cut part or idle valve part. You really need to have maps for ignition retard, ignition cut amount and fuel adder amounts and also possibility to adjust how much idle valve is opened when ALS is activated. Otherwise there is no way it will work or be usefull. Now this seems like gimmicky exhaust burble feature and not anti-lag.

PR #380, opened 2020-05-17
Configurable fallback ethanol pct if flex sensor reads outside range.

bera-n:
A feature to set a configurable default ethanol content for the ecu to fall back to in case the flex sensor readings fall outside range. This value is used for fueling calculations while advance and boost corrections use 0% as fallback for safety. Also the default value is used at init so that you don't need to wait up to 1,5s to get a sensor reading before start.


bera-n:
Also this includes 'live view' of the value in the flex-fuel table and logging of the value used for fueling

bera-n:
Made a som rebasing errfrs. Will open new pr with amended commit

PR #381, opened 2020-05-17
Configurable fallback ethanol pct if flex sensor reads outside range.

bera-n:
A feature to set a configurable default ethanol content for the ecu to fall back to in case the flex sensor readings fall outside range. This value is used for fueling calculations while advance and boost corrections use 0% as fallback for safety. Also the default value is used at init so that you don't need to wait up to 1,5s to get a sensor reading before start.

Now fixed some rebasing errors...


muuris:
Wouldn't it be better to just have settable fallback value for the effective eth sensor reading, so fuel/ign/boost all would be corrected in desired way?

Many of us use E85 as primary fuel and gas is used only rarely, when E85 isn't available somewhere, so it might be desired to have all values fall back to those relative to full eth content.

bera-n:
Was thinking about it as a failsafe, being able to start and drive though without risking overadvanced timing and too much boost.

muuris:
Nevermind the previous comment as that can be achieved with current tables. This is welcome addition!

noisymime:
I'm all for a fallback value being available, but I'm unsure on why we then need `ethanolPctFuel` to track it separately. My suggestion would be to simply keep using `ethanolPct` but set an error value to indicate that it's using the fallback value. See: https://wiki.speeduino.com/en/configuration/Errors

This is how the error system was intended to be used (Though admittedly hasn't been utilised particularly much)

noisymime:
Closing as stale

PR #388, opened 2020-05-22
Fix typo in condition

audaciousengineer:
CRANK_ANGLE_MAX_IGN = 90 should only be applied if there is only 1 IGN CHANNEL in 8 cylinder setup, not in any other case. This is the correct condition to check.

### DIFF
--- a/speeduino/init.ino
+++ b/speeduino/init.ino
@@ -705,7 +705,7 @@ void initialiseAll()
         channel4IgnDegrees = 270;
         maxIgnOutputs = 4;
 
-    #if IGN_CHANNELS >= 1
+    #if IGN_CHANNELS == 1
         if( (configPage4.sparkMode == IGN_MODE_SINGLE))
         {
         maxIgnOutputs = 1;


noisymime:
The existing code looks right. The `#ifdef` check is there just to ensure that there is at least 1 ignition channel free. The `if` statement inside the `#ifdef` is then what dictates whether the single channel mode is used or not.

audaciousengineer:
I could be wrong, but let me explain with an example:
* Standard 4 INJ CHANNELS + 5 IGN CHANNELS are configured and we want to run 8 cylinders.
--> #if IGN_CHANNELS >= 1 is true for 5 channels.

* IgnMode is set to Single
--> CRANK_ANGLE_MAX_IGN = 90
----> Now we have 4 timer channels firing every 90 degrees (kind of simultaneous).
------> The code that is there works, but I was under the impression that it could mess things up if 4 interrupts would fire one output in parallel.

noisymime:
The interrupts will queue (up to the queue limit), that could cause some weird affects, but unlikely. 
I might look at implementing a mac ign channels check on the schedule setting, but that will need to be tested very carefully to cover all combinations of cylinders/board/ignition modes. 

However, the code as presented in this PR will cause problems. Think about a standard setup with 4 ignition channels, but where the user has selected single channel mode. You'll get to `#if IGN_CHANNELS == 1` which will be false and the single channel stuff will never be setup for 8 cylinders.

audaciousengineer:
> However, the code as presented in this PR will cause problems. Think about a standard setup with 4 ignition channels, but where the user has selected single channel mode. You'll get to #if IGN_CHANNELS == 1 which will be false and the single channel stuff will never be setup for 8 cylinders.

It should not cause any problems to the general single channel setup. I tried and it worked for my tests. The part you refer to was introduced with PR #355 and this new PR is a correction to that. I made a new PR since in #355 there was this undesireable change to the standard channel configuration which I included by mistake (and did not know how to not include it into a PR based on that).

noisymime:
Ohh right, so the V8 single channel will actually operate 2 different ways, depending on how many ignition channels are available. I had assumed that you intended the single channel mode to always operate on a single timer, regardless of how many were actually available, which would (and does) work fine, but it sounds like that wasn't the intent?

noisymime:
Let me propose a scenario though. Imagine that IGN_CHANNELS is set to 2. Setup is V8 with single channel mode selected. 

With the code as it is today, that will work fine. If though you change it to `#if IGN_CHANNELS == 1` then the single channel mode will NOT work and will only get half the required pulses.

audaciousengineer:
Indeed. In this case, IGN_CHANNELS >= 1 will catch it. You convinced me.
So actually, the original version is fool proof 😆 (at least to some extent).

If the quadruple set or unset of the output does not cause weird things, we should leave it as it is.

PR #402, opened 2020-06-22
Incorporate AFR (stoich/target)

muuris:
This adds the option to incorporate AFR target to pulsewidth calculation. It makes VE table a better representation of actual VE, without AFR targets greatly affecting numbers. After VE table has been tuned, one can adjust an area richer or leaner just from AFR target table -- basically (within limits) without need to touch VE table.

The feature can not be used with "include AFR", hence if one is enabled, the other is disabled. This works in a similar way, but instead of measured AFR it uses stoich AFR, so it is not affected by EGO sensor. It uses the stoich ratio defined in the engine constants tab.

I had this feature in use in my daily driver last year, now made it compatible to current master.

image
image

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -322,7 +322,8 @@ page = 1
 
       fanWhenOff    = bits,   U08,     70, [0:0], "No",        "Yes" 
       fanWhenCranking = bits, U08,     70, [1:1], "No",        "Yes" 
-      unused_fan_bits = bits,  U08,    70,[2:7]
+      unused_fan_bits = bits,  U08,    70,[2:6]
+	  incorporateAFR  = bits,   U08,     70, [7:7], "No",        "Yes"
 
       asePct        = array,  U08,       71, [4],    "%",        1.0,       0.0,     0,    155,       0
       aseCount      = array,  U08,       75, [4],    "s", 		1.0,       0.0,     0.0,     255,    0 ; Values for the afterstart enrichment curve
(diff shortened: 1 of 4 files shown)


burdickjp:
I would like to use this feature. What can I do to help?

muuris:
If you want to try this "on your own" without it being in official fw (as it isn't currently), just merge the code into your build. 

I know there are some users who would benefit from this, but of course the first step is that it needs to be seen useful by larger number of users (or @noisymime at least) and pulled in the master code to become a part of official fw. 

These are for MS, but functionally the same thing:
https://youtu.be/Fwco3zlWF5I?t=1895 (from 31:35 to 35:10)
https://www.youtube.com/watch?v=-OlKdIhIYII

Tjeerdie:
@muuris. Nice work! I made the assumption it was already like your commit until i learned it was not... One suggestion, when keeping the old method, is it not better to place the enable/disable of this old function in EGO closed loop control? This is actually the section where is belongs? In the algorithm selection menu.

muuris:
@Tjeerdie The old function is separate from EGO closed loop control, even though the rough principle is the same. I see no reason for the old method to be there, but I'm quite sure there is a group of users with that enabled and used to it, so I didn't want to break down it's functionality.

Tjeerdie:
@muuris Thanks for the explanation. 

I do agree that the old function itself is useless, but needs to stay because of some users that might use it. (I have no idea how that would work :-D). But it is a closed loop EGO algorithm any way you look at it and keeping this in the VE location WIL keep confusing people setting up a system. A broken O2 sensor will make your system go completely nuts. Disabling all EGO closed loop controls will not help you in this case even if you expect it to. 

So i would still much rather have it removed completely OR put in a location where it actually makes sense. I think it is a mistake to keep it at this location. It was a mistake to put it at this location from the start. But that is just my opinion.

muuris:
I do agree with you on the place it belongs to, but I also think moving it to another dialog may cause similar confusion as removing it. Happy to include such change for this if @noisymime gives it a thumbs up :)

PR #405, opened 2020-07-01
STM32F407 boards battery voltage measurement pin correction

TBAMax:
For experimental black_F407VE boards. Changed battery voltage measurement pin to: "pinBat = PC5" 
(from PC6). PC6 was not analog capable pin so needed to be changed.
PC5 pin tested and working.
Corrects original typo.


noisymime:
No issues with the pin assignment change, just to check though, is there a reason the `-DUSBD_VID=0x0483` was removed from the build_flags?

TBAMax:
Sorry about the -DUSBD_VID=0x048 thing. I just left it in yesterday.  This is my very first pull request ever. I could not figure out how to revert this other change that came with it also.
As I found the Arduino framework currently defines the same USBD_VID=0x0483 by default (for black_F407VE). But the build flag was confusing it and it did not build successfully.

PR #409, opened 2020-07-07
Vbatt reading overflow fixed. sensors.ino

TBAMax:
Prevent vbatt reading form going negative due to the vatVoltCorrect and overflowing the uint variable.
Variable changed to signed int  and added check to correct if gone negative.
Thought given that all filters using the temp variable shoud remain working.

Fixes issues that vbatt reading was erratic when negative vbatt correction was applied and battery not connected ( circuit on usb power).

### DIFF
--- a/speeduino/sensors.ino
+++ b/speeduino/sensors.ino
@@ -498,7 +498,7 @@ void readBat()
 
   //Apply the offset calibration value to the reading
   tempReading += configPage4.batVoltCorrect;
-  if(tempReading > 0){
+  if(tempReading < 0){
     tempReading=0;
   }  //with negative overflow prevention
 
--- a/speeduino/sensors.ino
+++ b/speeduino/sensors.ino
@@ -488,7 +488,7 @@ void readO2_2()
 
 void readBat()
 {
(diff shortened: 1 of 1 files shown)


Autohome2:
It would be better to limit the corrected voltage to a minimum of zero volts rather than use a signed variable.
Battery voltage cannot go negative thus a corrected negative value from the calculation would be untrue.

TBAMax:
> It would be better to explain that the corrected voltage could NOT have been allowed to go negative before checking which would have caused the overflow.And that a limit was introduced to fix the voltage to a minimum of zero volts as true battery voltage cannot really be negative

Yes, thanks. That is better explanation!

VitorBoss:
I think is better to add a fail safe value, the voltage reading should never be < 4V and if this happens set a default 5V on the variable.

noisymime:
I don't think there's a need for a default value here as this is a case that should never occur with an actual running engine. It's only to resolve the strangeness that could appear in TS when testing and don't have a 12v source connected.

I'm comfortable with this solution given the specific problem.

PR #410, opened 2020-07-07
Update platformio.ini

TBAMax:
For black_f407ve board
"-DUSBD_VID=0x0483" build flag not working with the latest current arduino ST STM32 framework. It either wants both the VID and PID, or none. Checked that when this build flag is omitted, framework still assignes the same VID automatically.


TBAMax:
Already done in other commits. Close.

PR #403, opened 2020-07-01
PWM fan control

pazi88:
This PR adds PWM fan control for Teensy and STM32. The Arduino MEGA lacks timers to do this, so if user tries to select PWM fan mode with that MCU, Tuner Studio gives warning that it isn't available and PWM Fan related options are grayed out. Tested to work on STM32, but don't own Teensy, so for that I only tested that this compiles.

PR #411, opened 2020-07-08
07072020 obdp3

Autohome2:
This pr adds obd2 support to Teensy based speeduino @500k 11bit.
you can read the port using any std code reader or custom software.
It has some additional data values  ready for when speeduino supports those
it is also possible to read the auxin data using a pid22 request with data at addresses 0x7701 to 0x7710 (ch1 to 8)


VitorBoss:
Nice work, I think the commits number need to be smaller, I've wrote a step-by-step of how I do on #403

PR #412, opened 2020-07-09
[VVT] PID work

VitorBoss:
Tested on a half  moon CAM pattern as working, it should work on others too.

Added option to reverse the PWM output

Co-Authored-By: pazi88 <48950874+pazi88@users.noreply.github.com>


pazi88:
This is something that many have been waiting for.

PR #413, opened 2020-07-09
Flex fuel sensor temperature and temperature correction

iLeeeZi:
This is to read continental flex fuel sensor temperature and use it to compensate for fuel temperature. This is partially duplicate with #252 but more up to date


noisymime:
A nice high quality PR! Well thought out and everything has been done with consideration to performance etc. 

Thanks also for keeping this up to date.

PR #416, opened 2020-07-15
PWM Fan functionality

pazi88:
This PR adds PWM fan control for Speeduino. Currently fully tested to work on STM32, but don't own Teensy, so for that I only tested that this compiles. Mega implementation is un-tested at this phase.

This is second try of doing this PR. First one had lot of commits, and I couldn't get rid of those, so this has only one including all changes.


pazi88:
Adding the PWM fan functionality to mega too work in progress. So not ready to be merged yet.

pazi88:
This is starting to have again too many commits, so I open new PR, once everything is verified to work properly.

Issue #154, opened 2018-08-11
Water / Meth injection [$200]


DeeEmm:
This is a feature request / firmware enhancement.

I would like to see the integration of water / meth injection control, both a basic on/off version triggered at a pre-determined boost level and also a ratio-metric / dynamic system using MAP or TPS and RPM to enable, a simple offset table with +/-100% setting based on the current injector pulse width along with a 2D ignition correction table.

Some basic discussions relating this this can be found here - [
https://speeduino.com/forum/viewtopic.php?p=27696&sid=b8bb3b67641433a405b017719837a036#p27696](url)

Safeties / failsafes to be considered and at a minimum to include water level sensor to cancel ignition table correction (user enabled).

AU$200 bounty offered

DeeEmm:
Oops, think I raised it in US$ instead of AU$. 

Maybe we could include a shift light along with it too ;)

ByteVenom:
Are you looking for this to run off the speeduino itself? I think this might be more of a GPIO solution.

DeeEmm:
Bumping this as the bounty is about to expire and will be lost :(

DeeEmm:
Deadline is 11 August

DeeEmm:
Link to discussion on Speeduino site above seems to be dead. You can access it at the link below:

[https://speeduino.com/forum/viewtopic.php?f=13&t=2358&p=43710#p43710]

DeeEmm:
Bounty solved by Timu5 PR#417

Great work. Thank you !!!

PR #417, opened 2020-07-15
Water methanol injection

Timu5:
This pr fulfill feature request #154.
Implementation is based on description from forum post [Water / Meth injection - $200 bounty](https://speeduino.com/forum/viewtopic.php?p=43797#p43797).

WMI utilize VVT timer, therefor only one can work at any given time.  When WMI is enabled and WMI alternative boost map is on, the VVT table is being used as current boost map.

### DIFF
--- a/reference/Base Tunes/Speeduino base tune with wmi.msq
+++ b/reference/Base Tunes/Speeduino base tune with wmi.msq
@@ -1,0 +1,1889 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+<msq xmlns="http://www.msefi.com/:msq">
+<bibliography author="TunerStudio MS Lite! 3.1.03 - EFI Analytics, Inc." tuneComment="" writeDate="Tue Jul 21 17:57:23 CEST 2020"/>
+<versionInfo fileFormat="5.0" firmwareInfo="Speeduino+2020.06-dev" nPages="12" signature="speeduino 202306-dev-wmi"/>
+<page>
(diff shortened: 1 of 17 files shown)


noisymime:
Terrific PR! 

Most of this looks fine and I'm more than happy to pull in, I do have  few questions though:

1. What is happening with the vvt as boost stuff that's in there? Is this critical to the WMI work or can it be split out?
2. The current dev data version (In `updates.ino`) is 14, so no need to add a version 15
3. There's a few bugs in some of the boost code that's been changed

I'll go through and add comments and changes directly on the PR if you're OK with it.

Timu5:
1. Nope, it not critical, is just idea to reuse vvt table as alternative boost table to save some memory, it can be safely moved to its own map. Alternative boost map is allowing to use higher boost when water injection is operational(cooling effect of water on intake manifold).
2. Didn't know that. I will change it in free time.
3. Hmm, that's weird because its only selecting which map to use, and was working just fine when tested. Maybe I'm missing something, more details would be useful.

> I'll go through and add comments and changes directly on the PR if you're OK with it.

That's fine.

noisymime:
Can I ask that the vvt as boost table stuff be pulled out for this PR? Once the work is done to complete the memory consolidation, I had planned on adding a 2nd boost table anyway, so that feature will still be there. I just feel it's going to create too much confusion having a joint table like that is all.

Timu5:
I removed vvt as boost and change data version back to 14.

noisymime:
Terrific work! Thanks for the PR

PR #419, opened 2020-07-16
Update black STM32F407VET6 board pinout to SPECTRE V0.5

Tjeerdie:
Updated the pinout for STM32F4 experimental board to V0.5 of the board.

### DIFF
--- a/speeduino/init.ino
+++ b/speeduino/init.ino
@@ -1336,89 +1336,88 @@ void setPinMapping(byte boardID)
         pinCoil3 = 30;
         pinO2 = A22;
       #elif defined(ARDUINO_BLACK_F407VE)
-        //Pin definitions for experimental board Tjeerd 
+       //Pin definitions for experimental board Tjeerd 
         //Black F407VE wiki.stm32duino.com/index.php?title=STM32F407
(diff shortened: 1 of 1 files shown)


TBAMax:
Hi.
Is there some important changes with pins? For example some hardware limitations why some pin must be changed from the previous arrangement. Can you comment on those if there are.

Tjeerdie:
Hi

Well just for routing of the board, mainly and adding injectors 5 to 8. I
left the standard board the same. I could add another board to the list but
to my knowledge there is no other open source board using this pinout.

Tjeerdie:
Full story:

I added this board to speeduino.ini and code base 10 months ago for my then prototype STM32 based board. Also added the same pin out as (default) case, that is the speeduino 0.4 board. This design is now at version 0.5.1 and all looks to be working. Engine is running on this. See: https://github.com/Tjeerdie/SPECTRE 

Now I changed the board pin out to be correct with version 0.5 and 0.5.1 . And I left the  speeduino 0.4 board as before so that anyone using this pin out can select speeduino 0.4  board for testing/design. But I do not know of anyone using this specific pin out for a design yet.

PR #422, opened 2020-07-20
Improvements to new IAT and CLT calibration tables.

TBAMax:
IAT and CLT calibration is stored in 32 bin tables.
Values stores as byte with offset applied.
Axis not stored in EEPROM. But generated fresh at startup.
Axis (ADC) values are 9 bit(stored as 16bit) for more accurate reading.


TBAMax:
Just to comment. I tested this with the simulator and debugger on stm32 verified that memory arrays correspond to the values received. And values that are written to EEPROM and read back from EEPROM are also the same. But there is no update from previous table versions, I just commented this out. So when taking this version: IAT and CLT must be recalibrated from TunerStudio.
And big respect for all the previous work that has been done! It sets good tracks to go forward.

TBAMax:
Now made tests on Arduino Mega. IAT and CLT worked as expected. 
However O2 calibration with its still big 512B table needs several tries to get all of the table written. Suspect EEPROM write is not able to keep up with the data and serial buffer gets overfilled in the Mega. So it works when table write is done 4-5 times, as already written part is not rewritten, and then it adds something with each go. In my opinion this issue should not prevent the current PR to be added. Then O2 calibration can be transferred to 32 bins table later, this should take care of the issue.
BR
Teemo

noisymime:
A couple of questions here as there seem to be some missing pieces:

* You've converted the temperature table values to a `uint8_t` however values above 255 (2550) are valid and are sent by TS
* You've removed the conversion from F to C? Speeduino uses C everywhere internally. 
* Where are you actually doing the calibration burns to EEPROM? This has to happen at the latest by the end of the calibration receive routine otherwise you will get loss here on a restart (And it's loss that isn't picked up when TS connects)
* Pulling this is with a known broken O2 calibration makes me slightly nervous. I need to run the numbers on accuracy for converting that table to a 32 bin 2D table and see how quickly that can go in

VitorBoss:
image

This was generate using those indexes:
`2
6
22
38
54
70
86
102
118
134
150
166
182
198
214
230
246
262
278
294
310
326
355
387
403
419
435
451
467
483
499
508`

TBAMax:
Ok, need for some discussion here:
- @noisymime comment about temperature values not being capped to 255 degrees C is correct, so it is in the to do list.
- About the removed conversion from F to C. My research showed that TunerStudio send values already in C so there is not need for conversion. May be that is dependent on TunerStudio settings, but then one should use C also in TunerStudio when doing calibration. Nothing else to do about that.
-O2 calibration writes are still broken for slower EEPROMS. Calibration writes are done after all data is read, and apparently Tunerstudio do not wait enough time for writes to complete on slower EEPROMs, and resets the board during write. IAT and CLT are fine at the moment but O2 with bigger table is still broken if using slower storage. Did not seem elegant to put byte by byte write back in, because currently unsure if this is currently good and working for all different EEPROM emulations.

noisymime:
> About the removed conversion from F to C. My research showed that TunerStudio send values already in C so there is not need for conversion. May be that is dependent on TunerStudio settings, but then one should use C also in TunerStudio when doing calibration. Nothing else to do about that.

How are you determining this as it doesn't seem to match what I'm seeing. Changing the F or C setting in the calibration dialog does not change the values that are being sent at all. There's nothing in the data that is sent that would indicate whether the values are in C or F, so it's not something that should be user changeable (As the firmware has no way of knowing which is selected).

VitorBoss:
IAT is giving me resets here, only able to use CLT right now

noisymime:
Just as an FYI, I'll have some code I'll be comitting later today that takes a slightly different approach here and actually gets TS to do some of the heavy lifting (Eg it only sends 32 values etc).

noisymime:
Closing this one as it has been completed in other commits.

PR #424, opened 2020-07-22
Update on eeprom emulation

Tjeerdie:
Add EEPROM emulation on internal flash for STM32F7 with help of @960
By request of @VitorBoss a complete re-shuffle of all STM32 EEPROM emulation stuff. Now the selection of emulation type can be done in the globals.h. The selection of CS pin for SPI flash is now done in globals.h too. Test compilation of all other boards in PIO and that still worked. Tested compilation of STM32F407 and Mega2560 in arduino IDE is also working.  

Available selections in globals.h:
- EEPROM emulation on internal flash for STM32F4 and STM32F7.
- SRAM_AS_EEPROM  Use 4K battery backed SRAM, requires a 3V continuous source (like battery) connected to Vbat pin
- USE_SPI_EEPROM PB0 Use M25Qxx SPI flash
- FRAM_AS_EEPROM Use FRAM like FM25xxx, MB85RSxxx or any SPI compatible

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -79,7 +79,6 @@ board = black_f407ve
 ;lib_deps = EEPROM
 board_build.core = stm32
 build_flags = -fpermissive -std=gnu++11 -UBOARD_NR_GPIO_PINS -DARDUINO_BLACK_F407VE -DCORE_STM32_OFFICIAL -DENABLE_HWSERIAL2 -DENABLE_HWSERIAL3 -DUSBCON -DHAL_PCD_MODULE_ENABLED -DUSBD_USE_CDC -DHAL_UART_MODULE_ENABLED
-;-DUSE_SPI_EEPROM=56
 upload_protocol = dfu
 debug_tool = stlink
 monitor_speed = 115200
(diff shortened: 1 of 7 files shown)


TBAMax:
@Tjeerdie just a reminder that you need to add double including guard to the "backupSramAsEEPROM.h" as we discussed in the slack. I did that and then tested, it was all fine.
I would vote for this then for to be pulled in as soon as possible.

Tjeerdie:
@TBAMax I will add this to this PR. Its all to Noisy to add it or not, unfortunately i have no influence on adding this. Please be aware that all the EEPROM emulation implementations from STM32 are broken again because the i did not implement the EEPROM.put() function (it was not used before). And that function is now used by the new temperature calibration routine i guess. @noisymime I can add the fix to this PR or make a separate one after this one is pulled.

Tjeerdie:
Add the fix for the put() and Get() functions in this PR because i did not want to wait for this PR to be pulled in to make a new one (its already open for more than 2 weeks its going to take some more time i suspect). I did not want to make a separate one for the put() and get() functions because that will cause conflicts with this PR.

PR #425, opened 2020-07-31
Pin definition for user board

joesbox:
Pin definition for Teensy 3.5 based ignition-only board - "Juice Box"

Schematic & PCB layout - https://easyeda.com/joesbox/Speeduino-Clone


noisymime:
Looks good! Just to confirm, this board is released under an open license?

joesbox:
Correct, yes.

Joe


-------- Original message --------

joesbox:
Correct, yep

PR #428, opened 2020-08-06
Coolant overpressure engine protection

32PSI:
Pretty much a clone of the Oil pressure sensor code, but for Coolant pressure sensor. Uses 1 extra engineProtectStatus bit and 11 extra bits on config10. Could be useful in boosted engines to detect headgasket sealing problems, when running too much boost.


honsch:
Will this actually work?
As I understand it, the coolant (after warmup) runs a a fixed PSI from the rad cap.
Any overpressure is vented to the overflow.

pazi88:
> Will this actually work?
> As I understand it, the coolant (after warmup) runs a a fixed PSI from the rad cap.
> Any overpressure is vented to the overflow.

The coolant pressure isn't normally as high as is the pressure that the cap will let some steam out. So there is some room to see if the coolant pressure gets higher than normal, before it being vented out.

DeeEmm:
Generally...

- Coolant pressure will change with engine rpms
- System should not vent as it relies on the system being closed (pressurised) to be able to prevent the water from boiling (temps are > 100 deg) 
- The vent is a safety device not a pressure regulating device
- Standard radiator cap vents at something like 16psi, therefore you will never be able to read more than this at it dictates the max pressure.

I don’t know what pressure the cooling system will normally run at, no doubt it’s system specific, but you can assume that at 16psi there must be a safety margin, so theoretically you can set a limit somewhere near to this (15.5psi?) and then be able detect ‘overpressure’ conditions. However you may get false positives as in extreme conditions the system may vent slightly. Some cars are more prone to this than others.

If you are trying to detect a blown head gasket you also have oil passageways and external breaches to consider. 

/DM

32PSI:
My idea was not so much to detect a blown head gasket, but to check if a head gasket begins to not seal at high boost pressures because of the head lifting, you should be able to detect overpressure spikes if the coolant pressure sensor is placed properly. I have not tested this on an actual boosted engine yet, but my idea is based on HPA videos about engine protection and of course there could be more protections in case of low coolant pressure/high oil pressure, but I did not want to add too much to one pull request. Sources:
https://www.hpacademy.com/forum/off-topic-discussion/show/coolant-pressure-sensors
https://www.youtube.com/watch?v=PWz5schtFM4

PR #431, opened 2020-08-07
Code reliability improvements.

TBAMax:
Started work on determining one by one what variables are not really suppose to be global. I determine variable to be local variable when it is used only in the scope of one function (no cross usage).


muuris:
But doesn't this just end up with a lot more unnecessarily duplicated code without any real benefits?

TBAMax:
@muuris, it end up with slightly more lines yes. But at the same time makes code much more understandable. Also makes for less  prone to coding errors because in case of variable is accidentally attempted to be used out of the scope it throws error.

TBAMax:
I have been gradually adding some commits to this PR. But in this PR will only be work on decoders.ino and decoders.h

TBAMax:
Seems this is inessential, so close this now instead of maintaining.

PR #432, opened 2020-08-11
Updates based on testing CLT,IATand O2

TBAMax:
Made common 10bit axis for CLT,IATand O2
Removed axis storing on EEPROM because those are just generated values. And made them to be generated at calibration loading.
Renewed EEPROM addresses to mach the smaller storage needed now.
Made calibration receive routine easier, because there was some issue with probably missing type casts when using pointer to pointer.
Reviewed and renewed the axis arrays declarations in the globals.

### DIFF
--- a/speeduino/comms.ino
+++ b/speeduino/comms.ino
@@ -416,21 +416,21 @@ void command()
       Serial.println(F("Coolant"));
       for (int x = 0; x < 32; x++)
       {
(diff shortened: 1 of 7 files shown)


TBAMax:
This branch is stale now. Not to be used!

TBAMax:
No point keeping it here any more. Closed

PR #433, opened 2020-08-11
fix for 32bit timers on STM32

Tjeerdie:
Another fix when writing to 32bit timer compare registers and not explicitly casting to uint16_t first. All calculations for timing expect roll over at 16 bit. This is not automatically the case for 32 bit timers. It is the case for 16bit timers because these have 16bit registers. This causes weird behavior on STM32 systems with 32 bit timers. The fix is to cast to a 16 bit variable before writing to the register making it roll over at the expected time. This is a similar fix as already been done in scheduler.ino a few months back.


noisymime:
Closing as fixed in 5336e18e876b5a559da52ec2c323ee463451f80a

PR #437, opened 2020-08-17
Added option to use fuel/oil pressure in Bar

pazi88:
This PR modifies the speeduino.ini so that the fuel/oil pressure readings can be in Bar unit also in addition to PSI. No speeduino code changes. The pressure readings are still internally PSI and PSI format is default selection in the settings menu when doing new project.


noisymime:
This was added in 15d43db61a9c2849624d8150d47b9138aa49eee2

PR #438, opened 2020-08-17
Fix for negative pressure sensor readings

pazi88:
If the pressure sensor readings are negative, this fix will set the value to zero in that case. Previously the negative values caused the readings to roll over to highest possible value. Like what happens here:
image


noisymime:
Great pickup!

PR #449, opened 2020-08-24
Completely separated second ignition table from fuel table in firmware

DylanRooijakkers:
- The second ignition table should now work indepently of the second fuel table.
- I haven't tested this yet on my car nor on an ardustim board, but the current advance gauge in tunerstudio seems to be doing what it is supposed to do when changing the table mode from for example "added" to "multiply".
- My only concern is when you try to switch on both tables with the same pin, because you can configure the same pin aswell as pullup as pulldown.


DylanRooijakkers:
*As pull-up or no pull-up

noisymime:
Closing as this was resolved in 61cef9451c03701d8a24724ea5e95ad39664bbcd

PR #450, opened 2020-08-25
Added poll level cam trigger type for missing tooth decoder.

pazi88:
This poll level cam decoder is mainly meant to be used with the "half moon" cam trigger wheels have one big tooth that is half cam rotation wide. This is used with hall sensor in many BMW engines at least (m50tu, m52, m52tu, m54 etc.) plus in engines from other manufacturers too and effectively the cam input is high for one engine rotation and low for second engine rotation. So ecu can just poll the cam input at tooth#1 and know the engine phase immediately from the fact that is the cam input high or low. No need to wait for edge trigger and this allows very fast sequential sync.
Example cam trigger wheel:
image

As for example in this the yellow is the injector pulses and blue 60-2 cam signal. This is with regular one tooth at cam decoder:
image

This is how much faster sync the poll level decoder gets:
image

The decoder can be used with other type of cam trigger wheels if those use hall sensor and the section for low and high are wide enough around #1tooth.  Adding adjustable tooth for checking the phase would make this more versatile, but I couldn't get that working without too much of a performance hit. But #1tooth to check should work for majority of the cases.

There is pattern for ardustim to test this here: https://github.com/speeduino/Ardu-Stim/pull/3
This also works with closed loop VVT.


VitorBoss:
If instead of comparing it in every tooth you compare it in `(toothCurrentCount > (configPage4.triggerTeeth/4))` should improve the code a lot

pazi88:
It should compare only at toot#1. Not every tooth.

PR #451, opened 2020-08-26
Fix logger sequence.

celturbo:
Suitable according to the sequence of variables in comms.ino


noisymime:
This appears to have been resolved elsewhere at some point

PR #454, opened 2020-09-09
Condition vss idle adv

celturbo:
Condition of the VSS value, for activation of Idle Advance.


pazi88:
This would be quite useful small addition.

PR #458, opened 2020-09-22
Add option to run PWM idle valve before engine is cranked over

pazi88:
This PR adds option to run PWM idle valve before engine is cranked over. So the PWM idle valve will start running at cranking duty, soon as Speeduino has booted up, if this is enabled. Works in both closed and open loop PWM mode. Previously the PWM valve starts working only after RPM sync has been achieved.

This is super handy feature to have if you are debugging fresh installation or just checking if the PWM idle valve works as indented. Previously this has required to crank the engine while disabling injectors and try to see if the PWM valve reacts as it should while cranking the engine. Really annoying.

Also this can benefit the engine starting by allowing more air into engine before the RPM sync has been achieved. Thus making cold starts easier if the amount of air to the engine is problem. Many OEM ECUs run the PWM idle valve at cranking duty soon as ECU has been powered, for this reason.

Precautions have been made to prevent this from messing up other idle control algorithms.


muuris:
Nice!

PR #461, opened 2020-10-05
VVT2 addition

pazi88:
This PR adds second VVT output functionality to Speeduino. VVT2 is mostly duplication of VVT1 and it also uses mostly same settings as VVT1. But there is few VVT2 related settings added to make that usable. All three modes (on/off, open loop and closed loop) work for VVT2 output too.

VVT2 only works for missing tooth decoder with single tooth on the second cam. This is mostly due to the fact that I'm not aware of other trigger patters that could use dual VVT. In open loop or on/off mode the second VVT output and map can of course be used for other than VVT use cases. For example as universal PWM output.

Special thanks to @VitorBoss for making this possible in first place in 202008 VVT changes.


pazi88:
Closing this because this has gotten really messy with the several commits. I will do new fresh PR.

Issue #464, opened 2020-10-20
Programmable outputs with 2 conditions


sandmanWAjeep:
**What I wanted to happen:**

- have speeduino control the condenser fan and A/C compressor relay 

- When the A/C button in the car ( AUX Input 0) is pressed, the compressor relay ( programmable output 1, pin 6) will turn on followed by the condenser fan (programmable output 2, pin 43)

- TPS must be < 90% for output 1 to be on

- If CLT is >94*c, output 2 will be on regardless of Input 0

**I setup the programmable output rules as follows**

Select Rule Number = 1
Output Pin Num = 6
Activation Delay = 1
2nd Condition = AND
Condition 1 = Aux in 1 == 0
Condition 2 = TPS < 90

Select Rule Number = 2
Output Pin Num = 43
Activation Delay = 2
2nd Condition = OR
Condition 1 = Aux in 1 == 0
Condition 2 = coolant > 94

This did not achieve the desired outcome. I put some "serial prints" in the utilities.ino file to try and track down what was happening and it appeared as if the second condition for Rule Number 1 is being used when evaluating Rule Number 2.

I did achieve the desired outcome by setting up Rule Number 1 as a "dummy rule" and using rules 2 and 3 to control the outputs

Select Rule Number = 1
Output Pin Num = 44   (unused pin)
Activation Delay = 1
2nd Condition = AND
Condition 1 = Aux in 1 == 0
Condition 2 = TPS < 90

Select Rule Number = 2
Output Pin Num = 6
Activation Delay = 1
2nd Condition =  AND
Condition 1 = Aux in 1 == 0
Condition 2 = coolant < 90

Select Rule Number = 3
Output Pin Num = 43
Activation Delay = 2
2nd Condition =  OR
Condition 1 = Aux in 1 == 0
Condition 2 = VE1 > 94                                         (VE1 was chosen randomly. output 3 turns on when coolant is >94)

Side Note: It would be nice if the Aux input labels were the some on both the "Local Auxillary Input Channel Configuration" page and the "Programmable outputs" page. They are labeled 0 - 15 on the former and 1 - 16 on the latter.

elyrizal:
also it will not respond if selecting output polarity to active low. if select active low, output will always at high

bap168:
Editing the speeduino.ini fixed my issue too.

PR #465, opened 2020-10-21
Sandman w ajeep patch 1

sandmanWAjeep:
I found an issue with the speeduino.ini file with the programmable outputs. The second condition of Rule number 1 would be applied to the second condition of rule number 2. I opened issue #464 about this. The fix is shown in the atached jpeg. This is the first time i've used the "Submit Issue" or "Pull Request" so hopefully i'm going about it the right way.

bugFix


muuris:
It seems you found an issue and on a quick glance a fix for it too, but the pull request is a bit of a mess (see "files changed" and I believe you'll find there a bit more than intended to be included ;) ).

Your fork master should be equal with the master branch to start with (no personal mods or older version etc), then make a branch with a title and short description of the thing it changes (not "patch 1", but more like "fix programmable output 1 & 2 conditions") and only make the modifications inteded, then make a PR. When done like this, it's easy for everyone to see what were the changes -- not with a jpeg image but from the files changed. You can also compare your branch to master before making a PR.

VitorBoss:
What you are proposing is move the TS addresses for the conditions, that won't work, I had wrote it and tested it for 2 months before submitting the pull request, on my tests everything was working as intended, I didn't test the Aux input thing, maybe is something related to that.

sandmanWAjeep:
@muuris. 
Thank you for the information on correct pull request workflow. I made another pull request with only change being the bugfix for the programmable outputs. 

@VitorBoss
It is very possible that I'm wrong on this. But comparing the TS addresses associated with firstDataIn to those of secondData in, the addresses of secondDataIn are one digit off. This is exactly what i am seeing in tuner studio. If I had rule one set to "anyValue"  AND TPS ==0            and set rule two to            "anyTrueValue  AND  anyValue > 1,
rule two would go true any time the TPS was > 1

VitorBoss:
Just had a second look on that and you are absolutely right, somewhere during the changes it got messed up, the condition0 must have the same address of the contrition array

PR #472, opened 2020-11-02
Move 0.x boards forward

VitorBoss:
Empty FRAM reads all data as 0 instead of 255 of flash, this increase the first boards to properly detect an empty FRAM.


noisymime:
As a simpler option, why don't we just get rid of the v0.1 entry? Nobody will ever use it (or at least, they shouldn't be) and it means there won't be any issues with doing the offset move etc.

VitorBoss:
That will be perfect

VitorBoss:
@noisymime Just change the ini don't solve the issue, it need the line `if((configPage2.pinMapping == 255) || (configPage2.pinMapping == 0))` at `init.ino` and remove the `case 0:` from `switch `code

noisymime:
Should be good to go now.

PR #473, opened 2020-11-07
Fix 8x6 for stm32 f407

Tjeerdie:
Added correct number of outputs for the STM32F407 boards to reflect the actual hardware.


noisymime:
Silly question, but should this not be 8x8?

Tjeerdie:
Well the board i designed has only 6 ignition channels. And 8 injection.
But maybe others use the same board definition for an 8x8. So for me it is
fine to make it 8x8.

PR #475, opened 2020-11-12
Rover MEMS flywheel support

james-portman:
The code picks up all of the Rover MEMS flywheel variations

Tested with crank, with and without a cam sensor in tunerstudio - gets full sync and RPM displays correctly.
Tested with a logic analyser running as different combinations correctly:
- crank only - wasted spark, single throttle body injector (sets 2 outputs minimum) - screenshot attached
- crank and cam - wasted spark, 2 sequential port injectors - screenshot attached

Issues to be addressed:
- change the code where it syncs to set the correct tooth count (not just 1), then add code for rolling over from 37 teeth back to 1
- Unknown exact crank sensor location, plus it differs between cars even with the same flywheel - the flywheels and stock ECUs are mix-and-matched it seems, so no way to get this code correct for every car (unless the cam vs flywheel position is unique per setup?). Offset setting should allow this to work anyway

Screenshots:
Screenshot from 2020-11-13 18-37-32
Screenshot from 2020-11-13 18-34-16

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -129,6 +129,7 @@
     #define trigger_ThirtySixMinus222   = 16
     #define trigger_ThirtySixMinus21    = 17
     #define trigger_420a                = 18
+    #define trigger_RoverMEMS           = 20
 
 [Constants]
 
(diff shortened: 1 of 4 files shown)


adbancroft:

PR #477, opened 2020-11-17
Alternative Rover MEMS decoder

mike501:
Alternative Rover MEMS decoder. 

Inspired primary pattern approach from James Portman by building pattern into a variable which is then compared against a binary representation of the trigger wheel. Tested against ardu-stim which will be updated shortly with the four MEMS patterns. Code automatically detects which wheel is running and adapts as required (can change the wheel whilst running which proves the resync ability). 

Support secondary cam and in theory VVT & sequential (not tested).

Additional work needed on setting ignition points and removing compiler warnings from using signed / unsigned variables in comparisons.


noisymime:
Can this be closed now that #1032 is merged?

PR #478, opened 2020-11-19
Enable semi sequential and wasted COP on half sync

VitorBoss:
Enable semi sequential and wasted COP in case of CAM signal missing and 'half sync'.

Only applicable on 4 cylinder at the moment, running on 2 engines for 15 days without detectable bugs, reportedly better cranking.


VitorBoss:
@noisymime I sent a code modification to keep the changes as minimal as possible, could you take a look and share thoughts?

shiznit304:
omg yes! I noticed my car fired up faster when i used to run wasted spark.

shiznit304:
Does this adjust squirts per cycle to allow the car to run in semi sequential fueling? For semi seq I have to run 2 squirts per cycle instead of 1 squirt. 

Could you also update this PR to latest master

VitorBoss:
> Does this adjust squirts per cycle to allow the car to run in semi sequential fueling? For semi seq I have to run 2 squirts per cycle instead of 1 squirt.
> 
> Would this theoretically allow me to start the engine in full sequential with a faulty cam sensor?
> 
> Could you also update this PR to latest master

It automatically scales the req_fuel amount to the change bee seamless. Took me a bit longer to update but just did it

shiznit304:
I tested the PR and it works. Although my car was idling very poorly with 18+ AFR with the cam sensor unplugged. Would that be an issue with my dead times? Plugging in the sensor make it go into full sync and idled normally.

One thing I didn't test to see what would happen if I disconnected the cam sensor while it was full sync. I've done it in the past without this PR and the engine still runs fine in full sync.

VitorBoss:
If you loose CAM signal after sync it won't be a problem unless you got a lost sync.
The poor idle is related to the sequence in semi-sequencial, it pulses 1-4 and 2-3 making it wrong for sequential wiring which is 1-2-3-4, I think there is an issue related to that. The goal is to improve cranking and be able to run the engine without the CAM. It ain't perfect, but could be improved later

shiznit304:
Can it be changed to fire outputs 1,3 then 2,4? I did notice that my car does start up much faster with this PR.

KusE36:
Hello. How to make half sync on my bmw inline 6 ms 41 from pazi ? How to do that while i lost cam signal on full sequential setup ?

VitorBoss:
I don't know how the engine will behave but there is already code for 6 and 8 cylinder. Try it and let me know

digmorepaka:
https://github.com/noisymime/speeduino/compare/560537f1dab1...4d747e1aef7b this commit has allowed  configuration between the two possible pairing options of semi-sequential on a 4cylinder

VitorBoss:
The number of commits on this was out of hand after 18 months in cold bath, just forced update the branch, tested compilation and sent again fresh as new ;-)

Also added a small section to the update procedure to make sure the new flag is set automatically for a 4 cylinder using sequential injection as most use 1-3-4-2 firing sequence, and a valid value if not.

noisymime:
If I'm reading correctly, this is calling `changeHalfToFullSync()` every single loop . This seems fairly excessive as it only ever needs to be called when actually switching from one mode to the other doesn't it?  Is there a way to watch for this change and only call it as required?

VitorBoss:
Done

felipegaspari:
I merged this PR to the current master and tested it on my engine. Sequential injection had big issues, not sure if it happens when in half or full sync. 
Have you checked the new option to set injection pairing?

VitorBoss:
Not sure what issues you are experiencing, I have tested this and some guys are running this in their engines without issues.

Can you share what you are seeing wrong?

felipegaspari:
Well, it feels like the car is running on 2 cyls... but i didnt use your branch, i just merged your commits to current master :/ I'll try it again tomorrow with logging enabled

noisymime:
Was there ever any conclusions reached around the issue mentioned by @felipegaspari ?

I'd like to get this included for the next release, but need to understand if there are setups where this causes problems.

VitorBoss:
> Well, it feels like the car is running on 2 cyls... but i didnt use your branch, i just merged your commits to current master :/ I'll try it again tomorrow with logging enabled

Just added a sync status to log fields, can you do another log please, is really weird it took 6.4 seconds to start the problem. 
image

VitorBoss:
image
A capture with injector pairing 1+3/2+4
And here a 1+4 / 2+3
image

felipegaspari:
I tested it and the issue is fixed. Great job! @noisymime

noisymime:
I am seeing errors when switching from Semi-Seq to Sequential as a result of this pull. To reproduce, set:
* 4 cylinder
* Dual wheel
* Semi-sequential fuel
* Switch to Sequential fuel and Burn

When Burn is pressed an error is received.

I tracked this down to `changeHalfToFullSync()` where currently there is:
```
      case 4:
        inj1StartFunction = openInjector1;
        inj1EndFunction = closeInjector1;
        inj2StartFunction = openInjector2;
        inj2EndFunction = closeInjector2;
        channel3InjEnabled = true;
        channel4InjEnabled = true;
        break;
```

To resolve the issue, the below lines need to be added:
```
inj3StartFunction = openInjector3;
inj3EndFunction = closeInjector3;
inj4StartFunction = openInjector4;
inj4EndFunction = closeInjector4;
```

Before adding this for 4/6/8 cylinders in `changeHalfToFullSync()` I just wanted to check if there was a reason these were omitted initially?

VitorBoss:
It is assumed the injection mode was the same as on boot. To make it able to be changed on a running engine those changes need to be added indeed

PR #481, opened 2020-11-25
Closed loop idle improve

Tjeerdie:
Added improved closed loop idle control strategy "PWM open+closed loop idle control". Its added as a separate control function not to upset existing installs. Two main improvements:

1. Now open loop idle valve value is used as starting point for the closed loop controller. The PID controller adds or subtracts from this point to get to the target RPM. Some reading about the concept of feedforward: https://www.controleng.com/articles/feed-forwards-augment-pid-control/ It is a bit much for the Idle control because it is not so dynamic so the feed forward does not do much. It also does not hurt so i kept it in.

2. The PID controller was active always (even when cruising or accelerating). So the PID control constantly tried to reduce the RPM. closing the idle valve until the lowest limit. This was most noticeable because of the integral (I) term of the PID. When coming back from higher RPM to idle first the RPM dip and than the controller reacts to the error by filling the Integral buffer. This takes a couple of cycles. For an cold engine this can make it stall. Now there are two variables (RPM and TPS) in tuner studio to reset the integral (I) term to zero when either limit is reached. With zero integral term the open loop idle value is the main contributor smoothing idle control.


muuris:
Cool, this is more oe-like solution to closed loop.

noisymime:
Looking forward to testing this. Nice neat implementation with the separate control method too.

Thanks!

PR #482, opened 2020-12-01
STM32 CAN bus library addition

pazi88:
Added CAN bus library for STM32. The library is not yet "fully featured" so it's missing few things like using multiple mailboxes and filter usage. But it has everything that current CAN features made for Teensy require in the Speeduino Firmware. And it's made in a way that it works almost identically to Teensy FlexCan library. So there is no really changes needed in Speeduino code to bring the CAN bus working on STM32. Only few compile flag changes.

### DIFF
--- a/speeduino/board_stm32_official.h
+++ b/speeduino/board_stm32_official.h
@@ -276,6 +276,12 @@ void ignitionSchedule8Interrupt(HardwareTimer*);
 */
 #if defined(ARDUINO_BLACK_F407VE)
 //HardwareSerial CANSerial(PD6, PD5);
+#include <src/STM32_CAN/STM32_CAN.h>
+//This activates CAN1 interface on STM32, but it's named as Can0, because that's how Teensy implementation is done
+STM32_CAN Can0;
+
+static CAN_message_t outMsg;
+static CAN_message_t inMsg;
 #endif
 
 #endif //CORE_STM32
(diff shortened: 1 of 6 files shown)


noisymime:
Looks great, terrific work!

Only comment is that you've changed a few of the `CORE_TEENSY35` checks to `CORE_TEENSY` when you've added the stm32. I'll update those back to Teensy 35 as they will break the `TEENSY41` build (When it comes), and merge.

pazi88:
@noisymime That's more than fine. Also one thing about the compile flags. I have there that `defined(ARDUINO_ARCH_STM32)` but now later when I tried to add CAN support for STM32F401 and F411 that seem to have some kind of support in FW too, I noticed that those don't actually have integrated CAN bus. So that  `defined(ARDUINO_ARCH_STM32)`  will break building for those. Don't know how big problem that is, but I was thinking of maybe have new compile flag like: `defined(INTERNAL_CANBUS_AVAILABLE)` and have those set in the board files for the boards/mcu's that have that. But @Autohome2 said that for some reason that doesn't work for Teensy and there still would be that `defined(CORE_TEENSY35)` in there??

Tjeerdie:
@pazi88 Great work! So some form of CAN bus now works on the STM32? I might use this for a dashboard project when the time comes! 

Setting macros in any project file will not be usable in library source. The problem is that the library and Arduino framework is build before the project files are build. You could explicitly check for the  STM32F401 and F411 boards and skip building the library for this. And than in the speeduino project not initialize anything CAN based for the STM32F401. This is the way i have fixed it now for the internal flash as EEPROM emulation.

PR #484, opened 2020-12-02
Added boards from RM

VitorBoss:
Added boards made by Rafael Matos


noisymime:
Can you please confirm whether or not these boards are released under an open license?

PR #486, opened 2020-12-04
Only do ign math if channel is active

VitorBoss:
Currently all ignition angles are calculated, even the ones not enabled, this proposal is to do like the injection already does, only calculate angles for used channels.

### DIFF
--- a/speeduino/speeduino.ino
+++ b/speeduino/speeduino.ino
@@ -967,12 +967,13 @@ void loop()
   #endif
         
 #if IGN_CHANNELS >= 2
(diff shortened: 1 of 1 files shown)


noisymime:
I can't see any harm in this, but at a quick count the benefit of this is incredibly small. Without checking the ASM, there looks to be about 30 instructions per channel, so even in the case of say 2 ignition channels, at most you're only cutting out around 180 instructions (6x30). 

180 instructions works out at around 11uS, or about 0.01 loops per second :) 
So, no harm in it, but it's not a huge thing either way.

PR #487, opened 2020-12-04
Fix pages 13 and 14 EEPROM allocation

VitorBoss:
Fix as discussed today on development channel on slack


noisymime:
At a glance this looks to allocate roughly another 110 bytes. How much is showing as free with this change (On 2560)?

pazi88:
@noisymime The idea was to make the page 12 actually 192 bytes big as it was defined. And once that is done, maybe use the page12 for vvt2 map and thus not even create the page15 for vvt2, which seems to eat up bit too much RAM when made reasonable big.

VitorBoss:
That is just an EEPROM allocation fix, the page size and RAM usage says the same.

PR #489, opened 2020-12-06
Fix building stm32 f401

Tjeerdie:
Couple of things are now fixed and cleaned up for STM32

1. STM32F401CC builds again after addition of CAN bus support (STM32F401CC has no can bus)
2. BLACK_STM32F407VE builds again from Arduino IDE (Was broken by removing macro STM32_OFFICIAL_CORE)
3. General macro cleanup for CAN bus definitions (together with fixing nr1) This needs to be tested because i have no working CAN board yet.I do not expect issues.
4. Removed most BLACK_STM32F407VE macro's for the black eBay boards and replaced it with macro for generic stm32F407xx 
5. Now speeduino at-least builds when selecting one of the STM32F407xx generic boards. possibly extending the number of supported MCU. 
6. Cleaned-up some build flags for the STM32 boards in platformio.ini


VitorBoss:
`STM32_OFFICIAL_CORE` is defined by the library stm32duino, weird it is braking the compilation

Tjeerdie:
@VitorBoss hmmm it broke my arduino IDE build atleast. Not the PIO because it was defined also in the build flags. I will check if something else broke it.

VitorBoss:
My bad, I did check the stm32duino library and it don't have the definition I've mentioned.

Tjeerdie:
I just did the same :-D. check my edited response.

PR #490, opened 2020-12-06
Coolant temperature protection.

opelpanfan:
Coolant temperature protection:

1. Possibility enable coolant temperature protection routine
2. 2D map RPM based on Coolant temperature.

Benefits:
1. Engine overheating protection if the coolant temperature is too high.
2. Limit RPM if the engine is below the ideal working temperature.
3. Good feature to have as many aftermarket ecu's has this one.


dantob:
The x/y axis are flipped compared to oil pressure protection

Otherwise, I did a test merge of this, minor conflicts in speeduino.ini, and the disable on update needs to move to EEPROM_DATA_VERSION == 17.

noisymime:
This was implemented in #785

PR #493, opened 2020-12-16
Universal trigger pattern 202012

james-portman:
Tested on the bench with a Rover 36-1-1 pattern and a strange Renault 44-1-1, both working sensibly.

Configuration done through tunerstudio to select the trigger pattern but then shares the same code in decoders.ino

Ideas to squash the used memory down:
class to wrap saving/loading/shifting bits inside an array of bytes
toothAngles array is 48 bits apparently - could use
bitfields (in structs) - seem messy to access
64 bit types - don't seem to work properly

Hide the whitespace in the diff, there are loads of stray spaces etc that my editor keeps fixing:
https://github.com/noisymime/speeduino/pull/492/files?w=1

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -84,30 +84,30 @@
     fuelLoadMax    = scalar, U08, "", 1, 0, 0, 511, 0
     ignLoadMax     = scalar, U08, "", 1, 0, 0, 511, 0
     AUXin00Alias  = string, ASCII, 20
-    AUXin01Alias  = string, ASCII, 20 
-    AUXin02Alias  = string, ASCII, 20 
-    AUXin03Alias  = string, ASCII, 20 
-    AUXin04Alias  = string, ASCII, 20 
-    AUXin05Alias  = string, ASCII, 20 
-    AUXin06Alias  = string, ASCII, 20 
-    AUXin07Alias  = string, ASCII, 20 
-    AUXin08Alias  = string, ASCII, 20 
-    AUXin09Alias  = string, ASCII, 20 
-    AUXin10Alias  = string, ASCII, 20 
-    AUXin11Alias  = string, ASCII, 20 
-    AUXin12Alias  = string, ASCII, 20 
-    AUXin13Alias  = string, ASCII, 20 
-    AUXin14Alias  = string, ASCII, 20 
+    AUXin01Alias  = string, ASCII, 20
+    AUXin02Alias  = string, ASCII, 20
+    AUXin03Alias  = string, ASCII, 20
+    AUXin04Alias  = string, ASCII, 20
+    AUXin05Alias  = string, ASCII, 20
+    AUXin06Alias  = string, ASCII, 20
+    AUXin07Alias  = string, ASCII, 20
+    AUXin08Alias  = string, ASCII, 20
+    AUXin09Alias  = string, ASCII, 20
+    AUXin10Alias  = string, ASCII, 20
+    AUXin11Alias  = string, ASCII, 20
+    AUXin12Alias  = string, ASCII, 20
+    AUXin13Alias  = string, ASCII, 20
+    AUXin14Alias  = string, ASCII, 20
     AUXin15Alias  = string, ASCII, 20
 
     prgm_out00Alias  = string, ASCII, 20
-    prgm_out01Alias  = string, ASCII, 20 
-    prgm_out02Alias  = string, ASCII, 20 
-    prgm_out03Alias  = string, ASCII, 20 
-    prgm_out04Alias  = string, ASCII, 20 
-    prgm_out05Alias  = string, ASCII, 20 
-    prgm_out06Alias  = string, ASCII, 20 
-    prgm_out07Alias  = string, ASCII, 20 
+    prgm_out01Alias  = string, ASCII, 20
+    prgm_out02Alias  = string, ASCII, 20
+    prgm_out03Alias  = string, ASCII, 20
+    prgm_out04Alias  = string, ASCII, 20
+    prgm_out05Alias  = string, ASCII, 20
+    prgm_out06Alias  = string, ASCII, 20
+    prgm_out07Alias  = string, ASCII, 20
 
     ;Define aliases for all the triggers. Naming pattern matches that used in decoders.ino
     #define trigger_missingTooth        = 0
@@ -129,6 +129,7 @@
     #define trigger_ThirtySixMinus222   = 16
     #define trigger_ThirtySixMinus21    = 17
     #define trigger_420a                = 18
+    #define trigger_universal           = 20
 
 [Constants]
 
@@ -208,7 +209,7 @@
     #define PIN_OUT10inv = "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID"
     #define PIN_OUT16inv = "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID"
     #define PIN_16inv = "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID"
-  
+
     #define ANALOG_PIN = "A0", "A1", "A2", "A3", "A4", "A5", "A6", "A7", "A8", "A9", "A10", "A11", "A12", "A13", "A14", "A15", $PIN_16inv, $PIN_16inv, $PIN_16inv
     #define DIGITAL_PIN = "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "22", "23", "24", "25", "26", "27", "28", "29", "30", "31", "32", "33", "34", "35", "36", "37", "38", "39", "40", "41", "42", "43", "44", "45", "46", "47", "48", "49", "50", "51", "52", "53", "54", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID"
     #define ANALOG_DIG_PIN_LIST = $DIGITAL_PIN, $ANALOG_PIN
@@ -239,7 +240,7 @@ page = 1
 #else
       aeColdTaperMin = scalar, U08,       2,       "F", 1.8,    -22.23,    -40,    215,      0     ;AE cold adjustment, taper start clt (full adjustment)
 #endif
-    
+
       aeMode        = bits,   U08,       3, [0:1],  "TPS", "MAP", "INVALID", "INVALID"
       battVCorMode  = bits,   U08,       3, [2:2],  "Whole PW", "Open Time only"
       SoftLimitMode = bits,   U08,       3, [3:3],  "Fixed", "Relative "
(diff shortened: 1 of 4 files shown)


TBAMax:
Seems like a good idea. I generally like the idea of having basically single decoder and it can do all different wheels and only needs the bit pattern for the wheel.  
  Maybe it would even be possible to have it pick up the pattern automatically... lets say from the engine being turned over by the startermotor. And then it would be very convenient to put a microswitch in the spark plug hole of the cylinder1 and make it get the correct zero angle too.
  Definitely needs the pattern for the cam sensor also.

TBAMax:
Some discussion here would be proper, to discuss how the universal trigger would work.
I would do it the following way:
Initially the trigger searches for biggest difference between two pairs of sequential transitions in the pattern, and then start searching for the same ratio between the sequential incoming pulses. Once the macth is found, it would lock on, and start observing/expecting the following pulses according to the trigger pattern(then those incoming pulses can now be filtered as normal). Do not know if this is the method used in this PR.

james-portman:
> Some discussion here would be proper, to discuss how the universal trigger would work. I would do it the following way: Initially the trigger searches for biggest difference between two pairs of sequential transitions in the pattern, and then start searching for the same ratio between the sequential incoming pulses. Once the macth is found, it would lock on, and start observing/expecting the following pulses according to the trigger pattern(then those incoming pulses can now be filtered as normal). Do not know if this is the method used in this PR.

I don't think most people are interested in this now.
The idea was that you give your trigger pattern in binary so 1 for a tooth, 0 for a gap, and write the whole pattern out up front, and give the total tooth count.
I think trying to auto detect a pattern is not a good idea because of the engine compression and weak starter motor/battery, you don't always get a very clear or consistent output, it is more jumpy.

TBAMax:
> The idea was that you give your trigger pattern in binary so 1 for a tooth, 0 for a gap, and write the whole pattern out up front, and give the total tooth count.

The way I understand it there is a binary pattern and each bit represents one fraction of the engine cycle. I find it quite nice idea.
> I think trying to auto detect a pattern is not a good idea because of the engine compression and weak starter motor/battery, you don't always get a very clear or consistent output, it is more jumpy.

Yes doing the auto detection of the pattern is maybe getting too far ahead for the beginning of the development. But I would nevertheless keep that possibility in mind. Or later possibility of enabling "auto correction" of the pattern based on data collected during running engine (during tuning that is). 
Also there is already a tooth logger. I understand currently the tooth logger uses timestamps for the detected edges...

Interested in general what can be done with the universal bit pattern style. Can it in the end of the day outperform the usual missing tooth decoder? I can see no theoretical problem why it could not replace all of the decoders, all but some special very special ones. The flash memory saved from deleting all decoders and leaving only one seem already very appetising... (not to say here that we delete all the other decoders right away)

mike501:
I've written code based on this idea for the Rover decoders. It seems to
work well on the bench but I'm struggling to get people to test in real
life (got another now hopefully - I wrote it expecting demand as it's in a
lot of UK kit cars).

Off the back of this I tried to write a universal decoder, the problem was
the CPU power when I started to try to cover gaps of 2,3,4 sized teeth for
crank and cam. The principle worked, it just took a lot of CPU power doing
the gap size calculation. Intend revisiting it as someone has asked about a
Subaru decoder which has single and double gaps.

Rover code is here (it auto decodes 4 different Rover patterns using the
principle)
https://github.com/mike501/speeduino/tree/RoverMEMS

mbelt:
It would be very satisfying to eliminate `encoders.ino`. Code quality is all over the place, nearly-identical implementations inflating the bin, and really you shouldn't have to rebuild the firmware to define an encoder. 

I don't think a generic decoder based purely on encoding the angular position of each tooth will work well. It implies timing calculations, which aren't strictly necessary for all decoder types.

## Proposal

__Every decoder implementation is already a kind of state machine.__
Let's come up with a universal FSM, and a serializable encoding scheme of states and transitions to define decoders. 

This is your state.
```
state:
  in_sync: bool
  pri:
    current_tooth: number
    last_seen: time
  sec:
    current_tooth: number
    last_seen: time
```
On interrupt, the input for state transition is
```
input:
  channel: pri | sec
  timeΔ: usec # since last pri or sec, depending on channel.
```

`state` + `input` = `next_state`.

Unless something goes wrong, a default state transition can be used in all cases!
```
default_transition(channel, now_time) {

  switch channel:
  case pri:
    next_state.pri.current_tooth = (curernt_state.pri.current_tooth == num_pri_teeth) ? 1 : current_state.pri.current_tooth + 1;
    next_state.pri.last_seen = now_time;
    break;
  case sec: 
    next_state.sec.current_tooth = (curernt_state.sec.current_tooth == num_sec_teeth) ? 1 : current_state.sec.current_tooth + 1;
    next_state.sec.last_seen = now_time;
    break;
}
```

I think all encoders can be defined something like:
```
decoder:
  tooth_1_offset: degrees
  pri: 
    type: cam | crank
    teeth_count: number
  sec: 
    type: cam | crank
    teeth_count:number
  states:
  - valid_state: {}
    valid_input: {}
  - valid_state: {}
    valid_input: {}
    ...
```

The only time the default transition isn't used is when input isn't valid for a given state. When this happens, `in_sync` is set to `false` and the rest of the current state is set to the first valid state in the decoder definition. This will eventually be correct within one crank revolution. 

For dual-wheel type decoders, evaluating input as valid/invalid given the current state does not requires any time calcs, basically boils down to, "for each possible state (primary/secondary tooth combo), is a primary/secondary tooth expected or not."

For missing-tooth and pattern type decoders, evaluating input as valid/invalid requires time->angle calc.


## Example

```
# Nonsense two-wheel with crank teeth at 0, 120, 240 (360, 480, 500) and cam teeth at 0, 360, 680.
# The falling edges of cam and crank align perfectly at 0, 360, so interrupts could come in either order, the model accommodates this. 
# There are 9 states.
basic_hairbrain:
  pri:
    teeth: [0,120,240,360,480,500] # Only used in getCrankPos();
    type: crank
  sec: 
    teeth_count: 3
    type: cam
  states:
  - state:
    - current:
       pri: 0            # Index into teeth array, not degrees. Note switch to 0-based index compared to A/B.
       sec: 2 
      input:            # The expected input
        - pri: false   # We should see a secondary tooth before primary gets to 1
          sec: true
    - current:
       pri: 0           
       sec: 0 
      input:           
        - pri: true   # We should see a primary tooth next.
          sec: false
    - current:
       pri: 1
       sec: 0
      input:            
        - pri: true   
          sec: false
    - current:
       pri: 2
       sec: 0
      input:            
        - pri: true  
          sec: false
    - current:
       pri: 3          # 360 degrees
       sec: 0         # cam might be at 0 still
      input:            
        - pri: false   
          sec: true # expect secondary
    - current:
       pri: 3          # 360 degrees
       sec: 1         # cam could also be at 1
      input:            
        - pri: true  # expect primary
          sec: false
   - current:
       pri: 4
       sec: 1
      input:            
        - pri: true  
          sec: false
   - current:
       pri: 5
       sec: 1
      input:              # Demonstrate feature of model: On this tooth combo, validate using timing for no good reason.    
        - pri: false
          sec: 320     # Degrees of the crank since last sec tooth.
   - current:
       pri: 5
       sec: 2
      input:            
        - pri: true  
          sec: false
```

### Real Example (Miata - Type C)

```
miata:
  pri:
    teeth: [175, 355, 535, 715]
    type: crank
  sec: 
    teeth_count: 2
    type: cam
  states:
  - state:
    - current:
       pri: 0 
       sec: 1 
      input:            
        - pri: true   
          sec: false
    - current:
       pri: 1 
       sec: 0
      input:            
        - pri: false   
          sec: true
    - current:
       pri: 1
       sec: 0 
      input:            
        - pri: false   
          sec: true
    - current:
       pri: 2
       sec: 0 
      input:            
        - pri: false   
          sec: true       
    - current:
       pri: 2
       sec: 1 
      input:            
        - pri: true   
          sec: false     
    - current:
       pri: 3
       sec: 1 
      input:            
        - pri: true   
          sec: false     
```

james-portman:
> It would be very satisfying to eliminate `decoders.ino`. Code quality is all over the place, nearly-identical implementations inflating the bin, and really you shouldn't have to rebuild the firmware to define a new decoder.
> 
> I don't think a generic decoder based purely on encoding the angular position of each tooth will work well. It implies timing calculations in the ISRs, which aren't strictly necessary for all decoder types.
> 
> ## Proposal
> **Every decoder implementation is already a kind of state machine.** Let's come up with a universal FSM, and a serializable encoding scheme of states and transitions to define decoders.
> 
> This is your state.
> 
> ```
> state:
>   in_sync: bool
>   pri:
>     current_tooth: number
>     last_seen: time
>   sec:
>     current_tooth: number
>     last_seen: time
> ```
> 
> On interrupt, the input for state transition is
> 
> ```
> input:
>   channel: pri | sec
>   timeΔ: usec # since last pri or sec, depending on channel.
> ```
> 
> `state` + `input` = `next_state`.
> 
> Unless something goes wrong, a default state transition can be used in all cases!
> 
> ```
> default_transition(channel, now_time) {
> 
>   switch channel: 
>   case pri:
>     next_state.pri.current_tooth = (curernt_state.pri.current_tooth == num_pri_teeth) ? 1 : current_state.pri.current_tooth + 1;
>     next_state.pri.last_seen = now_time;
>     break;
>   case sec: 
>     next_state.sec.current_tooth = (curernt_state.sec.current_tooth == num_sec_teeth) ? 1 : current_state.sec.current_tooth + 1;
>     next_state.sec.last_seen = now_time;
>     break;
> }
> ```
> 
> I think all encoders can be defined something like:
> 
> ```
> decoder:
>   tooth_1_offset: degrees
>   pri: 
>     type: cam | crank
>     teeth_count: number
>   sec: 
>     type: cam | crank
>     teeth_count:number
>   states:
>   - valid_state: {}
>     valid_input: {}
>   - valid_state: {}
>     valid_input: {}
>     ...
> ```
> 
> The only time the default transition isn't used is when input isn't valid for a given state. When this happens, `in_sync` is set to `false` and the rest of the current state is set to the first valid state in the decoder definition. This will eventually be correct within one crank revolution.
> 
> For dual-wheel type decoders, evaluating input as valid/invalid given the current state does not requires any time calcs, basically boils down to, "for each possible state (primary/secondary tooth combo), is a primary/secondary tooth expected or not."
> 
> For missing-tooth and pattern type decoders, evaluating input as valid/invalid requires time->angle calc.
> 
> ## Example
> ```
> # Nonsense two-wheel with crank teeth at 0, 120, 240 (360, 480, 500) and cam teeth at 0, 360, 680.
> # The falling edges of cam and crank align perfectly at 0, 360, so interrupts could come in either order, the model accommodates this. 
> # There are 9 states.
> basic_hairbrain:
>   pri:
>     teeth: [0,120,240,360,480,500] # Only used in getCrankPos();
>     type: crank
>   sec: 
>     teeth_count: 3
>     type: cam
>   states:
>   - state:
>     - current:
>        pri: 0            # Index into teeth array, not degrees.
>        sec: 2 
>       input:            # The expected input
>         - pri: false   # We should see a secondary tooth before primary gets to 1
>           sec: true
>     - current:
>        pri: 0           
>        sec: 0 
>       input:           
>         - pri: true   # We should see a primary tooth next.
>           sec: false
>     - current:
>        pri: 1
>        sec: 0
>       input:            
>         - pri: true   
>           sec: false
>     - current:
>        pri: 2
>        sec: 0
>       input:            
>         - pri: true  
>           sec: false
>     - current:
>        pri: 3          # 360 degrees
>        sec: 0         # cam might be at 0 still
>       input:            
>         - pri: false   
>           sec: true # expect secondary
>     - current:
>        pri: 3          # 360 degrees
>        sec: 1         # cam could also be at 1
>       input:            
>         - pri: true  # expect primary
>           sec: false
>    - current:
>        pri: 4
>        sec: 1
>       input:            
>         - pri: true  
>           sec: false
>    - current:
>        pri: 5
>        sec: 1
>       input:         # Demonstrate feature of model: On this tooth combo, validate using timing for no good reason.    
>         - pri: false
>           sec: 320  # Degrees of the crank since last sec tooth.
>    - current:
>        pri: 5
>        sec: 2
>       input:            
>         - pri: true  
>           sec: false
> ```
> 
> ### Real Example (Miata)
> ```
> miata:
>   pri:
>     teeth: [175, 355, 535, 715]
>     type: crank
>   sec: 
>     teeth_count: 2
>     type: cam
>   states:
>   - state:
>     - current:
>        pri: 0 
>        sec: 1 
>       input:            
>         - pri: true   
>           sec: false
>     - current:
>        pri: 1 
>        sec: 1
>       input:            
>         - pri: false   
>           sec: true
>     - current:
>        pri: 1
>        sec: 0 
>       input:            
>         - pri: true  
>           sec: false
>     - current:
>        pri: 2
>        sec: 0 
>       input:            
>         - pri: false   
>           sec: true       
>     - current:
>        pri: 2
>        sec: 1 
>       input:            
>         - pri: true   
>           sec: false     
>     - current:
>        pri: 3
>        sec: 1 
>       input:            
>         - pri: true   
>           sec: false     
> ```

This PR is nearly 3 years old without being merged. Make a new PR with your ideas if you like

PR #494, opened 2020-12-27
Hella OPS+T Sensor interface

gdiciocco:
I recently discovered this OPS+T cheap and powerful Oil temperature + pressure sensor from Hella:

https://www.hella.com/municipal/assets/media_global/KI_OPS_T_HELLA_EN.pdf

I'm trying speeduino for the first time for a custom motorcycle project and being able to monitor oil temperature and pressure looks like a good catch. Instead of relying solely on the OEM on/off pressure switch with this 30€ sensor one can evaluate oil pump efficiency, relief valve status, oil cooling circuit behaviour and more.

**Core implementation:**
The sensor outputs a pretty fast (3khz) PPM signal made of three pulses, a diagnostic one and two for the data.

The implementation I made relies on triggering an interrupt by both edges of each pulse, timing them, discovering the first one (the shorter) and using it to sync to the pulse stream. 

The interrupt is activated from the existing 4hz timer and deactivated by the ISR itself so the impact on the MCU should be minimal. I appeal to your own experience if there's a more efficient way of doing this.

**TunerStudio integration:**
I've added interfacing of this sensor to the Fuel/oil pressure menu item.
The user can select it as an alternative to the analog oil pressure sensor. Choosing it will enable the necessary routines. As it will also make oil temperature readings available i've added a note in its section into the panel to inform the user. 
As it will populate the existing oilPressure variable all the existing features regarding oil pressure, logging alarms etc are functional as before.

Oil temperature is available in its own gauge.

**TODOs and known problems:**

It would be nice to implement alarms limits and tables as per the (existing) oil pressure for the oil temperature too.
I had some troubles catching the digital pin 52 in the teensy by using the existing methods (pinTranslate), using it drove to an erratic behaviour. I avoided using pinTranslate as it will be interfaced to digital pins exclusively.
The total number of pins defined for the teensy in its header was 34, the same as the number of the digital pins, I discovered this while trying to understand why my pin assignment from pinOilSensorOPSt in init.ino wasnt working. It's just an error or I'm missing something ?
Theres to rationalize the selectable pins for the sensor into its selector into its configuration panel as I just avoided the crucial ones and made available all the others for full compatibility; it can be useless and eventually dangerous (?).

I started with many clauses which would have made available this addition to the teensy only, however at the end I realized that it should work on other platforms too, I tried to compile it for the mega2560 and it compiled, soon i'll try to connect the sensor to a mega and see if everything is really ok.

Waiting for your feedback.

Kind regards.
Giacomo.

### DIFF
--- a/speeduino/sensors.ino
+++ b/speeduino/sensors.ino
@@ -749,16 +749,16 @@ void readOPSt () {
       {
         oilSensorOPStPulse.index++;
         oilSensorOPStPulse.gotSync = 1; 
-        oilSensorOPStData.status = (1024.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime;
+        oilSensorOPStData.status = ((1024UL*oilSensorOPStPulse.onTime)/oilSensorOPStPulse.totalTime)/4; // 640uS max, making it fit uint8
       } 
       else if (oilSensorOPStPulse.index == 1 && oilSensorOPStPulse.gotSync == 1) 
       {
-        oilSensorOPStData.temperature = ((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime-128)/19.2;
+        oilSensorOPStData.temperature = fastMap((4096UL*oilSensorOPStPulse.onTime)/oilSensorOPStPulse.totalTime, 128, 3968, 0, 200);
         oilSensorOPStPulse.index++;
       } 
       else if (oilSensorOPStPulse.index == 2 && oilSensorOPStPulse.gotSync == 1) 
       {
-        oilSensorOPStData.pressure = ((((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime)-128)/26.475+7.25); 
+        oilSensorOPStData.pressure = fastMap((4096UL*oilSensorOPStPulse.onTime)/oilSensorOPStPulse.totalTime, 128, 3968, 7, 152);
         oilSensorOPStPulse.index = 0;
         detachInterrupt(digitalPinToInterrupt(pinOilSensorOPSt));       
       } 
--- a/speeduino/sensors.ino
+++ b/speeduino/sensors.ino
@@ -617,7 +617,7 @@ byte getOilTemperature()
 
   if( configPage10.oilPressureEnable == 2 ) // Enabling OPS+T sensor will also enable temperature readings from it
   {
-    tempOilTemperature = oilSensorOPStData.temperature + 40; // Adding 40 to handle negative values
+    tempOilTemperature = oilSensorOPStData.temperature; 
   }
   //Sanity check
   // TODO if(tempOilTemperature > configPage9.oilTemperatureMax) { tempOilTemperature = configPage9.oilTemperatureMax; }
@@ -643,7 +643,7 @@ byte getOilPressure()
   } 
   else if (configPage10.oilPressureEnable == 2)
   {
-    tempOilPressure = oilSensorOPStData.pressure*14.5037;
+    tempOilPressure = oilSensorOPStData.pressure;
   }
   //Sanity check
   if(tempOilPressure > configPage10.oilPressureMax) { tempOilPressure = configPage10.oilPressureMax; }
@@ -739,7 +739,7 @@ void readOPSt () {
 {
   oilSensorOPStPulse.curEvent = micros();
 
-  if(oilSensorOPStPulse.lastLevel == 0 && digitalRead(pinOilSensorOPSt) ) // Last event was LOW and we have got a rising edge
+  if(oilSensorOPStPulse.lastLevel == 0 && READ_OPST_TRIGGER()) // Last event was LOW and we have got a rising edge
   {
       oilSensorOPStPulse.offTime = oilSensorOPStPulse.curEvent - oilSensorOPStPulse.lastEvent;
       oilSensorOPStPulse.totalTime = oilSensorOPStPulse.offTime + oilSensorOPStPulse.onTime;
@@ -753,12 +753,12 @@ void readOPSt () {
       } 
       else if (oilSensorOPStPulse.index == 1 && oilSensorOPStPulse.gotSync == 1) 
       {
-        oilSensorOPStData.temperature = ((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime-128)/19.2-40;
+        oilSensorOPStData.temperature = ((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime-128)/19.2;
         oilSensorOPStPulse.index++;
       } 
       else if (oilSensorOPStPulse.index == 2 && oilSensorOPStPulse.gotSync == 1) 
       {
-        oilSensorOPStData.pressure = (((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime)-128)/384.0+0.5;
+        oilSensorOPStData.pressure = ((((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime)-128)/26.475+7.25); 
         oilSensorOPStPulse.index = 0;
         detachInterrupt(digitalPinToInterrupt(pinOilSensorOPSt));       
       } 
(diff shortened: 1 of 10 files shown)


VitorBoss:
You are using floating points for calculations, please change that to integers

VitorBoss:
Include this code on `globals.h`
```
extern volatile PORT_TYPE *oilSensorOPSt_pin_port;

extern volatile PINMASK_TYPE oilSensorOPSt_pin_mask;
```

Include this code on `globals.ino`
```
volatile PORT_TYPE *oilSensorOPSt_pin_port;

volatile PINMASK_TYPE oilSensorOPSt_pin_mask;
```

Include this code on `sensors.h`
```
#if defined(CORE_AVR)

  #define READ_OPST_TRIGGER() ((*oilSensorOPSt_pin_port & oilSensorOPSt_pin_mask) ? true : false)

#else

  #define READ_OPST_TRIGGER() digitalRead(pinOilSensorOPSt)

#endif
```
Then you replace the `digitalRead(pinOilSensorOPSt)` with `READ_OPST_TRIGGER()` that is about 100x faster

speeduino/sensors.ino
```diff
@@ -707,3 +725,53 @@ uint16_t readAuxdigital(uint8_t digitalPin)
...
+        oilSensorOPStData.status = (1024.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime;
+      } 
+      else if (oilSensorOPStPulse.index == 1 && oilSensorOPStPulse.gotSync == 1) 
+      {
+        oilSensorOPStData.temperature = ((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime-128)/19.2-40;
+        oilSensorOPStPulse.index++;
+      } 
+      else if (oilSensorOPStPulse.index == 2 && oilSensorOPStPulse.gotSync == 1) 
+      {
+        oilSensorOPStData.pressure = (((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime)-128)/384.0+0.5;
```
VitorBoss, line 761: If the time is linear a simple map should do the trick without a expensive float point operation
Try `oilSensorOPStData.pressure = fastMap(oilSensorOPStPulse.onTime, 128, 3968, 1, 21);` that will be faster than floating point and increase the resolution to 0.5Bar

speeduino/sensors.ino
```diff
@@ -707,3 +725,53 @@ uint16_t readAuxdigital(uint8_t digitalPin)
...
+      oilSensorOPStPulse.lastEvent = oilSensorOPStPulse.curEvent;
+      if (oilSensorOPStPulse.totalTime <= 1024 && oilSensorOPStPulse.index == 0) 
+      {
+        oilSensorOPStPulse.index++;
+        oilSensorOPStPulse.gotSync = 1; 
+        oilSensorOPStData.status = (1024.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime;
+      } 
+      else if (oilSensorOPStPulse.index == 1 && oilSensorOPStPulse.gotSync == 1) 
+      {
+        oilSensorOPStData.temperature = ((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime-128)/19.2-40;
```
VitorBoss, line 756: `oilSensorOPStData.temperature = fastMap(oilSensorOPStPulse.onTime, 128, 3968, 0, 160) - 40;`

speeduino/sensors.ino
```diff
@@ -625,11 +640,14 @@ byte getOilPressure()
 
     tempOilPressure = fastMap10Bit(tempReading, configPage10.oilPressureMin, configPage10.oilPressureMax);
     tempOilPressure = ADC_FILTER(tempOilPressure, 150, currentStatus.oilPressure); //Apply speed smoothing factor
-    //Sanity check
-    if(tempOilPressure > configPage10.oilPressureMax) { tempOilPressure = configPage10.oilPressureMax; }
-    if(tempOilPressure < 0 ) { tempOilPressure = 0; } //prevent negative values, which will cause problems later when the values aren't signed.
+  } 
+  else if (configPage10.oilPressureEnable == 2)
+  {
+    tempOilPressure = oilSensorOPStData.pressure*14.5037;
```
VitorBoss, line 646: If you need values in PSI you must do it from TS, not here.

VitorBoss (changes requested):

speeduino/sensors.ino
gdiciocco: I've modified the calculations into the sensor ISR to produce the initial value in PSI scale as it will provide better resolution using ints and as it matches with the existing oilpressure scale.

speeduino/sensors.ino
gdiciocco: oilSensorOPStData.temperature = fastMap(oilSensorOPStPulse.onTime, 128, 3968, 0, 200); // 1-2 C less on 23C
oilSensorOPStData.temperature = fastMap((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime, 128, 3968, 0, 200); // Correct

I've tried various combinations using fastMap starting from the one you proposed with just partial success, the proportional correction of onTime based on the effective total pulse length totalTime compared to the ideal pulse length of 4096 is necessary and we don't really have a maximum value for it, as it can be less or more than the ideal (4096) and it depends on internal  sensor drifting, interfacing etc. Maybe you can help me find another way of optimization considering this.

gdiciocco:
Integrations based on VitorBoss reviews: 
direct port reads instead of digitalRead() on avr
moved floats to ints
psi scaled value produced at origin and not by subsequent conversion

speeduino/sensors.ino
noisymime: You need to get rid of those floats, this cannot be pulled in with those still there as the codebase does not allow their use.

If the second example you gave works, why not just shuffle things around to get rid of the floats? Eg

Replace:
`oilSensorOPStData.temperature = fastMap((4096.0/oilSensorOPStPulse.totalTime)*oilSensorOPStPulse.onTime, 128, 3968, 0, 200);`

With:
`oilSensorOPStData.temperature = fastMap((4096UL*oilSensorOPStPulse.onTime)/oilSensorOPStPulse.totalTime, 128, 3968, 0, 200);`

VitorBoss: On the datasheet show a minimum width of 128 and a maximum 3968, if only the high level width matters, why you have to divide 4096 by total time?

VitorBoss:
I have forgotten to tell you to include at `init.ino` at line 2585 this

 ```
  oilSensorOPSt_pin_port = portInputRegister(digitalPinToPort(pinOilSensorOPSt));
  oilSensorOPSt_pin_mask = digitalPinToBitMask(pinOilSensorOPSt);
```

speeduino/sensors.ino
VitorBoss: IMG_20201229_072756
This graph shows a period of 4096uS

IMG_20201229_072736

And those clearly shows the minimum and maximum width of a valid signal, as you can see, minimum is 128 for temperature and pressure, same for maximum of 3968

VitorBoss: If the total time is 4096 don't make any sense use this on math, `4096/totalTime` will be `4096/4096` and it can be ignored, another thing, only assume new values if S1 is smaller than 384uS

gdiciocco: > On the datasheet show a minimum width of 128 and a maximum 3968, if only the high level width matters, why you have to divide 4096 by total time?

4096 is the theoretical unit of time (theoretical totalTime) into where and in relation to which the symbol (theoretical onTime) is meaningful. 
In reality both drift (of a proportional amount), by various reasons, temperature, voltage, communication, interrupt firing time, computational time and so on. 
So, to be aware of the amount of this drift we need to compare a measured value with its fixed theoretical counterpart: totalTime and 4096, their difference will tell us the amount of drift we are having and considering that we can understand how our real onTime was affected by this drift.

We cannot directly (succesfully) use onTime because its only meaningful related to the unit of time (4096) and the unit of time is 4096 only in theory.

I've tried Josh suggested resolution compared to the original one and I saw no differences, so I'm going to implement it that way for all the three pulses.

I'd like to implement sensor fault codes on TS and oil temperature based functions but maybe better to close this one first and do that in a future pull request.

Thank you for your help!

frazierjason:
What is the model of this Hella sensor, and are there any vendors selling it these days?  Or if the Hella part number is obscure, is there an existing car or bike year/make/model/trim that features this sensor for easier lookup?  I don't see any info in the commits about what sensors conform to this code.  This combined sensor looks very interesting, but if the sensor is unobtainable, I don't have a compelling case to use it on my AATV.

gdiciocco:
> What is the model of this Hella sensor, and are there any vendors selling it these days?  Or if the Hella part number is obscure, is there an existing car or bike year/make/model/trim that features this sensor for easier lookup?  I don't see any info in the commits about what sensors conform to this code.  This combined sensor looks very interesting, but if the sensor is unobtainable, I don't have a compelling case to use it on my AATV.

Check the linked datasheet, product code is 6PR 010 378-101

mike501:
Sorry for being so slow to comment. Links to manufacturers using this part mid way down this page.
https://m.autodoc.co.uk/hella/7517971

PR #500, opened 2021-01-10
Fix Warmup Enrichment dialog responsiveness

karniv00l:
Having extra field in dialog like warmup enrichment breaks ability to properly resize the window. Instead bottom help section can be used for description.

## Before

before

## After

after


noisymime:
Nice work!

PR #519, opened 2021-02-03
couple of signed overflow fixes

dantob:


speeduino/secondaryTables.ino
```diff
@@ -88,17 +88,19 @@ void calculateSecondarySpark()
       currentStatus.advance2 = getAdvance2();
       //Spark 2 table is treated as a % value. Table 1 and 2 are multiplied together and divded by 100
       uint16_t combinedAdvance = ((uint16_t)currentStatus.advance1 * (uint16_t)currentStatus.advance2) / 100;
```
dantob, line 90: this line is still broken, what happens if we have negative values in one or both tables. they are going to be converted to unsigned values and change to positive during this multiply

speeduino/secondaryTables.ino
```diff
@@ -88,17 +88,19 @@ void calculateSecondarySpark()
...
+      //make sure we don't overflow and accidentally set negative timing, currentStatus.advance can only hold a signed 8 bit value
+      if(combinedAdvance <= 127) { currentStatus.advance = combinedAdvance; }
+      else { currentStatus.advance = 127; }
     }
     else if(configPage10.spark2Mode == SPARK2_MODE_ADD)
     {
       BIT_SET(currentStatus.spark2, BIT_SPARK2_SPARK2_ACTIVE); //Set the bit indicating that the 2nd spark table is in use. 
       currentStatus.advance2 = getAdvance2();
       //Spark tables are added together, but a check is made to make sure this won't overflow the 8-bit VE value
       uint16_t combinedAdvance = (uint16_t)currentStatus.advance1 + (uint16_t)currentStatus.advance2;
```
dantob, line 100: same here

speeduino/secondaryTables.ino
noisymime: Yep, agree completely here. If you want to make these `int16_t` values as part of this PR I'll happily pull it in.

speeduino/secondaryTables.ino
```diff
@@ -12,7 +12,7 @@ void calculateSecondaryFuel()
     {
       currentStatus.VE2 = getVE2();
       //Fuel 2 table is treated as a % value. Table 1 and 2 are multiplied together and divded by 100
-      uint16_t combinedVE = ((uint16_t)currentStatus.VE1 * (uint16_t)currentStatus.VE2) / 100;
+      uint16_t combinedVE = (uint16_t)currentStatus.VE1 + ((uint16_t)currentStatus.VE1 * ((uint16_t)currentStatus.VE2 / 100));
```
noisymime, line 15: This is not correct, there are at least 2 things going on here. 

1) The MULTIPLY mode for these tables works by using the secondary table as a percentage. 
Eg: If you have 80 in the primary table and 50 in the secondary, the resulting value should be 40 (50% of 80)

2) `((uint16_t)currentStatus.VE2 / 100))` will not work. This is all integer math, not floats, so any value below 100 will return 0, between 100 and 200 will return 1 etc.

noisymime:
I just added a comment in as it looks like you're trying to change the Multiply method on the secondary table. What was your thinking there? 

As far as I can see, it just needs the `uint16_t`s on the ignition secondary table to be changed to signed

dantob:
Ah, for some reason I had it in my head as a +- % adder like a trim table. I did forget about the floats though

it looks like it can still break if you have negative timing in the ign table and for some reason a negative value in the multiplier table. maybe force set the multiplier value to 0 if its negative?

noisymime:
Looks good now.

PR #520, opened 2021-02-03
created table3D_impl structs that contain table3D memory

brunob45:
Regarding the merge of #291 and commit 0e80fdf6d84dbe07ca2d91f3eaf94c65d16d87bf.
This is a better solution, again, but templates can be scary for non C++ people. 
I don't use virtual functions, so there is no overhead to using this feature. 

The main thing is, this is much easier to maintain and less error-prone than having multiple `#define`.

It's been a long time since I've been active here, but I hope I can still do relevant improvements.

I tested quickly, all tables are loading correctly, are editable in tunerstudio and keep their values upon reboot.

### DIFF
--- a/speeduino/globals.h
+++ b/speeduino/globals.h
@@ -360,19 +360,19 @@ extern const byte data_structure_version; //This identifies the data structure w
 extern const uint16_t npage_size[NUM_PAGES]; /**< This array stores the size (in bytes) of each configuration page */
 #define MAP_PAGE_SIZE 288
 
-extern struct table3D fuelTable; //16x16 fuel map
-extern struct table3D fuelTable2; //16x16 fuel map
-extern struct table3D ignitionTable; //16x16 ignition map
-extern struct table3D ignitionTable2; //16x16 ignition map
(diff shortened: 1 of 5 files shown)


adbancroft:
Nice.

I've been working on a similar solution, but it goes further and shaves a few more bytes of the size of a table. 

I like this though - the manual memory management was ugly.

brunob45:
Yeah, I could have removed the `columns` table and get the values with `data[i+size*y]` instead of `data[y][x]`, but I wanted to minimize the changes to the existing code

noisymime:
This was discussed a little over in #529 but my preference at the moment is to retain this as a C codebase rather the intentionally bringing in any C++. There are multiple reasons for this, but predominantly because it targets the MISRA C:2012  standard. 

Whilst it won't ever be 100% compliant, moving to C++ would mean that we would have to move away from that standard completely and begin targeting the C++ equivalent. This is not straightforward as it requires changing the scanning system, CI/CD etc. I'm not sure whether there even exists a free scanner for MISRA C++:2008 (Or newer).

PR #523, opened 2021-02-08
Reduce flash usage of CRC

adbancroft:
Swap out "FastCRC" for "[CRC](https://platformio.org/lib/show/1202/CRC32)"

- Removes ~6Kb from our Flash usage.
- calculateCRC() can be reduced to one algorithm for all pages.

Tested on a Mega 2560 with Tunerstudio v3.1..05


noisymime:
What's the impact on the loops/s from this change? The flash space isn't a huge concern, but the performance is, so it needs to be on par with the fastCRC in use at the moment. 

If you can test with a 36-1 trigger, comms data rate set to Max reads per second and do before/after checks at 2000 and 6000 RPM that would give a good indication of the change.

adbancroft:
> If you can test with a 36-1 trigger, comms data rate set to Max reads per second and do before/after checks at 2000 and 6000 RPM that would give a good indication of the change.

I don't have a stim, but I can get performance numbers (E.g. ms per CRC) for old & new with a custom sketch. Would that work for you?

calculateCRC() is only called during serial comms, so I'm surprised it's a performance focus. I guess at 115200 baud, there's only time for 138 instructions per sent byte received.

adbancroft:
Closing this one - any new solution is too complex or slow versus potential improvement.

PR #525, opened 2021-02-11
Dwell map addition

pazi88:
Added option to run 4x4 RPM/MAP based Dwell map.


diy-efi:
I've had a Tuner who requested this feature test it today on his dyno.
It solved the issue he was having with the old coil pack he was using needing a lot more dwell at lower RPM.
The dwell gauge in Tunerstudio showed what he was expecting to see.

noisymime:
Great work!

PR #526, opened 2021-02-11
for a new board, eeprom values should be set via the base tune

dantob:
it looks like disabling outputs were accidentally duplicated here (c59153a)

setting the can id was added here, its defaulted to 256 elsewhere, but for some reason 512 on a new board (02564d3)

### DIFF
--- a/speeduino/updates.ino
+++ b/speeduino/updates.ino
@@ -330,7 +330,7 @@ void doUpdates()
     {
       configPage10.flexAdvAdj[i] += 40;
     }
-    
+
     //AE cold modifier added. Default to sane values
     configPage2.aeColdPct = 100;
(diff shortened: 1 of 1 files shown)


noisymime:
I think these might've been added intentionally. My guess would be that they were potentially causing an issue on some new boards that would prevent booting (and hence the ability to load a tune) and so this was added to force disabling them.

@VitorBoss Was it intentional for these to be included twice?

VitorBoss:
For programmable outputs was, the canID was there already, I had included there again because most of users are upgrading from a previous firmware, but with random data from a first run was locking up booting for some users on all platforms(AVR, STM32 and Teensy)

noisymime:
Closing as this was done intentionally.

PR #529, opened 2021-02-13
Isolate table3D member access in storage.ino

### LINKED ISSUE #528: EEPROM storage - EEPROM_CONFIG11_YBINS in "wrong" place
`#define EEPROM_CONFIG11_YBINS 2369` 

It's not really wrong, but could be changed to 2368 & save 1 byte of EEPROM storage. But that would (should) change all subsequent storage locations, which means the **tune must be reflashed** from TS.

adbancroft:
This is one of a _series_ of PRs that will gradually isolate/encapsulate calls to table3D members, in preparation for a PR to reduce table3D SRAM usage.

This PR targets storage.ino:
- ~30% read & ~5% write performance improvements
- Much clearer code: easier to add new pages & tables
- (minor) 7% reduction in flash usage

Fixes #528: **Requires tune to be reflashed**


noisymime:
I was about to ask if you could remove the use of the undocumented EEPROM things (`EERef`) but saw that they're gone already. 

My only other real concern with this is the introduction of templates. Speeduino is, unless it can't be avoided, a C codebase rather than C++ and this is something I would like to retain. Whilst there are elements in the Arduino library code that cannot (easily) be changed, the codebase aims for MISRA:2012 compliance, which requires the use of C only. 

That said I don't see the removal of the templates here as being particularly difficult?

adbancroft:
>I don't see the removal of the templates here as being particularly difficult?

Done.

adbancroft:
> aims for MISRA:2012 compliance, which requires the use of C only

@noisymime So no C++usage at all? I ask because table iteration logic is duplicated in many places. Minimizing that duplication is possible with C-style changes like those in this PR. But the only way I can see to truly eliminate that duplication requires the use of classes - though even then I'm not sure it's possible.

noisymime:
> So no C++usage at all? I ask because table iteration logic is duplicated in many places. Minimizing that duplication is possible with C-style changes like those in this PR.

Preferably not if it can be avoided. Introducing C++ intentionally will mean that it can no longer even work towards MISRA:2012 compliance. Switching to the C++ version is a consideration, but needs scanning, remediation etc all to be in place first if that is the way we want to go.

PR #530, opened 2021-02-13
Add Ford ST170 Decoder

diy-efi:
This adds a decoder for the Ford ST170 engine.
This setup uses a 36-1 primary and a unique 8-3 secondary trigger.
The existing missing tooth decoder with fixed parameters is used as the primary to reduce duplication.

This has been tested on the bench on 2 separate boards.
A logic analyser was used to make sure it the outputs were consistently aligned with the reference tooth on the secondary at startup.

Some testing has been done on my personal car, see video on Facebook.
Angle tracking was confirmed with another ECU brand that already supports this pattern.


noisymime:
Looks good! Nice work.

PR #531, opened 2021-02-17
RTC fixes and changes

Tjeerdie:
Fixed RTC stuff for STM32 and some other fixes/upgrades
1. Remove rtc trim as it is not used (for now)
2. Cleaned up RTC setting time applet a bit (remove rtc trim)
3. The RTC applet and 'w' command did not work correctly on linux with STM32. Now it waits for RTC bytes OR times out
4. The year is communicated big endian by the RTC applet (at least by me) needs changing to little endian
5. Use correct year for STM32 rtc (2020 must be 20 etc)

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -400,10 +400,8 @@ page = 1
 
       iacTPSlimit               = scalar, U08,  121,        "%",   1,         0,     0,        100,      0               
       iacRPMlimitHysteresis     = scalar, U08,  122,        "RPM"  10,        0      10      2500,     0
-
-      rtc_trim        = scalar,  S08,    123,               "ppm",      1, 0, -127, +127, 0
-      
-      unused2-95    = array,  U08,      124, [4],   "%",        1.0,       0.0,   0.0,     255,      0
+     
+      unused2-95    = array,  U08,      123, [5],   "%",        1.0,       0.0,   0.0,     255,      0
 
 ;Page 2 is the fuel map and axis bins only
 page = 2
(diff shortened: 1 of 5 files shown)


Tjeerdie:
@noisymime we had the discussion about the 'w' command to remove the timeout I added on that one. Do you still feel this is necessary? I can remove it in this PR and set it to the original operation (but that will break it at least for the STM32 with TS in Linux I did not test how other platforms and combinations behave)

noisymime:
> @noisymime we had the discussion about the 'w' command to remove the timeout I added on that one. Do you still feel this is necessary? I can remove it in this PR and set it to the original operation (but that will break it at least for the STM32 with TS in Linux I did not test how other platforms and combinations behave)

Ultimately this needs to be tied back into the `cmdPending` mechanism that checks whether a command is already in progress. So execution can return to the main thread, but return to the same spot in the `w` command once the serial data has been received.

I know it seems like a small thing, but these loops were what used to be used in the comms code everywhere and caused intermittent missed pulses etc as it became dependant on serial comms which are very slow compared to the operating speed of these MCUs. It's something I'd prefer not to go back to if we can avoid it.

Tjeerdie:
Hi

it seems a similar problem to sdcard writing :-D. I changed it back to the
original setup, at least the rest (RTC backend) is working ok now. I will
try to fix this with a return to main loops someday. Should not be to
difficult todo.

Op di 23 feb. 2021 om 23:08 schreef Josh Stewart <notifications@github.com>:

Tjeerdie:
Its all added to the sdcard PR

PR #534, opened 2021-02-22
Basic onboard sdcard logger

Tjeerdie:
First attempt sdcard logging for STM32 and Teensy (For the mega the ram usage and overhead is too much)

Features
- On boot or disable logging
- Changing delimiter for CSV file
- Bin vs CSV style logging
- Change log rate to 1Hz, 4Hz, 10Hz, 30Hz.

TODO
- One major TODO is the asynchronous filesystem access (Non-blocking) to prevent locking up the main loop. At the moment the main loop is sometimes hold off for too long messing up the schedules (when the logger is running)
- Implement Megalog view file type for logging (mlv file type)
- Create triggered logging (the menu and settings are already in)
- Get log files of the sd-card using Tuner Studio or general mass storage over USB.

NICE TO HAVE
- Triggered log to spi eeprom  
- Safe the tune in ini format when starting a log 

@noisymime please remove the menu access in the ini for this. And existing tunes will probably enable this feature accidently because the part of the EEPROM is not used yet (and therefore set to 255). So if the EEPROM versions upped to a new version also set all settings for sdcard to zero. The default values in the ini are all set to the correct values.


Tjeerdie:
@noisymime build fails for teensy 35 and 36 i see. Does travis use the platformio.ini to see lib dependencies? Before summitting the PR i tested the teensy build in PIO and i think those worked fine..

Tjeerdie:
Part of this PR is mangled and used into the current master, but in such a way it is not compatible or to be easily made compatible with STM32 or this PR. Timing issues will be the main hurdle to take see https://speeduino.com/forum/viewtopic.php?t=4400&start=10 for an explanation. I spend countless hours on solving the timing issue i am not doing it all over again for the new setup.

noisymime:
There was a lot of good stuff that came out of this PR, so it's had some good value from it. All the TS configuration and layout has been copied over, though not all options are supported yet. 

The biggest concern I had with it was around it's portability and maintainability. The FatFS library is a fairly dated one with little active support or maintenance. It's performance is not as good as other libraries and it doesn't natively support things like ring buffers and preallocation. 

SDFat is a considerably more modern library that has support for multiple hardware platforms, including stm32. It's worth noting though that some of the stm32 hardware and driver limitations certainly hold it back with this library (See https://github.com/greiman/SdFat/issues/250). I do believe it will be workable on stm32 though as the speed/bandwidth requirements for logging here aren't particularly high (Compared to things like audio/video). All the file pre-allocation and buffering things that are in the code now are architecture agnostic, so stm32 will benefit from those immediately, but I haven't tested this performance on anything but Teensy 3.5/4.1 for now.

Tjeerdie:
We talked about this on slack but for clarity i share the same message here.

Ok, To be clear the goal of this PR was to make a generic sdcard data logger that uses only generic sdcard Arduino functions for compatibility with multiple sdcard libraries across platforms. This would mean teensy could use  sdFAT if that is better supported. 

The second important goal was to not use external hardware for the STM32 version and use the onboard slot of the Chinese dev boards. These slots use the SDIO interface of the MCU. If external hardware is needed it defeats the purpose. You could make a external logger using the secondary serial interface using external hardware, that simplifies the software part a lot. 

Now by choosing to use the sdfat liberary for teensy and using sdFAT library functions not available in fatFS (exFAT file system + file pre-allocation) goal nr1 is broken. FatFS library might be old but ST supports the SDIO/SDMMC-hardware for STM32 using this library. And it is somewhat maintained. Is it ugly? YES! Is it hard to maintain because we need changes in the library YES! But it has a basic generic sdcard interface on the arduino speeduino FW side that could have been maintained. 

Now the other option to regain STM32 compatibility would be to use the sdFAT library for STM32. But as very clearly shown in the greiman/SdFat#250 the maintainer is not willing to implement this for the  SDIO/SDMMC-hardware. Leaving the SPI as the only easy option to use sdFAT on STM32. But that requires external hardware with an Sdslot on the SPI interface. This breaks goal nr2  As far as i know the SDIO/SDMMC-hardware is not accessibly as if it is a SPI interface. Also the pins are not reconfigurable to an SPI interface. On top of that this is assuming there is no long latency blocking writes using the SPI interface. (Bandwidth has never been an issue, speeduino its low bandwidth but it should also be low latency. <5ms maximum blocking per SDlib function call.)

So I would suggest the following for anyone wanting to implement this to regain use full STM32 sdcard data logging on the internal sdcard slot.
Implement the SDIO/SDMMC-hardware interface for sdFAT. And while doing so extensively test the time each sd write takes. I am not saying it is definitely a problem it depends highly on the sdFAT library, but it should also not be ignored. don´t forget to test latency when making a new file if the first file is full. I suspect this is the hardest part in latency, but not sure. higher latency would manifest itself as missing single fuel injection pulses, so its pretty hard to spot if this happens only once every minute or so. This route would benefit both sdFAT and speeduino.

noisymime:
I agree that the SDIO driver path is the way to go for any boards where SPI is not an option. It looks like there have been a few attempts at this (Eg https://github.com/stevstrong/Arduino_STM32/tree/master/STM32F4/libraries/SDIO) 

Just tossing ideas out there, but would software SPI over the SDIO pins be an option at all (If hardware SPI pins cannot be reassigned)?

Tjeerdie:
Software SPI would even be slower than SPI on the STM32 (which apparently is already pretty slow). So i would not expect much from it. But if someone is willing to take the time and setup some bench test to show the even higher latency and CPU intensiveness of software SPI is not an issue for speeduino who knows.

noisymime:
I think it's worth trying (Alongside the SDIO driver option), the logging requirements here aren't that demanding (Even at 30Hz logging it's under 4KB/s), we just need to be careful of the impact to other areas of the firmware.

Ultimately this comes down to a being a fairly specific list of hardware limitations for some stm32 boards that we're trying to overcome:

1. SDIO performance on the pre-H7 stm32 MCUs (such as the stm32F407) being severely hampered by their small hardware FIFO (128 bytes)
2. A specific board (The stm32F407 black) only having the SD slot connected via SDIO rather than with SPI as well
3. No ability to reallocate SPI pins to use the hardware SPI on the above board

I am more than happy to look at options for overcome these limitations, but I feel weighing down the entire project with an older and modified SD library just to overcome these problems on 1 hardware configuration isn't worth it.

pazi88:
Should the current SD logger in the speeduino FW work if the SD card slot is changed to use hardware SPI pins instead of SDIO? Haven't looked how hardware specific current implementation is.

IMO the F407 black boards aren't really that good for Speeduino use, because of many reasons. So if the SPI SD-card support can be acheived "easily" I think it's worth trying.

noisymime:
> Should the current SD logger in the speeduino FW work if the SD card slot is changed to use hardware SPI pins instead of SDIO? Haven't looked how hardware specific current implementation is.

It's not currently enabled as I have no stm32 hardware with a SPI SD slot to test this on right now, but it's very simple to enable this. All the needs to happen is to add a `SD_CONFIG` definition to tell SDFat to use SPI, Eg:

```#define SD_CONFIG SdSpiConfig(SD_CS_PIN, SHARED_SPI, SPI_CLOCK)```

And then declare `#define SD_LOGGING`.

That should be it on the firmware side (You will need to update the ini to make the option visible for stm32 though)

I did just notice that there's already an example in SDFat for using software SPI, so testing that might not be too difficult: https://github.com/greiman/SdFat/blob/master/examples/SoftwareSpi/SoftwareSpi.ino

pazi88:
Thanks! I will try out how well it works out, if it will at all. It's not big task to change the sd-card slot to use spi pins. I don't think the software spi is even eorth trying.

Tjeerdie:
> Should the current SD logger in the speeduino FW work if the SD card slot is changed to use hardware SPI pins instead of SDIO? Haven't looked how hardware specific current implementation is.
> 
> IMO the F407 black boards aren't really that good for Speeduino use, because of many reasons. So if the SPI SD-card support can be acheived "easily" I think it's worth trying.

If you have the hardware to test please do :-D. To get it working should be fairly easy. Please thoroughly check compatibility with spi EEPROM and also check the latency issue very very well. As it is easy to miss.

pazi88:
Modifying what I have currently isn't really big task to do. And yeah naturally checking the latency is the first thing to check when/if it works.

noisymime:
Also if there's nothing else on your SPI bus, you can use:

```#define SD_CONFIG SdSpiConfig(SD_CS_PIN, DEDICATED_SPI, SPI_CLOCK)```

It should significantly improve bandwidth

PR #536, opened 2021-02-23
New VR design in KiCAD

TechColab:
more compact, adds mounting holes & optionan regulator & pin headers


noisymime:
This looks good. There's a few small changes I would like to make to it based on some experience I've had with the existing boards, but overall I am happy to make this the next formal revision of the VR.

Are you comfortable with this being released under the current open license used by the Speeduino hardware?

TechColab:
Yes, I'm happy with the same licence as Speeduino hardware.  I will look 
at your mods later todat.

noisymime:
The largest change I made was to increase the input resistors to 1206 and 1/2W. Under some (very) exceptional circumstances I have seen issues with the 1/8W or 1/4W resistors.

TechColab:
I've just finished an update (Rev. 3.7) to give more space to the bigger resistors & tidy symmetry a bit - it's now same width & height and still under 20mm.  I think I just did a pull request but I'm still fumbling with GitHub.

DeionSi:
I think this belongs in https://github.com/speeduino/Hardware

PR #539, opened 2021-03-02
[WIP] Initial high resolution crank

VitorBoss:
**W**ork **I**n **P**rogress
This simple macros shifts the angles increasing the resolution 8 times for spark tables interpolation, this gives an accuracy of 0.125º,  it _can be disabled_ by setting the const `ANGLE_SHIFTS` to `0` _without any other changes to the code_, right now the focus is exclusively at angles math but I could see come differences at injectors pulses as well.

The idea was to keep as simple as possible to better reading and understanding of the code.

@noisymime As discussed at slack I,m opening this in the regard of mutual collaboration on this work in progress code

I hope @Tjeerdie, @diy-efi, @iLeeeZi, @pazi88, @Autohome2 and many others find this interesting and share ideas to improvement.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -18,6 +18,7 @@ lib_deps = EEPROM, Time
 test_build_project_src = true
 debug_tool = simavr
 
+
 [env:megaatmega2561]
 platform=atmelavr
 board=ATmega2561
--- a/speeduino/auxiliaries.ino
+++ b/speeduino/auxiliaries.ino
@@ -308,8 +308,8 @@ void vvtControl()
     if( (configPage6.vvtMode == VVT_MODE_OPEN_LOOP) || (configPage6.vvtMode == VVT_MODE_ONOFF) )
     {
       //Lookup VVT duty based on either MAP or TPS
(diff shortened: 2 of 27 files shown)


VitorBoss:
Right now the side effects are a 10% hit on the loops counter due to changes in the angles variables sizes, the hit 7% for 0.5º accuracy as well, didn't test with more accuracy yet but that should't increase another 4% for 0,015625º(1/64)

mike501:
Before you take this code too far, have you measured the accuracy of the spark timing code? I did a lot of investigation at the end of last year which helped identify the problem with spark, 5,6,7,8 timing then subsequently fixed in 202012. I don't have my figures to hand but as I recall, trying to improve above approx. 0.25 of a degree won't have any effect as the existing code inaccuracy will overrule your improvements.

pazi88:
@mike501 was that on Arduino mega, or with some other mcu. Indeed, with arduino mega the extra timing accuracy isn't really needed. But with stm32 for example it looked like it could really take advantage of increased accurasy. But I haven't yet tested how well this PR actually improves situation.

VitorBoss:
Yes, I did teste my claims, the key here to improve it is change the return values from 3dtable interpolation to match the resolution, if you just increase the angles values without changing all angle related stuff like corrections you won't get any accuracy at all.

PR #542, opened 2021-03-11
[WIP] Isolate table3D member access in comms.ino

### LINKED ISSUE #541: comms.ino: receiveValue(): could overwrite random memory
wmiMapPage has a 80 byte "gap" in it's virtual addressing between wmiTable & dwellTable. There is no protection if TS is configured incorrectly and tries to write a byte in that gap: the write will go through & overwrite some random memory location.

adbancroft:
Another PR to isolate/encapsulate calls to table3D members, in preparation for a PR to reduce table3D SRAM usage.

This PR targets comms.ino:
- Move page related functions & definitions into pages.h, pages.cpp
   - `getPageValue `
   - `setPageValue` (was `receiveValue`)
   - `calculateCRC32`
- Refactor `getPageValue `& `setPageValue`: extract the code that maps an offset to an entity & share between the get/set
- Refactor `calculateCRC32 `to use one algorithm for all pages
- Centralize access to axis factors
- Much clearer code: easier to add new pages & tables
- 18% perf improvement in `setPageValue`
- 25% perf improvement in `calculateCRC32`
- (minor) 4% perf improvement in getPageValue 
- (minor) 8% reduction in flash usage

Because I went with a .cpp file, it revealed One Definition Rule violations in table.h, board_stm32_generic.h, board_stm32_official.h, board_teensy35.h, board_teensy41.h, utilities.h. 

Fixes #541


adbancroft:
This turned out to be way bigger than I anticipated!

PR #549, opened 2021-03-22
Enable use of disabled functions pins

VitorBoss:
A lot of beginners struggles with unused pins/outputs been forbidden to use if the function is disabled, this replaces some macros with a function which checks if the pin function is enable, the default pin definition have priority, in case of function activation it will function normally.

That allow for easier use of unused FETs on the boards.


shiznit304:
This is awesome! My UA4C board only has 4 dedicated outputs. This should hopefully allow me to use all the other pins (Idle2, Ign3, Ign4) that I'm not currently using for a different purpose

shiznit304:
Any chance of this making it into master? I have helped several Miata owners who have the PnP boards wire up the boost pin or ign3,ign4 pin to do something other than boost control or ignition. I have them wire the boost pin to run the aircon relay using programmable outputs but can't do it without this PR.

AlessandroEmm:
Another 👍  for this one. I was trying to reassign High Power FETs for Idle (dont need idle control), but I couldnt since they are not exposed through TS. So only thing was to roll my own pin assignment. :-/

smb29:
sewer rats

PR #550, opened 2021-03-24
Update misra texts

VitorBoss:


noisymime:
Unfortunately we cannot include the word for word definitions from the official MISRA file like this. They are copyright and strictly not authorised for distribution in open source projects. 

This is why the rules in the current txt file are using definitions written by myself in my own words and are only done for the most common issues.

PR #551, opened 2021-03-26
VVT2 addition and closed loop VVT improvement

### LINKED ISSUE #541: comms.ino: receiveValue(): could overwrite random memory

pazi88:
Third time's the charm..

This PR adds remaining things to get the VVT2 working. VVT2 mainly uses same settings as VVT1 for now. Separate settings for VVT2 can be added if those are needed.

But this PR also includes lot of improvements on closed loop VVT to make it work **a lot** better. This is a result of three weeks of testing and improving the closed loop VVT in car:

- VVT PID gains aren't anymore frequency dependent as they were earlier (this made lower VVT frequencies unusable for closed loop VVT)
- The direction adjustment now affects the PID direction and not just invert the VVT output(s). Inverting output(s) didn't work at all in practice.
- The duty and angle accuracy is now 0.5 instead of 1, without adding extra computing load. This makes the closed loop to work more accurately and improves also open loop VVT. 
- The PID now uses different mode where Proportional gain affects both changing target angle and actual angle. Previously the Proportional gain only affected changing actual angle, but not changing target and this arrangement didn't work at all in practice.
- Added filtering for cam angle to smooth out readings and in that way calm PID adjustment so that bigger PID gains can be used without causing oscillation = faster cam response.
- Added safety limits for cam angle that in closed loop mode drop the VVT duty to 0% if limits are exceeded. This makes the car drive able even if there is some kind of problem in cam angle reading.
- Added indicators for previously mentioned VVT safety limits to show if the VVT control is in error state or not. For this the wmiEmpty indicator byte has now been changed to proper indicator bit. The wmiEmpy indicator has been tested to work after this change.
- Fix problems in writing log, where values for VVT1 and 2 weren't written properly.
- PID improvements from @VitorBoss to make it faster and reduce duplicate code. These or other more minor changes are mandatory to make the PID work in reverse direction too.
- Also few really minor bug fixes that I did come across while doing this.

Closes #541


pazi88:
Closes #541

pazi88:
Been using this for over a week now. The dual closed loop vvt works great. 
image

diy-efi:
ST170VVT_Log
Tested on my ST170 engine, VVT works much better than before.
It did need the cam offset angle changing, but I can't see that effecting many people as the ST170 pattern is so new anyway.

PR #553, opened 2021-03-28
STM32 fixes and tweaks

VitorBoss:
Removed generic STM32F103 options from PlatformIO
Added optional serial compilations on PlatformIO
Make sure all timers are stopped at boot
Small speed improve at SPIasEEPROM
Added timers usage table to board_stm32_official.h


noisymime:
Was this meant to include the hardware additions or were they added by mistake?

VitorBoss:
Yeah, I meant to include it, if here isn't the best way of do it I can open a separate PR for it

noisymime:
All good, I just wanted to make sure you weren't publishing something that you didn't actually want to be there.

PR #562, opened 2021-04-22
Use the TunerStudio page layout for EEPROM storage

### LINKED ISSUE #527: EEPROM storage - EEPROM_CONFIG9_START in "wrong" place
`#define EEPROM_CONFIG9_START  1710` 

It's not really wrong, but could be changed to 1701 & save 9 bytes of EEPROM storage. But that would (should) change all subsequent storage locations, which means the **tune must be reflashed** from TS.

adbancroft:
### This PR requires the user to reflash their tune

The goal of this PR is to remove some direct usage of `table3d` members, this time in `comms.ino` (this the penultimate PR before shrinking 3D table .bss usage).

Changes
- (Major) Use page & table iterators for EEPROM I/O (this is the last place where I/O was hard coded).
  - This effectively reads and writes the TS page layout
  - Pages are arranged end-to-end automatically.
  - Now all that's needed for a new page or table is to update the page mapping (in `pages.cpp`): CRC, EEPROM & Serial comms will all "just work".
- The rest of the PR is a minor cleanup of other storage functions: encapsulate duplicate code etc.

Fixes #527

### DIFF
--- a/speeduino/board_avr2560.h
+++ b/speeduino/board_avr2560.h
@@ -15,8 +15,10 @@
   #define COUNTER_TYPE uint16_t
   #ifdef USE_SPI_EEPROM
     #define EEPROM_LIB_H "src/SPIAsEEPROM/SPIAsEEPROM.h"
+    typedef uint16_t eeprom_address_t;
   #else
     #define EEPROM_LIB_H <EEPROM.h>
+    typedef int eeprom_address_t;
   #endif
   #define RTC_LIB_H "TimeLib.h"
   void initBoard();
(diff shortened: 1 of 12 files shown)


adbancroft:
@noisymime Any feedback on this one? Thx

neovin87:
does this fix the issue with differing write and read addresses or what i can make of the code?

For example the programmable outputs, is not an issue in 202008 version but in current and master dev cannot get any change in tunerstudio to register

adbancroft:
@neovin87 What symptoms are you seeing?

In master the programmable outputs are stored in config13, which is read & written from address EEPROM_CONFIG13_START. I don't see anything else written in that location.

adbancroft:
Closing - see #620

PR #564, opened 2021-04-23
Fix and cleanup of PR#545 for STM32F401 build

Tjeerdie:
As per title and comments in Slack.


Tjeerdie:
already fixed in newer PR

PR #572, opened 2021-04-28
Add RPM switch point for instantaneous map sample method. 

pazi88:
What: This simple PR adds RPM switch point for instantaneous MAP sample method. So that below this set RPM instantaneous MAP sample method is used instead of IE. cycle average.

Why: For example for 4-stroke engines cycle average method gives smoothest and most stable map readings, which is preferred. Unfortunately at low RPM close to idle, the cycle time is actually quite long and especially at idle the MAP readings change really slowly. So slowly that it affects quite a bit for throttle response, making the engine slow to react throttle input at low RPM. TPS AE helps with this, but even with it tuned right, the result usually could be better. If instantaneous MAP sample method is used, throttle response at low RPM is way better than with cycle average. But unfortunately the MAP reading at higher RPM gets usually really noisy and this can negatively affect many things. Like MAP based AE and/or closed loop VVT if those are used. With this rather simple PR, you can get best of both methods. Instantaneous MAP readings at low RPM for fast throttle response and steady MAP readings at high RPM. I have personally used this small addition for weeks with 1500-1700RPM switch point and the engine behaves better than it has ever done.


noisymime:
Tested this and all checks out. I think the only suggestion I'd make is to switch out the 64 factor on `MapSwitchPoint` and make it 100. I understand the intent here in order to do efficient divides, but using a multiply x 100 (In place of a divide) will not have any real impact on performance. 

And just for the sake of consistency, can you switch `MapSwitchPoint` to have a lowercase m at the start? 😄

pazi88:
> Tested this and all checks out. I think the only suggestion I'd make is to switch out the 64 factor on `MapSwitchPoint` and make it 100. I understand the intent here in order to do efficient divides, but using a multiply x 100 (In place of a divide) will not have any real impact on performance.
> 
> And just for the sake of consistency, can you switch `MapSwitchPoint` to have a lowercase m at the start? 😄

No problem. I just thought I will use the bit shift to reduce ram usage. But multiply x 100 is fine for me if you think it's better. And yeah I will fix that typo.

PR #573, opened 2021-04-28
Boost by Gear

DylanRooijakkers:
This pull request adds the functionality of boost by gear.

Currently there are 2 different modes:
    1. Multiplied %
        In this mode the values specified per gear are used as a percentage value to the boost target table.
    2. Constant limit
        This mode sets a constant boost target per gear in kPa and neglects the boost target table.

This feature relies on the VSS input to determine the current gear.
Boost control has to be in closed loop.

This might not be very usefull for everyday driving, but it might come in handy at the dragstrip.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -972,10 +972,10 @@ page = 9
       iacStepperInv       = bits,   U08,   153, [0:0], "No",        "Yes"
       iacCoolTime         = bits,   U08,   153, [1:3], "0", "1", "2", "3", "4", "5", "6","INVALID"
 
-      boostByGearEnabled  = bits,   U08,   153, [4:4],      "Off", "On"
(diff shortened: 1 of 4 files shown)


pazi88:
This seems quite simple and well done PR. Few things came to my mind. Maybe reserve at least 2 bits for the `boostByGearEnabled` bit. And mark other 3 modes as INVALID in TS config for now. So that later we would for example have different boost by gear modes which would for example be % of boost target map value. And secondly the *2 can be made faster by using bit shift by one ( `<<  1` )

DylanRooijakkers:
@pazi88 Thank you for your comment, I really appreciate it. I have now made the changes to the PR according to your advice. Hopefully this gets pulled in soon as I'm aware that a lot of people have been waiting for this feature.

noisymime:
Looks good. Basic testing and everything seems to show as working 👍

PR #574, opened 2021-05-01
Prevent looping though an array like fsIntIndex which only contains t…

Zeiberstein:
Prevent looping though an array like fsIntIndex which only contains the indexes in fsIntIndex that refer to double-byte values.
By introducing a (bit-)array (fsValueIndex) for each element it is possible to define per index whether it is single-byte versus double-byte.
The original fsIntIndex expired.

P.S. This is my first pull request ever. Please give feedback if I should have done anything differently.


adbancroft:

PR #575, opened 2021-05-05
Bugfix on Programmable outputs: 1) Delayed activation not shown in TS 2) Inverse logic not shown in TS

Zeiberstein:
There are some issues with the passing of Programmable Output values to TunerStudio. 

1) 
When the conditions for a programmable Output Rule are fulfilled AND a delay is set for this rule, the actual output pin gets its new value after the delay has passed. That is expected behavior. 
But the corresponding value in TS gets its new value already on the moment the conditions get fulfilled (regardless of the delay). So that is too early.  

2)
When an output rule uses inverted logic, it is reflected in the value presented on the output pin. This is expected behavior.
But the corresponding values in TS always are the values without inverted logic. So the value in TS does not reflect the right values in case of inverted logic. 

And I took the liberty to clean up the code of ‘checkProgrammableIO()’ in ‘utilities.ino’. If that is too much or unwanted, I can make a PR with only the essentials of this bugfix.

### DIFF
--- a/speeduino/utilities.ino
+++ b/speeduino/utilities.ino
@@ -108,83 +108,94 @@ void setResetControlPinState()
 //*********************************************************************************************************************************************************************************
 void initialiseProgrammableIO()
 {
-  for (uint8_t y = 0; y < sizeof(configPage13.outputPin); y++)
-  {
-    if ( (configPage13.outputPin[y] > 0) && (configPage13.outputPin[y] < BOARD_MAX_DIGITAL_PINS) )
-    {
-      if ( !pinIsUsed(configPage13.outputPin[y]) )
-      {
+  for (uint8_t y = 0; y < sizeof(configPage13.outputPin); y++)  {
+    if ( (configPage13.outputPin[y] > 0) && 
+         (configPage13.outputPin[y] < BOARD_MAX_DIGITAL_PINS) ) {   // Note: no need for former check on '(configPage13.outputPin[y] <= 128)' in checkProgrammableIO()
+      if ( !pinIsUsed(configPage13.outputPin[y]) )  {                                                             
         pinMode(configPage13.outputPin[y], OUTPUT);
-        digitalWrite(configPage13.outputPin[y], (configPage13.outputInverted & (1U << y)));
         BIT_SET(pinIsValid, y);
+        procesOutputValueConditionally(y, false );                  // Assume conditions are not met at powering up. 
+        if (configPage13.outputDelay[y] >= 255){                    // Note: In 202103 there was a check on outputDelay[y] being smaller than 255. Just making it smaller than 254 is more robust
+          configPage13.outputDelay[y] = 254; 
+          } 
       }
     }
   }
 }
 
+/*
+ * This procedure is called every 0.1 sec. It checks the conditions for each active Programmable Output Rule.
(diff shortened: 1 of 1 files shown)


VitorBoss:
Had you checked the outputs? TS values don't read the pin status, if only that is the trouble you can move the bit set/clear with the pin change.

Now, if your intention was to make is more clear and readable I need to point some things out.
`pinIsValid` is a variable to hold which output pin is a valid pin to avoid do all calls on an invalid pin.
At `compare` the `if` comparison don't need to occur on each line, add a `else if` on the other lines or like this:
```
bool compare(int16_t data1, int16_t data2, byte comp) {
  if ( (comp == COMPARATOR_EQUAL)          && (data1 == data2) )
  || ( (comp == COMPARATOR_NOT_EQUAL)      && (data1 != data2) )
  || ( (comp == COMPARATOR_GREATER)        && (data1 >  data2) )
  || ( (comp == COMPARATOR_GREATER_EQUAL)  && (data1 >= data2) )
  || ( (comp == COMPARATOR_LESS)           && (data1 < data2) )
  || ( (comp == COMPARATOR_LESS_EQUAL)     && (data1 <= data2) )
  { 
    return true;
  }
  return false;   // shouldn't happen
}
```
Also the return on each `if` statement will trigger MISRA errors, use a `bool` variable and a single return point instead
All looks very good, just a small tip for speeduino code, please read this: http://speeduino.com/wiki/index.php/Style_code

Zeiberstein:
@VitorBoss Excellent feedback. And thank you for your guidance. :-) 

Background of this PR:
I wanted a long delay for my AFR sensor. I powered this sensor with Programmable_Output_1: 'rpm>600' with a delay of 20 seconds. 
In de logs I saw that PrgmOut1 got enabled at the moment rpm>600, BUT the AFR sensor got activated 20 seconds later. So I concludes that PrgmOut1 and the arduino's output pin for this programmable output were not properly synced as they should do. 
I checked this on my desk with an Mega fed with an increasing rpm. I checked TS's  'Programmable Output 1' AND a led connected to the configured, corresponding output pin.

shiznit304:
image

Zeiberstein:
@shiznit304 

It's unlikely there is een general compilation problem with this Pull Request, since:

- This Pull Request passed the automated build checks, so it compiles and can (technically) be build into Speeduino's master branch.  
- The declaration of the 'missing' `procesOutputValueConditionally `can be clearly seen on line 207
- image

You might consider looking into your second error message error downloading `https //downloads.arduino.cc/packages/package_index.json`. Googling into that, there are suggestions that this is caused by a misconfigured Arduino IDE setup.   

If you would accept any help on this, I'd like to look into this together with you. But in a pm and not in this thread.

noisymime:
Is this PR still required now that #589 is merged?

Zeiberstein:
> Is this PR still required now that #589 is merged?

@noisymime  
I tested @VitorBoss 's #589 Pull Request specific for this #575 issue (concerning that the actual status of the programmable outputs wasn't reflected correly in TunerStudio). I used the same setup as described above on may 6th. 

Testresult: the reported #575 bug doesn't occur any more in Vitor's reworked code.
This #575 PR isn't needed any more and can be closed. Thanks for fixing.

adbancroft:

PR #577, opened 2021-05-10
Generate the page ASCII output code

adbancroft:
Parse the INI file and generate `printPageAscii()`.

Parsing code is:

- Python (cross-platform, widely used)
- Integrated into the PIO build pipeline

We will need to keep the code & INI file more closely synchronized. But that's a good thing.


reference/speeduino.ini
```diff
@@ -275,8 +275,8 @@ page = 1
 
       reqFuel       = scalar, U08,      24,        "ms",        0.1,       0.0,   0.0,    25.5,      1
       divider       = scalar, U08,      25,        "",          1.0,       0.0
-      alternate     = bits,   U08,      26, [0:0], "Simultaneous", "Alternating"
-      multiplyMAP1   = bits,   U08,      26, [1:1], "No",       "Yes"
+      injTiming     = bits,   U08,      26, [0:0], "Simultaneous", "Alternating"
```
noisymime, line 278: Unfortunately `alternate` is a reserved variable name used by TunerStudio. Changing this breaks the `req_fuel` calculator and so will need to be left the same.

reference/speeduino.ini
```diff
@@ -286,7 +286,7 @@ page = 1
 
       ; Config1
       mapSample     = bits,   U08,      36, [0:1], "Instantaneous", "Cycle Average", "Cycle Minimum", "Event Average"
-      twoStroke     = bits,   U08,      36, [2:2], "Four-stroke", "Two-stroke"
+      strokes     = bits,   U08,      36, [2:2], "Four-stroke", "Two-stroke"
```
noisymime, line 289: Same as with `alternate` above. This is a reserved variable name and must be kept (unfortunately)

adbancroft:
@noisymime Is there a list of the TS reserved words somewhere?

Hilariously, TS didn't give me any errors/warnings :-(

reference/speeduino.ini
adbancroft: Fixed

reference/speeduino.ini
adbancroft: Fixed

adbancroft:
Reworked this PR to improve back compat & reduce churn: no name changes to INI entry names or codebase

PR #578, opened 2021-05-10
Optimizations and refactoring of ignition schedules

TBAMax:
Made ignition timing be calculated in the schedulers from the IgnitionEndAngle. Should be more accurate because IgnitionEndAngle is directly the angle that the spark is suppose to be fired at.

Improvements to the dwell timing: 
-Dwell time is directly applied to timers now without intermediate conversion to the dwell angle domain and back. 
-Dwell time is controlled entirely in the interrupts, so even when main loop completely locks up, it do not cause changes to the dwell time.

Also reduced code repetition and made all ignition channels to use the same function. To get them all channels accept the changes with less effort. 

Cons: Unknown effect to the per tooth ignition timing ,at the moment!

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -62,7 +62,7 @@ board = black_f407ve
 lib_deps = stm32duino/STM32duino RTC
 board_build.core = stm32
 build_flags = -std=gnu++11 -UBOARD_MAX_IO_PINS -DENABLE_HWSERIAL2 -DENABLE_HWSERIAL3 -DUSBCON -DHAL_PCD_MODULE_ENABLED -DUSBD_USE_CDC
-upload_protocol = dfu
+upload_protocol = stlink
 debug_tool = stlink
 monitor_speed = 115200
 
@@ -145,9 +145,9 @@ monitor_speed = 115200
 
 [platformio]
 src_dir=speeduino
-default_envs = megaatmega2560
+;default_envs = megaatmega2560
 ;The following lines are for testing / experimentation only. Comment the line above to try them out
-;default_envs = black_F407VE
+default_envs = black_F407VE
 ;default_envs = teensy35
 ;default_envs = teensy40
 ;env_default = LaunchPad_tm4c1294ncpdt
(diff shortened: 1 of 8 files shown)


mike501 (commented):
Downloaded this version and I'm seeing issues with spark 1. Spark 2 to 4 look good. Run with ardustim generating data on 36-1 with single tooth cam. Spark 1 drops out for a few spark events, when it restarts it usually has a very large spark for the first event. Will post image of logic level output on slack of what i can see.

TBAMax:
Now the issues that was initially with running on Amega2560 platvorm are fixed.
Also tested updates from this pull request on actual car with running Speeduino v0.4 board (Arduino Mega).

TBAMax:
> Those also need to be `COMPARE_TYPE`.

@VitorBoss sorry, what exactly do you mean also need to be COMPARE_TYPE? 
 Now I was looking the `inline uint32_t getIgn1Counter();` small functions. Oh, yes those too!

TBAMax:
Currently seems this has broken fixed cranking override on some decoders. Under investigation.

TBAMax:
At the moment this is still a draft because some bugs were discovered and need to be fixed here. All the fixes are already done in the https://github.com/TBAMax/speeduino/tree/mainbranch but need to be brought over. Can not do this by simple cherry-pick but must manually copy paste over because in the mainbranch there are also changes to the fuel chedulers. Thinking of maybe doing a joint PR for new fuel and ignition both, but unsure if there is any interest at the moment.

TBAMax:
@mike501 left a good comments on the slack about how to test the ignition jitter with the Sigrock PulseView. Too bad it is gone from there by now. Need to do some tests and figure that out again.

TBAMax:
Replaced by #804

mike501:
The comments may be gone, but I am not! Give me a shout on Discord or Slack and I'll figure out what I told you before. It will have been along the lines of set up a 'timing' decoder for each ignition channel, remove average, select 1 edge (usually falling) and delta. You can then see how much the individual ignition event moves on that specific channel by looking at the delta. To improve my debugging I also added some code to turn on a single specific pin every time any ignition event occurred & set a flag high. I then added code in the main loop to turn off the pin if the flag was high and reset the flag. This allowed me to monitor every ignition event compared to each ignoring the specific ignition channel via another 'timing' decoder.

PR #585, opened 2021-05-26
Update table.ino

Zeiberstein:
Simplify interpolation for table2D data.

### DIFF
--- a/speeduino/table.ino
+++ b/speeduino/table.ino
@@ -160,19 +160,13 @@ int table2D_getValue(struct table2D *fromTable, int X_in)
     int16_t yMax = table2D_getRawValue(fromTable, xMax);
     int16_t yMin = table2D_getRawValue(fromTable, xMin);
 
-    //Float version
-    /*
-    int yVal = (m / n) * (abs(yMax - yMin));
+    /* Float version (if m, yMax, yMin and n were float's)
+       int yVal = (m * (yMax - yMin)) / n;
     */
-
+    
     //Non-Float version
-    int16_t yVal = ((long)(m << 6) / n) * (abs(yMax - yMin));
-    yVal = (yVal >> 6);
-
-    if (yMax > yMin) { yVal = yMin + yVal; }
-    else { yVal = yMin - yVal; }
-
-    returnValue = yVal;
+    int16_t yVal = ( ((int32_t) m) * (yMax-yMin) ) / n;
+    returnValue = yMin + yVal;
   }
 
   fromTable->lastInput = X_in;


Zeiberstein:
In the Facebook groep ‘Speeduino - Open Source, low cost engine management’ someone relayed a posting from Discord where was stated there was a problem with a casting in line 169 of table.ino in the calculation of an interpolated value from a 2D table. 
See [this FB-link](https://www.facebook.com/groups/191918764521976/permalink/1428913560822484/) . I agree with the author that there might be a problem with larger values of m, but I don’t know if that would occur in real life. 


As I looked into the code, I noticed that this calculation can be made much more consize  and precise, thereby preventing the bit-shifting in the current calculation (and preventing possible problems with larger values).


**Elaboration:**
For the interpolation a mathematical expressing must be calculated: `(A/B)*C`
When implementing this as int’s  `{ (intA/intB)*intC } `, the division gets problematic. 
Fill in: intA=1, intB=3 and intC=27.  The integer division `(1/3) `would yield zero and so the end result would be zero. 
The mathematical precise expression would give `((1/3)*27)=9`. 
In the Speeduino code this was solved by multiplying intA with 64 before the division and divide the result afterwards with 64. {` (((intA*64)/intB)*intC)/64 }`
With the same numbers as above, this would give: `(((1*64)/3)*27)/64 = (21 * 27)/ 64 = 8 `

But, this multiplying and dividing by 64 is not needed. 
Look at the mathematical expression : `(A/B)*C`. That can also be phrased as `(A*C)/B.`
When we do the integer calculation this way `(intA*intC)/intB`, we would get: (`1*27)/3=9`
Since we multiply to int’s in de numerator, we must be prepared that this product doesn’t fit in an int. So we will have to do this multiplication as long’s. 

**Test Code**
```
int16_t int_a,int_b, int_c, int_result;
double dbl_a, dbl_b, dbl_c, dbl_result;
uint32_t counter = 0;

void loop() {
   int_a= random(-32768,32767);
   int_b= random(-32768,32767);
   int_c= random(-32768,32767); 
   dbl_a = int_a;
   dbl_b = int_b;
   dbl_c = int_c;

   // calculate the reference result with the slow but accurate double calculation
   dbl_result = (dbl_a / dbl_b) * dbl_c;

   //Plausibility check. With realistic choices for the input's, the result must fit in an int16_t (since it is an interpolation between two existing int16_t's).
   // If that's not the case,the random input choices were too gross and we can skip this round 

   if (abs(dbl_result) <= 32767) { 
   
     int_result = ( ((int32_t) int_a) * int_c ) /  int_b;
   
     if ( abs(dbl_result - int_result) > 1) {
       Serial.print(int_a);
       Serial.print(" ");
       Serial.print(int_b);
       Serial.print(" ");
       Serial.print(int_c);
       Serial.println(" ");
       Serial.print(dbl_result);
       Serial.print(" ");
       Serial.println(int_result);
       Serial.println();
      }
     }
     if ( (++counter %10000) == 0) {
       Serial.println(counter);
     }
}

```

noisymime:
Great work in finding this. I'm always a little nervous in modifying these sections of the code as they can have some significant consequences if they break an edge case, but this looks to be good. I agree it's a much nicer implementation. 

For reference here is a small test application for comparing the old and new routines that I used for testing: https://replit.com/@noisymime/2D-Lookup-change#main.c

Zeiberstein:
> I'm always a little nervous in modifying these sections of the code as they can have some significant consequences if they break an edge case, but this looks to be good. I agree it's a much nicer implementation.
I did this often back in in my 'assembly days'. First do all calculations that won't give any loss of accuracy (multiplication) and end with the calculations that will give rise to accuracy loss (like integer division).

PR #588, opened 2021-06-02
Added values to CANcomms

adamg-86:
I added missing status values to the CANcomms to able to use them on a display or other devices.


Autohome2:
This needs some additional checks added.
It breaks the legacy A request!
An extended list is supported via the n request ONLY ( this was added to enable enhanced lists just like this pr proposes) so you need to add checks to ensure it never tries to send the longer list via A

ric355:
Agreed this breaks the A request and should not be integrated in its current form as it will break all thirds party applications that use the A request (mine being one of them!).

adamg-86:
I don't understand why it would affect the "A" request?? It still sends the 75 first values with the CAN_PACKET_SIZE = 75. The "n" request is the one linked to NEW_CAN_PACKET_SIZE that I changed to 118. So you can access the new data with "n" or "r"

Autohome2:
It looks like I already added the needed check between A and n request when I added n for future use :-)
It would be helpful if you were able to confirm that the A request tests correctly and remains unaffected to assist noisymime in the merging process.

adamg-86:
I tested the 'A' , 'n' and 'r' look to be working ok.

PR #589, opened 2021-06-05
Enhanced programmable outputs

VitorBoss:
Added possibility to set minimum programmable output time.

Added names for the comparators to make it more user friendly, TS will complain about the change for existing tune.

I got a lot of requests for this feature.

Keeps the RAM usage the same.


VitorBoss:
image
image

DeeEmm:
Do you also have 'NAND' and 'OR' logical operators? 

I would say that it is also worth having 'NOT' for dealing with digital (on-off / 0-1 / true-false) signals

/DM

VitorBoss:
> Do you also have 'NAND' and 'OR' logical operators? I would say that it is also worth having 'NOT' for dealing with digital (on-off / 0-1 / true-false) signals /DM


image

You can do `AND`, `XOR` and `OR` operations.
With those basics functions and this PR you make any condition, instead of using `NAND` you can invert the signal to compare.

All comparators signals and bit-wise fit in a byte, and with this PR no bit left for further improvement without changing the way those are defined and "consuming" more memory

DeeEmm:
Good stuff. 

I missed that. Didn't notice they were in the other field.

👍

noisymime:
Just reviewing this and everything looks pretty good. 2 questions:

1. Any idea what's going on with that `239U` sizing? Would be good if we can get to the bottom of it rather than including a seemingly random number check is all
2. What are the additional RAM and EEPROM requirements from this? (This is going to be a standard question for all impacting PRs going forward)

VitorBoss:
This don't took any additional RAM than the weird case of the 10 for for the 239 case, I did with 239 to be possible to increase the number of rules to 16. It can be any valid number, I just used it for simplicity sake.

VitorBoss:
Please see #575, I can rework this PR based on that in a few minutes.

shiznit304:
image
Programmable output 8 doesn't show the Limit time active feature.

Zeiberstein:
I did some exploratory testing on this exhanced programamble output feature.  Works nice. My Compliments.  :-) 

There is an almost cosmetic remark. 
Below 'Limit time active' you show a a text in blue: 'Set 0.0 to disable the limit'. 
This is only true in combination of 'Minimum=0.0'. If I choose 'Maximum=0.0' the output will never ever be activated regardless of fulfilment the conditons.
 I think it would be better to change the text in blue to 'Set minimum = 0.0 to disable the limit'.

PR #590, opened 2021-06-07
fix: install STM32RTC from Github

wilmardo:
This fixes the build on the latest `ststm32` platform. With release of the platform [2.0.0](https://github.com/stm32duino/Arduino_Core_STM32/releases/tag/2.0.0) it moved some RTC files from the core to the `STM32RTC` library. 

Platform side:
https://github.com/stm32duino/Arduino_Core_STM32/pull/1110

Library side:
https://github.com/stm32duino/STM32RTC/commit/bf74a2c0c62aab65380e699c4d51c6414f67345a

But the library is wrongly tagged in the PlatformIO library manager, the latest release is 1.1.0 of May 2020. Since that release is missing the moved core files it breaks the build.
I opened an issue to fix it but the maintainer isn't that keen to look into it:
https://github.com/stm32duino/STM32RTC/issues/49

I will followup on the issue on the PIO side but for now this will fix the build.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -59,7 +59,7 @@ platform = ststm32
 framework = arduino
 ;board = genericSTM32F407VET6
 board = black_f407ve
-lib_deps = stm32duino/STM32duino RTC
+lib_deps = https://github.com/stm32duino/STM32RTC#1.1.0
 board_build.core = stm32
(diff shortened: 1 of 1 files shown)


wilmardo:
Followup issue on the PIO side:
https://github.com/platformio/platformio-core/issues/3966

noisymime:
Nice work investigating this. I had gotten as far as discovering something was broken on the RTC library, but not dug any deeper.

PR #591, opened 2021-06-08
Add Doxygen markup for grouping decoders.

ohollmen:
Tidy-up doucumentation just enough to make Doxygen 'Modules' page neat / somewhat consistent (w/o stripping
any info, on the contrary some info is now added for few decoders that had only name present).
Add minimally modified Doxygen Doxyfile configuration.


noisymime:
Terrific work! Thank you for this PR, it's great to see this type of change coming in.

ohollmen:
Thanks Josh, it's great to hear that from you. I can continue (similarly, honoring existing documentation) adding doxygen markup in suitable batches as you seem okay it. I started with decoders because I've noticed those (and rpm and sync input in general) catalyze large portion of questions from Speeduino starters and having a bit of geeky doc detail available would possibly guide mid- to highly techical users nicely.

noisymime:
@ohollmen These types of PRs will always be VERY welcome on the project. Like so many codebases, there are good intentions here for documentation that haven't always been given the time they need, so if it's something you're interested in doing then that would be fantastic to see!

PR #595, opened 2021-06-19
Fix boost cut

adamg-86:
Reset boost cut bit if not enabled and fix map reading.
When in Map reading = event average and boost cut engaged (cut ignition) the map reading got stuck because there was no ignition event. With this pull request, it defaults to instantaneous reading if engine protection is engaged.


noisymime:
Nice find and fix!

PR #598, opened 2021-06-24
fix Programmable outputs indicator

adamg-86:
added the time delay to activate de programmable outputs indicators in TS


adamg-86:
PR #589 fixed this issue

PR #602, opened 2021-06-27
Enabled softRevLimit time limit

VitorBoss:
Currently this functions is inactive


pazi88:
I have noticed too that there isn't any code there to even use this setting.

VitorBoss:
A friend of mine told me it was't working, when I did check there was no code for it. Now it should work just fine

PR #603, opened 2021-06-27
Sincronize tacho events with coil events

VitorBoss:
Current mode is asynchronous making electronic reading of tacho be totally unusable.

This change make the tacho pulse at beginning of coil charge because pulsing at end is buggy.

The tacho time setting is a minimum duration pulse at low level, sometimes it is bigger because the nature os asynchronous operation.

Sincronize tacho events with coil events

### DIFF
--- a/speeduino/globals.h
+++ b/speeduino/globals.h
@@ -733,7 +733,7 @@ struct config2 {
   byte crankingPct;     ///< Cranking enrichment (See @ref config10, updates.ino)
   byte pinMapping;      ///< The board / ping mapping number / id to be used (See: @ref setPinMapping in init.ino)
   byte tachoPin : 6;    ///< Custom pin setting for tacho output (if != 0, override copied to pinTachOut, which defaults to board assigned tach pin)
-  byte tachoDiv : 2;    ///< Whether to change the tacho speed ("half speed tacho" ?)
+  byte tachoDiv : 2;    ///< Whether to change the tacho speed (1 = "half speed tacho")
   byte tachoDuration;   //The duration of the tacho pulse in mS
   byte maeThresh;       /**< The MAPdot threshold that must be exceeded before AE is engaged */
   byte taeThresh;       /**< The TPSdot threshold that must be exceeded before AE is engaged */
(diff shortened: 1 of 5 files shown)


noisymime:
Can you explain what the issues you're seeing with the tacho pulse being at the end of the dwell rather than the beginning? The timing a pulse takes place shouldn't matter to a tacho, only the frequency and duration.

VitorBoss:
Sorry for the delay, my internet stopped working for 4 hours.

On current way of pulsing tacho the event timings(period) are all over the place because tacho are driven by the 1ms interrupt, for a analog tacho is fine, but for a electronic reading is impossible to know what are the RPM, that happens because the tacho output will be driven with a delay, from 0 up to 1ms.

For low RPM is possible to guess, for high RPM the uncertainty is a lot.
At 6K RPM(10ms between events) the tacho output timing can be 10 or 12ms, that is a huge difference.

Tested on my bench in a 4 cylinder high dwell(8ms+) and high RPM(6K+) everything worked as intended.

muuris:
I've experienced the buggy behaviour also, but I doubt this will fix it. Some electronic tachos need a constantly 50% duty signal with variable frequency. The way tacho signal is currently made won't allow small code footprint addition for this.

VitorBoss:
> I've experienced the buggy behaviour also, but I doubt this will fix it. Some electronic tachos need a constantly 50% duty signal with variable frequency. The way tacho signal is currently made won't allow small code footprint addition for this.

Try to write a simple arduino frequency read program, the falling edge on the speeduino side is consistent, or the rising edge after the transistor/ULN.
For 50% duty a timer will be required

muuris:
> > I've experienced the buggy behaviour also, but I doubt this will fix it. Some electronic tachos need a constantly 50% duty signal with variable frequency. The way tacho signal is currently made won't allow small code footprint addition for this.
> 
> Try to write a simple arduino frequency read program, the falling edge on the speeduino side is consistent, or the rising edge after the transistor/ULN.
> For 50% duty a timer will be required

Exactly, it would require a timer and even if there was one available (others than Arduino-based), adding such tacho mode would mean rewriting the whole tahco code to avoid making it very messy. This is actually out of scope of this PR..

noisymime:
I agree that a different approach is needed here. The calling of an additional function from the interrupt here was already something I wasn't happy about and if it's not going to resolve the core issue we might need to think about an alternative. This might include needing a separate mode for tachos that require a 50/50 duty cycle.

VitorBoss:
This keep the timing for sub millisecond accurate. Tested here

VitorBoss:
Since #825 this is unnecessary, closing it.

PR #605, opened 2021-07-04
Disable EGO during DFCO

shiznit304:
This PR adds another condition for EGO activation. This prevents EGO corrections from accumulating during DFCO. It will simply set EGO correction to 100% if DFCO is active.


noisymime:
Nice addition. Thanks

PR #607, opened 2021-07-05
Open Loop Boost Control by MAP or TPS

shiznit304:
Allows user to switch between TPS load and MAP load for open loop "boost control". This doesn't take up any additional memory. The Y axis of the user's tune will need to be rebinned as the y axis variable has changed


noisymime:
Boost control using MAP as a load source can be very problematic and is why it has never been included before. When you change the boost control, you're altering the MAP, so if you're also using that as the load source you can end up with terrible feedback and oscillation. This is way it's typically controlled with something other than MAP

Do you have a specific use case in mind for this?

pazi88:
Yeah this doesn't make much sense to me either. Why have something you want to control as input.

shiznit304:
I added this feature to allow the boost control feature to control generic PWM things. I currently use it to control my purge valve (valve opens up during cruise)

PR #608, opened 2021-07-07
Increase TPS resolution to 0.5%

VitorBoss:
Current TPS acceleration enrichment can be problematic with very small throttle changes(0 to <2%).

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -438,7 +438,7 @@ page = 2
    ;  name       = scalar, type,    offset,        units,     scale, translate,    lo,      hi, digits
       veTable    = array,  U08,       0, [16x16],"%",          1.0,      0.0,   0.0,   255.0,      0
       rpmBins    = array,  U08,     256, [  16], "RPM",      100.0,      0.0,   100.0, 25500.0,      0
-      fuelLoadBins = array,  U08,   272, [  16], { bitStringValue(algorithmUnits ,  algorithm) },        2.0,      0.0,   0.0,   {fuelLoadMax},      0
+      fuelLoadBins = array,  U08,   272, [  16], { bitStringValue(algorithmUnits ,  algorithm) },       fuelLoadRes,      0.0,   0.0,   {fuelLoadMax},      fuelDecimalRes
       ;fuelLoadBins = array,  U08,   272, [  16], { bitStringValue(algorithmUnits ,  algorithm) },        2.0,      0.0,   0.0,   { arrayValue(rpmBins , algorithm) },      0
 
 ;--------------------------------------------------
@@ -447,7 +447,7 @@ page = 2
 page = 3
       advTable1   = array,  U08,     0,[16x16],  "deg",       1.0,       -40,   -40,   70.0,      0
       rpmBins2    = array,  U08,   256,[   16],  "RPM",       100.0,     0.0,   100,   25500,      0
-      mapBins1    = array,  U08,   272, [  16], { bitStringValue(algorithmUnits ,  ignAlgorithm) },        2.0,      0.0,   0.0,   {ignLoadMax},      0 ; This name has to be used in order for the table to show up in MLVs table. I can't find the regexs that make this work :(
+      mapBins1    = array,  U08,   272, [  16], { bitStringValue(algorithmUnits ,  ignAlgorithm) },     ignLoadRes,      0.0,   0.0,   {ignLoadMax},      ignDecimalRes ; This name has to be used in order for the table to show up in MLVs table. I can't find the regexs that make this work :(
 
 ;--------------------------------------------------
 ;Start Page 4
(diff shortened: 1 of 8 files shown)


shiznit304:
Wow neat stuff! Does this mean I can lower my TPSdot threshold to below 40 ?

VitorBoss:
I'm using 20 and considering use 10

HWright9:
Any reason that TPS resolution was not increased 4x instead of 2x? Why not use the full range of of the ADC? Also why not increase the resultion for the load tables? Fuel load is an int16_t.
Asking because I am adapting from a factory Alpha-N motorcycle ECN to Speeduino and their resolution is 0.1% TPS, the load tables appear to need this. 
Perhaps I will look at this as an extension to this PR.

muuris:
> Any reason that TPS resolution was not increased 4x instead of 2x? Why not use the full range of of the ADC? Also why not increase the resultion for the load tables? Fuel load is an int16_t.
> Asking because I am adapting from a factory Alpha-N motorcycle ECN to Speeduino and their resolution is 0.1% TPS, the load tables appear to need this.
> Perhaps I will look at this as an extension to this PR.

-ADC always has some noise, the full raw value isn't the same as usable resolution.
-Check storage.h for EEPROM layout, and you'nn notice there wouldn't be space for 2-byte value tables in Arduino Mega.

HWright9:
Ok, thanks for the background. While what you say is generally true I will say that my own application of a cruise control on a Arduino mega that reads the TPS in a vehicle does not have issues with the ADC accuracy, at least not 10x the available resolution. Better resolution on the input allows more options for filtering in the software.

I did have a look at the code and I notice that the axis points are embedded with every table. It would be a fairly major code change, but if the axes were seperated from the tables and made common where appropriate we could save a lot of memory even if we increased the resolution on those axes. For example I don't think each individual cylinder trim needs it's own load and rpm axis. Spark and fuel tables could be common etc.

It depends on the direction of development. If everything moves to teensy then a lot of these issues go away.

VitorBoss:
> Any reason that TPS resolution was not increased 4x instead of 2x? Why not use the full range of of the ADC? Also why not increase the resultion for the load tables? Fuel load is an int16_t.
> Asking because I am adapting from a factory Alpha-N motorcycle ECN to Speeduino and their resolution is 0.1% TPS, the load tables appear to need this.
> Perhaps I will look at this as an extension to this PR.

Yes, it is possible to do all the math on the fly, but the settings will remain on same `0.5%` resolution steps, need to be tested, I'm using this PR for a while and it is working fine, did't use all TPS related functions, just the very basic ones.

shiznit304:
@VitorBoss looks like the change isn't working correctly on programmable outputs. I have to double my TPS condition to make my programmable output work right.

VitorBoss:
That is expected, internally the firmware see the TPS as 0 to 200 because it don't use floating point. I'm testing a 10x MAP resolution increase and it have the same behavior on the programmable outputs

theKraid:
Supporting this PR! Setups with ITB + Alpha-N especially (I personally have three cars like this) benefit from the increased resolution

jonbills:
I support this being merged - it's just what I'm looking for.

noisymime:
This is looking good and I support the change. 

We do need the updates to existing tune tables to be included though, painful I know but it's something that will mess up a lot of tunes otherwise.

jonbills:
Whats involved in updating existing tunes? is it just opening each tune and making sure any tables with TPS bins have sane values?
I'm happy to spend some. time on that.

noisymime:
No, this is about having the routine in `updates.ino` perform the required changes to the impacted axis on first startup after someone updated to this firmware version.

VitorBoss:
@adbancroft Could you help me with table iterator changes? It need to be able to change the scale between 2 and 1 and I don't know how to do it, the scale could reflect the status of a bool making the MAP as default scale

adbancroft:
> @adbancroft Could you help me with table iterator changes? 

Of course!

adbancroft:
How will this work if the user changes load source? E.g. from TPS to MAP & back to TPS. Won't they have to resend/burn their tune?

VitorBoss:
As far as I can tell it might not require a rework on the axis, but every time you change something TS will suggest a burn, even if you roll back

noisymime:
I've just finished testing (Bench and on vehicle) for this and it looks really good. Tested on Alpha-N and MAP based setups and both updated from `202108` smoothly, no errors/warning and the axis did exactly what I'd expect them to.

Switching back and forth between load sources obviously scales them, but this is the expected behaviour and not something that is a common change to be making anyway. My biggest concern here was around ensuring there was a smooth upgrade path for existing users and this seems to be fine.

It's a big change though and coming in late in the cycle, so if you have any concerns please call them out, otherwise I'll merge this later today.

theKraid:
Is it normal that the 0.5% increments are not showing on the TPS gauge in Tunerstudio?

noisymime:
> Is it normal that the 0.5% increments are not showing on the TPS gauge in Tunerstudio?

TunerStudio doesn't automatically update the gauges unfortunately. If you right click on the gauge and then select the TPS reading from the sensors area again it will update to include the extra resolution

theKraid:
Perfect, thank you!

theKraid:
@VitorBoss Just wanted to say thank you so much for implementing this. It improved the AE in my ITB cars tremendously without having to change anything in the tune, especially when starting from low load.
Apart from this I'm sure an implementation of #697 would improve it even more especially when starting from high load.

VitorBoss:
@theKraid I had done something similar to that, but with linear scaling 2 years ago, I'll dig my PC to see if I still have the code, that is a must have to avoid over enrichment at some loads.

runesmdk:
> @theKraid I had done something similar to that, but with linear scaling 2 years ago, I'll dig my PC to see if I still have the code, that is a must have to avoid over enrichment at some loads.

See #1210 and #697 
Did you find that code?

PR #612, opened 2021-07-17
Enhanced cranking

VitorBoss:
Added possibility to scale up priming size, some pure ethanol(E100) vehicles struggle to start, doing prime pulse 2 or 3 times makes it start on first cranking, possible to add 2 to 500% of table content. An enable toggle was added.

Added possibility to add a second ignition pulse during cranking, some OEM ECU already does that for cold start, the spark duration must be grater than 0, a scale to the second pulse was added as the second discharge may or may not need the same dwell. An enable toggle was added.
imageimage

### DIFF
--- a/speeduino/scheduler.h
+++ b/speeduino/scheduler.h
@@ -175,6 +175,7 @@ struct Schedule {
   COMPARE_TYPE nextEndCompare;        ///< Planned end of next schedule (when current schedule is RUNNING)
   volatile bool hasNextSchedule = false; ///< Enable flag for planned next schedule (when current schedule is RUNNING)
   volatile bool endScheduleSetByDecoder = false;
+  volatile bool outputHadRepeated = false;
 };
 /** Fuel injection schedule.
 * Fuel schedules don't use the callback pointers, or the startTime/endScheduleSetByDecoder variables.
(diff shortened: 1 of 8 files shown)


rafolg:
Hello,

Im trying to push a commit to this PR but I am getting the below error message

image

rafolg (approved):

VitorBoss:
Removed RAM usage by 32 bytes

speeduino/scheduler.h
```diff
@@ -173,8 +177,9 @@ struct Schedule {
 
   COMPARE_TYPE nextStartCompare;      ///< Planned start of next schedule (when current schedule is RUNNING)
   COMPARE_TYPE nextEndCompare;        ///< Planned end of next schedule (when current schedule is RUNNING)
-  volatile bool hasNextSchedule = false; ///< Enable flag for planned next schedule (when current schedule is RUNNING)
-  volatile bool endScheduleSetByDecoder = false;
+
+  volatile byte scheduleFlags;        ///< All flags used in the scheduler
```
TBAMax, line 181:
```suggestion
 volatile  uint8_t hasNextSchedule :1;
 volatile  uint8_t endScheduleSetByDecoder :1;
 volatile  uint8_t scheduleRepeatIsActive :1;
 volatile  uint8_t ignitionRepeatCount :5;
```
This should save even more ram as repeat count can also be squeezed into the same byte with flags. Also flags retain their original usage style in the code, so sadly, yes then you would have to revert all that too.
(Note 'ignRprCnt' renamed to 'ignitionRepeatCount' this is out of pure respect for others reading the code, not having to figure what 'ignceint reptile canteen' or what it is.)

adbancroft:

PR #617, opened 2021-07-22
AE graph unit for O2 sensor

VitorBoss:
I know the add new features windows has closed, but hope that simple 3 lines code get pulled in, this add the selected O2 unit to the acceleration enrichment graph, currently shows AFR even with lambda scale selected.


noisymime:
Easy done 👍

PR #623, opened 2021-07-29
Tacho trigger wheel output

HWright9:
This creates the option to output the tachometer signal as a number of "pulses per rev" of the trigger wheel using the trigger interrupts. This creates a 50% duty cycle square wave that is directly linked to RPM with minimal overhead.

Depending on the decoder this can be adjusted as long as the number of pulses are evenly divided amongst the available real teeth. Software protection is provided to make sure the number of pulses matches the trigger wheel and missing teeth capabiltiy. And the old method of using ignition triggers is still available.

I provided support for common decoders such as toothed wheel, dual wheels, basic distributor, GM7x and GM24x. It may be possible to generate outputs off the other wheels but right now I stuck to generating 50% duty cycle square waves.
Tested on the bench and in my vehicle with a speedhut.com tachometer.

Screenshot Tacho wheel output

TunerStudio Tacho Settings


noisymime:
If I'm reading this right this will change the existing behaviour for the tacho out, which isn't ideal. The addition looks good, but the wheel based tacho option should be the 3rd one for `tachoMode` rather than being the first. Replacing the first option would mean existing tunes would change to use this new mode without knowing.

Is that right?

muuris:
This is a welcome addition, as tacho should be tied to decoder wheel rather than ign pulses (which may be cut on rev limiter etc). Also some tachometers require the 50% duty pulsing and won't display correct rpm on the whole range with current constant-length pulse.

HWright9:
> If I'm reading this right this will change the existing behaviour for the tacho out, which isn't ideal. The addition looks good, but the wheel based tacho option should be the 3rd one for `tachoMode` rather than being the first. Replacing the first option would mean existing tunes would change to use this new mode without knowing.
> 
> Is that right?

Agreed. I can swap it around.

HWright9:

noisymime:
Just to confirm as I'll need to update the documentation around this, which decoders does this work with? Looks like the missing tooth and dual wheel ones for full functionality? The 24X and 7X ones look to have some support, but not for selected the number of output pulses?

HWright9:
This should be what you need.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/wrighe/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/wrighe/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">



Tacho Output   Supported Triggers. |   |  
-- | -- | --
Function | Supports User   Defineable Pulses Per Rev     (tachoPulsesPerRev) | Comments
triggerPri_missingTooth() | Yes | The   number of tacho pulses per revolution of the primary wheel.  The maximum number of  pulses is actual teeth / 2 and is limited   to generating a pulse every missing teeth * 2 by the software.
triggerSetup_DualWheel() | Yes | The   number of tacho pulses per revolution of the primary wheel.  The maximum number of  pulses is actual teeth / 2 and is limited   to generating a pulse every missing teeth * 2 by the software.
triggerSetup_BasicDistributor() | Yes | The   number of tacho pulses per revolution of the distributor (CAM Speed). The   maximum number of  pulses is actual   teeth / 2
triggerPri_GM7X() | No | The number of pulses   is equal to actual teeth /2 ignoring the odd spaced tooth 3.
triggerSec_24X() | No | The number of pulses   is 1 pulses per CAM rev based on secondary trigger
triggerSec_Jeep2000() | No | The number of pulses   is 1 pulses per CAM rev based on secondary trigger



</body>

</html>

adbancroft:

PR #624, opened 2021-07-29
Non linear TPS

HWright9:
My application is a 2004 Moto Guzzi 1100cc V-Twin. The factory bike uses a non-linear TPS as its primary load input. This TPS has 2 slopes to provide more resolution at lower throttle positions which are critical in ITB applications.

I added a simple 3 point table to map the non-linear voltage to a linear 0-100% TPS value so I could maintain the benifts of the enhanced resolution and more easily translate the factory tune. Also things like TPS-dot will work correctly.

I realise this sensor type is rare but I made this change for myself. Others adapting some motorbikes may use the same style of TPS. The old linear method is retained for backwards compatabilty but is essentialy redundant since using 2 points in this table achieves the same outcome. I only needed 3 points to map the curve, could easily be expanded if others need that functionality.

TPS Non linear

TPS Non linear_graph

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -198,7 +198,7 @@
 
     endianness          = little
     nPages              = 14
-    pageSize            = 128,   288,     288,    128,     288,    128,    240,     384,    192,    192,    288,    192,    128,    288
+    pageSize            = 128,   288,     288,    128,     288,    134,    240,     384,    192,    192,    288,    192,    128,    288
 
     ; New commands
     pageIdentifier      = "\$tsCanId\x01", "\$tsCanId\x02", "\$tsCanId\x03", "\$tsCanId\x04", "\$tsCanId\x05", "\$tsCanId\x06", "\$tsCanId\x07", "\$tsCanId\x08", "\$tsCanId\x09", "\$tsCanId\x0A", "\$tsCanId\x0B", "\$tsCanId\x0C", "\$tsCanId\x0D", "\$tsCanId\x0E"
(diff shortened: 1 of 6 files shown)


jfigueredo:
Hello. Currently trying to wire a Bosch 0 280 122 201 4 pin TPS to a speeding. 
So for my understanding I need to wire the 2 pins in the same TPS input on the speeding right or am I wrong?

HWright9:
Hey that type of sensor has two independent TPS sensors. One reading from 0-23deg rotation and the other from 20-90deg. So you need two independent inputs to the speeduino to read it. You can't wire them to the same pin. 
It would be possible to write some code to do this and relatively simple to switch between analogue inputs at some % value. But even this PR for non-linear TPS is not pulled in yet. I don't know what the appetite is from the general community to support "wierd" TPS, as generally found on Motorbikes using TPS as the primary load input.  If anyone wants to chime in with support I can put the effort in to write a more generic function to deal with this sort of stuff but only if there is support for pulling it in.

jfigueredo:
> Hey that type of sensor has two independent TPS sensors. One reading from 0-23deg rotation and the other from 20-90deg. So you need two independent inputs to the speeduino to read it. You can't wire them to the same pin.
> It would be possible to write some code to do this and relatively simple to switch between analogue inputs at some % value. But even this PR for non-linear TPS is not pulled in yet. I don't know what the appetite is from the general community to support "wierd" TPS, as generally found on Motorbikes using TPS as the primary load input. If anyone wants to chime in with support I can put the effort in to write a more generic function to deal with this sort of stuff but only if there is support for pulling it in.

If you can get it to work I can pay you. I unscrewed the tps and it has a I shaped connector to the throttle body, and I think its difficult to manage to get a replacement 3 pin one with this kind of connector.

jfigueredo:
Can you give me your Facebook contact or any other way to contact you?

HWright9:
> Can you give me your Facebook contact or any other way to contact you?

Hey, just join the slack group and you can message me on there. Dont worry about payment. I just do this for fun. I can write some test code over the next few days. But can you test it for me?

HWright9:
Ok version 2 of non-linear TPS.
Strategy now is just to add a couple more variables to save memory. Also added calibration option for dual sensor TPS. (Bosch 4 wire). A lot of these non linear or dual rate TPS are used on motorbikes using TPS only systems. 

Example of how ADC is mapped to TPS shown below

Screenshot TPS Parameters

Example of the TunerStudio setup.

Screenshot TPS setup New

jfigueredo:
Really, really thanks for this! Tested it with the developer(dual sensor tps) and it works like a charm!
For people like us that are using bike engines, or bike parts we really need this because they are mainly tuned by tps, and what better tps than the factory one right??
This way we don't need to buy another tps and adapt it since this fork covers most of the tps on the market!

digmorepaka:
i think it would be useful to merge this with https://github.com/noisymime/speeduino/pull/608 as it's somewhat related

HWright9:
Confirming that this update is compatible with the 0.5% TPS resolution update. Adding this PR along with #608 will complete the updates to enhance support for motorcycle TPS based applications.

HWright9:

PR #625, opened 2021-07-29
DRZ400 Trigger Decoder

mike501:
Implement DRZ400 trigger decoder. Based on duel wheel with modification to take into account how the signals are generated on this bike engine.


Zeiberstein (commented):
Lines 2525, 2527, 2528 seem to miss a comma in between the field definitions. Maybe TS does the right thing with it, but this seems a bit unfinished.

mike501:
> Lines 2525, 2527, 2528 seem to miss a comma in between the field definitions. Maybe TS does the right thing with it, but this seems a bit unfinished.

Good spot. Not sure why that wasn't highlighted when I merged the latest code release into it. Makes me worried I've done something wrong in git.

mike501:
missing 

> Lines 2525, 2527, 2528 seem to miss a comma in between the field definitions. Maybe TS does the right thing with it, but this seems a bit unfinished.

Commas added & updated to the latest firmware release

PR #627, opened 2021-08-02
Performance and economic fields of tunerstudio

nuosu:
Added custom channels to use the performance and economic fields of tunerstudio, becoming accurate and usable


noisymime:
Hi. Not quite sure what you're trying to achieve with this change? There is only a single line changed in the ini file and that doesn't look like it enables any of the mentioned fields.

nuosu:
i have tested  that function in tunerstudio .  but get wrong data. The
vehicle speed is too much higer than speeduino .after that. I tried to add
a custom output channel in accordance with the requirements of tunerstudio
. The problem is solved

Josh Stewart ***@***.***> 于 2021年8月2日周一 下午3:11写道：

noisymime:
Closing as I can't see what the actual changes are here. Please open a new PR if needed

pazi88:
I have the economy fields currently baked into speeduino.ini. I will make PR once testing is complete.

PR #633, opened 2021-08-15
change sequence of injection in Semi Seq. 4 cyl

adamg-86:
I know it is a sensitive subject but as it is now the sequence is not correct. The current sequence is correct if you wire inj1 to output 1, inj2 to output 2, inj3 to output 3, and inj4 to output 4. If you want to switch to full sequential later this order is not correct. This pull request fixes this issue and it really helps idle if you are wired for full sequential. I tested it for 2 weeks now and when the temp sensor is heat-soaked on hot start, the car idle way better at high ARFs. 
image


jackscr12:
Injector drivers fire in numerical order so you have to wire 1-1, 2-3, 3-4, 4-2  to have proper injector sequence . SO how can 1-1 2-2 3-3 4-4 be correct for a traditional 1-3-4-2 firing engine?

noisymime:
Closing as an injector pairing option was added in beedfb44e41052361083f00778f8d7b450d38895

PR #634, opened 2021-08-15
Launch control Vss threshold activation

adamg-86:
Set a maximum speed which the Launch control can be activated. Prevent activating on rev matching downshift.


shiznit304:
I'm not sure if this feature is really needed. You have to set your launch/flat shift switch to a low number. You shouldn't be hitting your launch rpm switch when downshifting. For example I have my Launch/Flat shift switch set at 1300 (Clutch needs to go in below 1300 to activate launch control, otherwise it does flatshift limiters if configured)

pazi88:
This is actually quite usefull feature. The launch rpm I for example use is 4500rpm and it's quite easy to accidentally activate that while driving. Also it makes drag racing easier when you can immediately disable launch when wheels start turning, but the clutch pedal may not be fully up yet to disable launch.

adamg-86:
the goal is to not be able to activate ANY cut between max vehicle speed threshold and launch/flat shift switch RPM. You set the max speed to ex: 10 kph and switch RPM to 5200RPM, so you flat shift over 5200 and can rev-match under 5200 with no possible cut. At the track I was annoyed with that, if I rev-match with a lot of throttle (over the threshold) it would cut.

noisymime:
So, nearly 3 years to the day since you submitted this one, but it does look like it has value. 

Thanks for submitting!

pazi88:
Nice

adamg-86:
I just realized this got merged. Nice

PR #635, opened 2021-08-17
Read analog values via CAN BUS

celturbo:
Enhancement for reading 16 analog signals via native CAN-BUS. with selection of Edianness.
can read


pazi88:
This is feature that should be definitely included in the FW. Standard feature to any aftermarket ecu. But does this really work? Looking at the code, I can't find you actually reading CAN bus when this is enabled? Just read what ever has left as inMsg from previous read. Also what if the OBD2 feature is in use? How do you define if the CAN messages is request for that or Analog CAN values?

celturbo:
> Yes. But it only reads current CAN message in currently present the `inMsg`. But after that there is the `while (Can0.read(inMsg)` Which reads through all the messages in the CAN interface receive buffer. And uses the messages that are related to OBD2 and discards others. So you could have for example 10 CAN messages in the receive buffer. First one is for analog CAN messages. You read that and then code moves to `while (Can0.read(inMsg)` There is still 10 CAN messages in the buffer and maybe one of those could be for ODB2. So the `can_Command()` -function reads the OBD2 related and discards the others that might include for example analog CAN data. So as end result you might have missed 8 analog CAN messages with this code. So to make this work, you should move the code under `readAuxCanBus()` to `can_Command()` so that you properly parse all the CAN messages based on message id.
> 
> @Autohome2 Can correct me if I'm wrong.

ok, you're right, i made the modification, i hope that's it, i tested the reading again and it's ok.

sbtoonz:
Works grand thanks for this :D

sbtoonz (approved):
tested against 4518bb6c40cca95f7130dc6d731aec309bb407fc

PR #637, opened 2021-08-18
PWM Fan addition

pazi88:
This PR adds PWM fan control for Teensy and STM32. The Arduino MEGA lacks timers to do this, so if user tries to select PWM fan mode with that MCU, Tuner Studio gives warning that it isn't available and PWM Fan related options are grayed out. Tested to work on STM32, but not yet on Teensy. But planning to do it soon. Teensy 4.1 needs some attention that I'm not sure how to do.

NOTE! This PR might seem huge because the many files changed. But it's mainly because the added timer. The PR itself tried to be as simple as possible.


noisymime:
Looks good. My only suggestion here would be that if this option is chosen on a mega2560 that the fan operates in On/Off mode. 

Eg if the PWM value is anyone above 0, the fan is turned on. 

I think this is a safer option than doing nothing is all. If someone selects the wrong option and ignores the message, then at least their fan will still run rather than potentially cooking their engine.

pazi88:
> Looks good. My only suggestion here would be that if this option is chosen on a mega2560 that the fan operates in On/Off mode.
> 
> Eg if the PWM value is anything above 0, the fan is turned on.
> 
> I think this is a safer option than doing nothing is all. If someone selects the wrong option and ignores the message, then at least their fan will still run rather than potentially cooking their engine.

Yeah I can do that.

pazi88:
@noisymime Ok, the Fan output is now active on mega if the duty is bigger than zero. I also made a change where the fan duty works on mega too and the duty reading is available on secondary serial. In this way external PWM controller can be used and the duty/temp curve is still configurable on TS. This was request regarding this PR I got while ago, and in a matter of fact I can also use this feature in that way in the setups I don't have STM32.

noisymime:
Great work @pazi88

PR #639, opened 2021-08-19
[WIP] CAN broadcasting menu

pazi88:
This PR adds CAN broadcasting menu. Currently this is only for broadcasting CAN messages for couple of different OEM instrument clusters, but support for more OEM instrument clusters, aftermarket dashes, gauges etc. can be added rather easily. The BMW one is currently bench tested to be working. Also fully configurable CAN broadcast outputs could be nice thing to have. The amount of different things to broadcast can quickly escalate, so these are now under new files and not in already rather big cancomms file.


speeduino/canBroadcast.ino
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// switch case for gathering all data to message based on CAN Id.
+void DashMessage(uint16_t DashMessageID)
+{
+  switch (DashMessageID)
+  {
+    case CAN_BMW_DME1:
+      uint16_t temp_RPM;
+      temp_RPM =  currentStatus.RPM * 6.4; //RPM conversion
```
VitorBoss, line 38: Use of floats must be avoided at the moment, it is on lines 38 and 100.

pazi88: > Use of floats must be avoided at the moment, it is on lines 38 and 100.

Fixed

PR #641, opened 2021-08-21
[WIP] Coolant Based VVT Control

shiznit304:
VVT will only run when coolant is greater than set value. The VVT minimum coolant condition shares the same value with Accel enrichment cold taper max value. Didn't want to take up any more free memory. I made this PR to help @ThresherBuilt VVT Miata.

Later version could include tapering the PID values between two coolant temps.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -2860,6 +2860,7 @@ menuDialog = main
 
     dialog = vvtSettings, "VVT Control"
         field = "VVT Control Enabled",    vvtEnabled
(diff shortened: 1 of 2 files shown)


pazi88:
Not sure what has been done, but seems awful a lot changes, for such minor modification. Also the file comparison shows few places where vvt2 has been replaced with vvt1 and so on.

shiznit304:
@pazi88 The comparison check makes it look like i added more than I did. The majority of the changes is because of an if statement encapsulating the entire VVT code

PR #648, opened 2021-08-27
[Request for comment] Abstracted timers from schedulers

rob-deutsch:
This PR abstracts the 'timer functionality' out of the igntion scheduling functions, and creates a single common ignition scheduling function (`setIgnitionSchedule`) which is used for all ignitions.

This allows more flexibility for which/how timers are used, and will allow quicker/easier improvements to the scheduler. Not to mention, this is needed for the ESP32 port.

The same sort of abstraction could be added to the fuel scheduling functions.

### DIFF
--- a/speeduino/decoders.ino
+++ b/speeduino/decoders.ino
@@ -305,50 +305,50 @@ static inline void checkPerToothTiming(int16_t crankAngle, uint16_t currentTooth
   {
     if ( (currentTooth == ignition1EndTooth) )
     {
-      if( (ignitionSchedule1.Status == RUNNING) ) { IGN1_COMPARE = IGN1_COUNTER + uS_TO_TIMER_COMPARE( fastDegreesToUS( ignitionLimits( (ignition1EndAngle - crankAngle) ) ) ); }
-      else if(currentStatus.startRevolutions > MIN_CYCLES_FOR_ENDCOMPARE) { ignitionSchedule1.endCompare = IGN1_COUNTER + uS_TO_TIMER_COMPARE( fastDegreesToUS( ignitionLimits( (ignition1EndAngle - crankAngle) ) ) ); ignitionSchedule1.endScheduleSetByDecoder = true; }
+      if( (ignitionSchedule[0].Status == RUNNING) ) { IGN1_COMPARE = IGN1_COUNTER + uS_TO_TIMER_COMPARE( fastDegreesToUS( ignitionLimits( (ignition1EndAngle - crankAngle) ) ) ); }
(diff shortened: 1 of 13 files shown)


adbancroft:
I'm all for these abstractions.

It's an invariant that the callbacks are always in pairs, correct? And mostly processed by the code as pairs. So we could encode that invariant into the source code, so it can't be easily violated:

```
typedef extern void (*ignStartFunc)();
typedef extern void (*ignEndFunc)();

typedef struct IgnitionCallback {
  ignStartFunc startFunc;
  ignEndFunc endFunc;
} IgnitionCallback;

extern IgnitionCallback ignitionCallbacks[8];
```

It's the same memory footprint, just as fast (if not faster) and improves readability

adbancroft:
Should the callback array be of size IGN_CHANNELS?

noisymime:
I'm OK with the abstraction here, in particular wrapping the schedulers up in arrays and reducing set/interrupt functions down to singles. At a read over though, this **will** introduce some slower areas and the impact of these changes will need to be tested. 

In particular increasing the number of function calls (even inline ones) from interrupts is something that should only be done if it can't be avoided. Likewise anything that introduces more overhead in interrupts (Such as what `setIgnitionCompare()` and `getIgnitionCompare()` are doing) isn't ideal. The benefits may outweigh the drawbacks, but it needs to be measured. 

Some of this might be able to be avoided by using macros rather than functions, which avoids another context switch, but it may be difficult in other areas.

rob-deutsch:
Thank you @noisymime, I appreciate the comments. I have some ideas and some thoughts to mull over. I'll submit improved updates.

@adbancroft, you are correct, and there are many areas of the codebase that can be improved (both readability and performance) because of this abstraction. I would like to keep this PR limited to the 'mininum' needed to implement to the abstraction, though, so its easier to understand and reason about. Future improvements can then be tackled in seperate PRs.

To respond to each of your particular suggestions:

1. Combining the two callbacks into a single struct seems neat. That hadn't occured to me. I also note that #578 suggested adding references to the callback functions in `struct Schedule`.
2. Correct, the callback array should be size of IGN_CHANNELS. But, I would not suggest changing that in this PR, since we'd end up 'going down a rabbit hole' and also refactoring part of `initialiseAll()`.

DeionSi:
Why not include the register addresses in the Schedule struct? That way the getIgnitionCounter and setIgnitionCompare functions can be replaced by ignitionSchedule[x].ignitionCounter and ignitionSchedule[x].ignitionCompare. I'm not good with pointers but I think it should work and be pretty fast?

DeionSi:
So I replaced getIgnitionCounter and setIgnitionCompare with arrays and macros. This gives no change in performance.

adbancroft:

PR #649, opened 2021-08-27
VVT Min Coolant Condition & Timed Delay

shiznit304:
This PR adds a condition for minimum coolant temp before activating VVT. I also added a configurable delay from 0 to 1275 seconds for the purpose of oil warmup. 

If car is started below min coolant temp, it will wait until engine reaches min coolant temp, then starts counting the delay, before then activating VVT.
If the car is restarted while hot, it will skip the timed delay.
If car (somehow) goes below min coolant temp, VVT will deactivate. Once it goes back over, VVT will reactivate.


ThresherBuilt:
This works perfectly. I have mine set to delay to 180°F and then add 350 seconds to that. So far, in slightly varied conditions, it has worked great. Oil is at operating temp, so when VVT engages, it acts normal and hits all targets

PR #650, opened 2021-08-28
Closed Loop O2 Improvements

HWright9:
Single Sensor O2 control closed loop. Control improvements for single sensor control. Future enhancement to dual sensor groundwork is laid also.

Main changes:

O2 algorithm updates are linked to engine 720deg Cycles (revolutions / 2). There is a low load and high load value. Set this to ensure exhaust gas from the closed loop correction reaches the sensor. This automatically speeds up the calculation at high load and with engine speed.
There is also a sensor delay in secconds. The sensor delay is the response time of the sensor. The longer of the two delays will be used to trigger an algo update.

Conditons to run closed loop are divided into those that inhibt the algo running and set the correction back to 100% and those that freeze the update and hold the last value. I might put a few more handles to choose what each of the conditions does later but for now try it out.

A key condition is the AFR Target Min setting. Using the AFR target table and this value you can window where closed loop will operate and at what air/fuel ratios.

The simple algorithm is removed as it's functionality can be replicated by the PI controller. (just calibrate P gain to 0, Then I gain to -1 and 1 either side of 0 error)

The PI controller has direct values in it, not scalar constants as textbook PID. So if propotional gain is 5, this is 5% fuel correction to be applied. Same with integral control, it's X% per loop update of the integrator. 

There is an option to delay the integrator after proportional control. This is intended for narrowband switching. The code is looking for an intentional overcorrection on the proportional control to create a rich to lean or lean to rich transition. If that does not occur for X control loops, only then does it start integral control for mean correction. Best to start with it at 0 and adjust once other parameters are ready.

Example of my settings shown below:
O2 Calibration Example


VitorBoss:
The idea is awesome, just some small caveats
You can't grow a page size, they are stored in sequence, doing so will give a lot of unexpected behaviors, I suggest you to move all stored variables to page 10, there is a lot of free space there. Leave unused variables as `unused_xx` which xx is the byte number in the sequence (available on the ini) 

Please don't change the sequence of declared variables inside a page, that will cause errors.

HWright9:
Hey victor. Whats the actual issue with changing the eeprom, and page sizes? Because so far i have seen that tuner studio just writes the values back by name in the correct order as long as you load the .ini before connection.
I havent seen any strange issues.

Its just going to be a lot neater to keep all the ego stuff together.

pazi88:
You will be writing over the following page on the eeprom, if you incresse the page size in the middle.

muuris:
Check updates.ino, storage.h and storage.ino, you haven't modified those. Eeprom space is limited and can't overlap as pazi88 just mentioned.

When changing data order inside structs in version update, Speeduino does data reordering to keep old tunes compatible, as otherwise the old tunes would be corrupted useless. New features are safe to set disabled so users who won't go through every dialog won't experience unexpected behavior.

HWright9:
Hmmm. So just learned a lot! I had assumed the whole thing was just a copy of the structure defined in the RAM. Currently it's overwriting part of the boost table (which I wasn't using).

Any suggestions? I can jam the single cal values in somewhere but I need 15 bytes for the axis and 2 tables. Page 10 looks like it only has 2 bytes free.

VitorBoss:
Page 9 have 29 free bytes

HWright9:
Ok, fixed the memory pages. and added a couple of switches and conditions. Also got rid of a few global variables. This should be pretty much good to go now.
Latest settings I am running shown below.
image

HWright9:
Hi guys, so the last update is working well except for the time delay functions. Relating to **egoSensorDelay** and **egoFreezeDelay**. When I try and calibrate these to anything other than zero I don't get the delay I am expecting. Sometimes very short (< 0.2sec) and sometimes very long (up to 12.5sec). I think I am using runSecsX10 correctly to time against these parameters but haven't been able to resolve it so far. I plan to keep investigating myself but if anyone can have a look at the code and has some ideas then let me know. Thanks.

HWright9:
Ok timers fixed, was a memory order difference between globals.h and the speeduino.ini. Also fixed DFCO comparison to prevent overflow from disabling the delay at 109 mins into a drive. As far as I know the code is working now for a single O2 sensor.

HWright9:
Background on why alternating rich and lean pulses are required for 3 way catalysts:
(for inclusion in the Wiki)

3 Way converters
Allows reduction of NOx (nitrogen oxides) to N2 (nitrogen) and O2 (oxygen)
Allows oxidation of CO (carbon monoxide) to less-harmful CO2 (carbon dioxide)
Allows oxidation of HC (unburned hydrocarbons) to CO2 (carbon dioxide) and H2O (water)

Uses advanced catalyst chemistry to store and release O2, in conjunction with an O2 monitoring and control system.  
This system utilizes one or more O2 sensors to oscillate the fuel mixture between lean and rich conditions. This oscillation, combined with the O2 storage and release on the catalyst surface, allows for optimum reduction of all three emissions. 
In practice roughly a 3% swing rich and lean is required.

reference/speeduino.ini
```diff
@@ -420,7 +420,12 @@ page = 1
       rtc_trim        = scalar,  S08,    123,               "ppm",      1, 0, -127, +127, 0
       idleAdvVss      = scalar, U08,      124,        "km/h",       1,       0.0,   0.0,     255,      0 
       mapSwitchPoint  = scalar, U08,      125,        "RPM",      100,       0.0,   0.0,     16320,    0
-      unused2_95      = array,  U08,      126, [2],   "%",        1.0,       0.0,   0.0,     255,      0
+      
+      unused2a_126    = scalar,  U08,     126,         "",        1.0,       0.0,   0.0,     255,      0
+      unused2b_127     = bits,    U08,    127, [0:0],
+      egoResetwAFR    = bits,    U08      127  [1:1], "Freeze", "Reset"
+      egoResetwfuelLoad = bits,    U08    127  [2:2], "Freeze", "Reset"
```
karniv00l, line 427: missing commas:

```ini
egoResetwAFR      = bits,    U08,    127,  [1:1], "Freeze", "Reset"
egoResetwfuelLoad = bits,    U08,    127,  [2:2], "Freeze", "Reset"
```

reference/speeduino.ini
```diff
@@ -2396,23 +2423,38 @@ menuDialog = main
...
+      field = "AFR Target Condition Config",  egoResetwAFR,           {(includeAFR == 0) && egoType && (egoAlgorithm < 3) }
+      field = "Load For High Update Rate ",   egoFuelLoadMax,     { (includeAFR == 0) && egoType && (egoAlgorithm < 3) }
+      field = "Load Condition Config",        egoResetwfuelLoad,           {(includeAFR == 0) && egoType && (egoAlgorithm < 3) }
+      field = "Max Change in Fuel Load ",     egoFuelLoadChngMax,    { (includeAFR == 0) &&egoType && (egoAlgorithm < 3) }
+      field = "Re-enable delay after Freeze", egoFreezeDelay,      { (includeAFR == 0) &&egoType && (egoAlgorithm < 3)
+      field = "Integrator Delay",           egoIntDelay,         {(includeAFR == 0) && egoType && (egoAlgorithm < 3) }      }
+      
+    dialog = egoPID, "ego Control Closed Loop", xAxis  
+      panel = ego_Prop_Table       {(includeAFR == 0) && egoType && (egoAlgorithm < 3) }
+      panel = ego_Integral_Table   {(includeAFR == 0) && egoType && (egoAlgorithm < 3) }
```
karniv00l, line 2452: missing commas:

```ini
panel = ego_Prop_Table,       {(includeAFR == 0) && egoType && (egoAlgorithm < 3) }
panel = ego_Integral_Table,   {(includeAFR == 0) && egoType && (egoAlgorithm < 3) }
```

HWright9: Thanks.

HWright9:
This PR supersceeded by Dual O2 closed loop #732

PR #653, opened 2021-09-03
Consolidation of fuel schedules

TBAMax:
Implement the common function for setting fuel schedules.
Implement common function for fuel schedule interrupts.


TBAMax:
Replaced by PR #804

PR #660, opened 2021-09-15
Add secondary serial selection to Tuner Studio

Tjeerdie:
This PR adds the function where the user can select the serial port used for secondary serial. For the arduino MEGA it is still hard coded to serial3 because it will take to much ram to have serial1 and serial2 on standby for Tuner studio selection.


Tjeerdie:
Not enough interest and again broken because if all kind of changes and merges.

PR #661, opened 2021-09-17
chore(stm32): set the RTC lib back to the PlatformIO library manager package

wilmardo:
Followup of #590 

My issue over at PlatformIO just got closed that is should have been fixed automatically and that indeed seems the case:
https://github.com/platformio/platformio-core/issues/3966#event-5320698911

The library now shows the `1.1.0` tag at the same date as they Git tag:

image

https://platformio.org/lib/show/5502/STM32duino%20RTC

image

https://github.com/stm32duino/STM32RTC/releases

There is nothing wrong with the current approach but it can be reverted back if needed :)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -62,7 +62,7 @@ platform = ststm32
 framework = arduino
 ;board = genericSTM32F407VET6
 board = black_f407ve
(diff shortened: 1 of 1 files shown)


noisymime:
I much prefer to have these on the releases rather than github master, so good to see this come though. 👍

PR #662, opened 2021-09-18
Reduce 3d table memory usage (>500 bytes)

adbancroft:
Changes that reduce RAM usage. Specifically:

1. .bss section: save **518** bytes
2. .data section: save **196** bytes

Total: **714** bytes (reported by compiler & map/elf file analyses). A stub test program calling freeRam() showed savings of **721** bytes.

Savings mainly come from pushing table value & axis array declarations directly into the 3d table struct(s). Since there are multiple tables sizes, this requires multiple separate 3d table types (I.e. structs). Alternative is C++ templates - but that is not an option. 

There are a number of supporting macros that help manage the added complexity of multiple table types.

The memory savings also have a positive effect on performance:

1. TunerStudio I/O speed up: 20%
2. Table interpolation speed up: 10% to 30% (depends on which axis bin & where in the bin)
   2.1 I suspect the speedup is due to not dereferencing the value & axis array multiple times
   2.2 the interpolation algorithm has not been touched - just enough changes to work with the rest of the PR


pazi88:
Wow sounds great. I have zero idea what is happening on this PR. But I have been wanting to add initial duty map for closed boost control (like other modern aftermarket ecus have), but haven't done it because there is so little free ram left on mega. This would allow it and many other features to be added.

adbancroft:
>  I have zero idea what is happening on this PR

@pazi88 Is that a maintainability concern?

pazi88:
> > I have zero idea what is happening on this PR
> 
> @pazi88 Is that a maintainability concern?

No. Just my lack of coding skills 😅

VitorBoss:
Interesting, as far as I understood you change the table to be a contiguous data on memory, clever.
Looks really awesome, did you measure the loop speed impact?

noisymime:
Where's the RAM free up actually coming from here? I can understand the movement of the data into the struct, but that doesn't inherently free anything up as far as I can see. Is the reported difference a consequences of `freeRAM()` not taking into account fragmented RAM?

At a quick play, 2 things I noticed:

1. The tune needs to be reloaded and I imagine an update routine to fix that would be fairly painful. Not a major thing, but worth considering
2. For some reason I'm seeing one of the interval unit tests failing with this. Not obvious why, but I'll have a look into it.

Otherwise this looks really promising. The code is definitely a little messy having to work around the fixed table sizes, but it might be worth it.

adbancroft:
> 
> 
> Interesting, as far as I understood you change the table to be a contiguous data on memory, clever.

Yup. It ends up as a series of structs like this:
```
struct table3d8RpmTps {
  static constexpr const table_type_t type_key = table_type_key(8, axis_domain_Rpm, axis_domain_Tps);
  table3DGetValueCache get_value_cache;
  table3d_value_t values[8 * 8];
  table3d_axis_t axisX[8];
  table3d_axis_t axisY[8];
};

```

> Looks really awesome, did you measure the loop speed impact?

Only indirectly by testing the performance of various functions that might be impacted. ```get3DTableValue``` in particular seemed a priority - as noted, it has sped up under all conditions I measured.

adbancroft:
> Where's the RAM free up actually coming from here? I can understand the movement of the data into the struct, but that doesn't inherently free anything up as far as I can see.

That is a great question. I can see 2 types of savings per table:

1. Elimination of 9 bytes of overhead: axis sizes, pointers to the arrays etc. 9 bytes isn't a lot, but it adds up
2. The old tables stored values as an array of pointers to rows PLUS each row array. By allocating the values as a single block, that array of pointers isn't necessary. This adds up quickly.

I know........we could apply all of (2) & some of (1) to the existing design. But having a table be self contained in a simple struct is a much more maintainable design.

adbancroft:
> Where's the RAM free up actually coming from here?

I also see that the `_3DTable_heap` seems to be overallocated somehow. 

- We set the array size to 2765. 
- PIO Inspect reports the size as 2937. 
- A map file + elf file analyser reports 2825 bytes.

Not sure which one to believe.....

adbancroft:
> Is the reported difference a consequences of `freeRAM()` not taking into account fragmented RAM?

You mean if the heap is more fragmented in master versus this PR? Very unlikely: the PR doesn't add or remove any malloc() calls. There is a slim possibility an underlying library is behaving differently, but that seems like a stretch.

adbancroft:
Using the speedy itself to report free ram (send 'm' via the terminal):

- Master: 151
- PR: 842

Diff: 691

adbancroft:
>     1. The tune needs to be reloaded and I imagine an update routine to fix that would be fairly painful. Not a major thing, but worth considering

@noisymime It shouldn't - the INI hasn't changed & the EEPROM layout _shouldn't_ have changed either. I can write a tune using master, open it using the PR & vice versa. I.e. round trip What are you seeing?

adbancroft:
>     2. For some reason I'm seeing one of the interval unit tests failing with this. Not obvious why, but I'll have a look into it.

For me, running PIO Test generates the same 2 failures on master & the PR: 
```
test\test_misc\tests_misc.cpp:57:test_initialisation_outputs_V03:FAIL: Expected 1 Was 0. Stepper Enable [FAILED]
test\test_misc\tests_misc.cpp:86:test_initialisation_outputs_MX5_8995:FAIL: Expected 1 Was 0. Stepper Enable    [FAILED]
```
Let me know what you find

noisymime:
> For me, running PIO Test generates the same 2 failures on master & the PR:
> ...
> Let me know what you find

All good, this turned out to be unrelated to this PR

PR #670, opened 2021-09-29
Enable internal pressure sensor for Baro updates

HWright9:
Intended for TPS only configurations. This re-purposes the **unused_bit** in VVT control which was obsolete and not available for calibration. 
This memory location is now used for enabling the internal pressure sensor for Baro updates. This setting is mutually exclusive with the existing **Use external Baro** setting so that it is not possible to enable both.

Baro updates are limited to valid ranges of the MAP sensor and only update at 1kpa per sec to limit any effects of noise or sensor dropout on the baro signal.

image

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -617,7 +617,7 @@ page = 6
       useExtBaro  =   bits, U08,      13, [6:6],      "No", "Yes"
       boostMode   =   bits, U08,      13, [7:7],      "Simple", "Full"
       boostPin    =   bits, U08,      14, [0:5],      "Board Default", "INVALID", "INVALID", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "22", "23", "24", "25", "26", "27", "28", "29", "30", "31", "32", "33", "34", "35", "36", "37", "38", "39", "40", "41", "42", "43", "44", "45", "46", "47", "48", "49", "50", "51", "52", "53", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID", "INVALID"
-      unused_bit  =   bits, U08,      14, [6:6],      "No", "Yes"
+      useIntBaro  =   bits, U08,      14, [6:6],      "No", "Yes"
       useEMAP     =   bits, U08,      14, [7:7],      "No", "Yes"
       brvBins     = array,  U08,      15, [6],        "V",        0.1,    0,    6,    24,         1 ; Bins for the battery reference voltage
       injBatRates = array,  U08,      21, [6],        "%",        1,      0,    0,    255,        0 ;Values for injector pulsewidth vs voltage
@@ -1848,6 +1848,7 @@ menuDialog = main
   includeAFR        = "When enabled, the current AFR reading is incorporated directly in the pulsewidth calculation as a percentage of the current target ratio. VE table must be retuned when this value is changed. "
   incorporateAFR    = "When enabled, the AFR stoich/AFR target -ratio is incorporated directly in the pulsewidth calculation. When enabled, AFR target table affects pulsewidth even with EGO disabled. VE table must be retuned when this value is changed."
   useExtBaro        = "By Default, Speeduino will measure barometric pressure upon startup. Optionally however, a 2nd pressure sensor can be used to perform live barometric readings whilst the system is on."
+  useIntBaro        = "Use the internal MAP sensor for BARO updates. Make sure Internal MAP Sensor is not connected to the engine. Use with TPS only setup."
 
   flexEnabled       = "Turns on readings from the Flex sensor and enables the below adjustments"
   flexFreqLow       = "The frequency of the sensor at 0% ethanol (50Hz for standard GM/Continental sensor)"
@@ -2748,6 +2749,7 @@ menuDialog = main
 
         field = "#Baro Sensor"
         field = "Use external Baro sensor", useExtBaro
+        field = "Use Internal MAP Sensor for Baro", useIntBaro, { (useExtBaro == 0) }
         field = "Analog pin to use for ext. Baro sensor", baroPin,  { useExtBaro }
 
         settingSelector = "Common Pressure Sensors",  { useExtBaro }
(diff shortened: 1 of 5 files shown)


noisymime:
As part of any change, I think we probably need to improve some of the wording on these options to indicate exactly what they refer to. 'External Baro sensor' has started to become confusing since these were included on many boards anyway. 

Around the method of reading the sensor, why not just do a regular read but with a higher filter value? This should give a more accurate view of the baro pressure I would think.

HWright9:
Yes. I thought the pressure sensor interface should be re-written as:
Step 1: Enable and Calibrate Pressure Sensors A, B, C. on Pins XXX
Step 2: Assign Pressure sensors A, B, C to signals (MAP, BARO, EMAP) etc.
This will require a bit more abstraction in the code about how to read and filter the various signals.

But this PR as it is now is the minimum change required to the existing code to implement a baro feature using the "internal" sensor. Perhaps we can get this one in and then work on a more comprehensive re-write for a future update?

With regards to the filter. I wanted a way to prevent spurious MAP sensor readings from updating baro incorrectly and making major changes to fuelling. Also the filtering should always be very slow since Baro changes slow, 1kpa /sec is the slowest a low pass filter could update the signal anyway. This pr implements both of these requirements with two lines of code.
The inital baro value is still set on startup as it was before so a slow filter will not affect cranking.

HWright9:
"As part of any change, I think we probably need to improve some of the wording on these options to indicate exactly what they refer to. 'External Baro sensor' has started to become confusing since these were included on many boards anyway."

- Do you want me to rework the .ini file to remove the assigment of "internal" and "external"? before approving this PR?

noisymime:
In `readBaro()` you've got the new code for checking the MAP reading (https://github.com/noisymime/speeduino/pull/670/files#diff-cf6893cab52605f0de8b25fc56b15565daa8ea927d1ca9ed99a47ea299d71fcbR493) and updating it, but nothing to prevent this reading happening when the car is running. 

Surely you can only be reading the MAP sensor for baro readings when the engine isn't running? 

Also, why not just use the existing ADC filter macro?

HWright9:
> In `readBaro()` you've got the new code for checking the MAP reading (https://github.com/noisymime/speeduino/pull/670/files#diff-cf6893cab52605f0de8b25fc56b15565daa8ea927d1ca9ed99a47ea299d71fcbR493) and updating it, but nothing to prevent this reading happening when the car is running.
> 
> Surely you can only be reading the MAP sensor for baro readings when the engine isn't running?
> 
> Also, why not just use the existing ADC filter macro?

The MAP sensor is not connected to the engine, it's for a TPS only based setup.

Not using the ADC macro, because it's not required. Baro should only update very slowly, it's a simple 1 line of code to have it update at a maximum of 1 kpa /sec. To flip this around, why should we spend extra controller resources to update Baro faster?

Also in the same line of code it performs the function of inhibiting a baro update if the MAP sensor value exceeds the range of allowed baro. This prevents a dropout in the MAP sensor from having Baro jump to it's minimum value and affect engine fueling. Remembering that this setup is only supposed to be used with TPS as the fuel load input so that MAP sensor is otherwise not doing anything. This should allow the vehicle to drive well enough to get the owner home if the MAP sensor failed. Continuing the logic from the "read once on startup" function.

HWright9:
Ok, as per discussion in #development. 

You can enable Baro sensor and assign it to the same Pin as the MAP sensor with no code changes to use the default on-board MAP sensor as a baro sensor. If you are otherwise not using the MAP sensor for anything (i.e. in a TPS only setup) then this is handy.
Baro should be filtered pretty heavily. But there is a problem with the inline macro filters in that they do not converge when heavy filtering is applied. I have an easy fix for this which turns the inline marco into a little function with a check for non-convergence.
#define ADC_FILTER(input, alpha, prior) (((long)input * (256 - alpha) + ((long)prior * alpha))) >> 8
becomes
uint16_t filterADC(uint32_t input, uint32_t alpha, uint32_t prior)
{
  uint16_t result = (uint16_t)prior;
  if (input != prior) // only need to calculate filter if there is a difference
  {
    result = (((input * (256 - alpha)) + (prior * alpha))) >> 8;
    if (result == (uint16_t)prior) { result = (uint16_t)input; } // Ensures filter convergence to input value with heavy filters. In practice when alpha filters are greater than 128.
  }
  // else result = prior = input. and nothing to do.
 
  return result;
}
3. Baro when read by an external sensor does not initialize correctly. it starts from a zero value and runs through the filter, this could be a big problem when keying on and cranking immediately with heavier filters. Also a relatively simple fix to initialize the baro value and the range of baro should be limited somewhat in case of a sensor drop-out (as per MAP).
4. Pressure sensor config adds a calibratable pin for MAP and then all sensors are fully adjustable by the user. So whichever physical sensor you want to call "baro" and which one you want to call "MAP" is able to be set by the user. I imagine for many, using a dedicated MAP sensor (likely factory sensor) mounted to the engine manifold and connected to some ADC pin and the on-board sensor for BARO on A0 (most boards) would be a good option.

I did not limit baro to plausible ranges (60 and 105kPa) as I did before. But baro does now have a check for sensor dropout as per MAP.

HWright9:
I have no more changes for this PR. If approved pull it in.

HWright9:
Requesting again that this PR be pulled in.
I merged to the latest master and added a small change to make it more efficient.

To recap this PR does the following:
- Fixed a non-convergence issue with the existing ADC filters - This is a major fix! 
- Pressure sensor configuration updates, improving legabiltiy.
- Having No MAP sensor and using the onboard sensor for baro only (with TPS as fuel load) is clearer and neater.
- Ensures that all sensors initalise correctly on startup with no filter for the first loop. Otherwise running resets and quick cranking can cause strange behaviour.
- ~10% Loops/sec performance increase. This is maintained even when ADC filters are turned to maximum.

Testing on the simulator @1500rpm:
Baseline Master: 783 Loops/s
This PR: 850 Loops/s

Corey-Harding:
> Ensures that all sensors initalise correctly on startup with no filter for the first loop. Otherwise running resets and quick cranking can cause strange behaviour.

I notice random issues daily driving if I crank too fast I may have to wait a second and try again or even key off key on.  Continually holding the key trying to crank causes more issues.  And I feel like my tune is in a decent state.  So hoping this PR gets merged because a lot of users dont know to turn the key and wait before cranking and sometimes neither do I.

PR #672, opened 2021-10-03
Cylinder Trims to use Fuel Load

HWright9:
Changed from MAP to fuel load so these work with TPS. Now the code matches the .ini file
Added 2 and 3 cylinder engines to fuel trim (sequential must be enabled).
Fixed Logging issue not showing PW2 and instead showing PW1 again.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -2916,21 +2916,21 @@ menuDialog = main
 
     ;Fuel trim composite dialog
     dialog = inj_trim1TblTitle, "Channel #1"
-        panel = fuelTrimTable1Tbl,      { fuelTrimEnabled }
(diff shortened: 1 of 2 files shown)


reference/speeduino.ini
```diff
@@ -2916,21 +2916,21 @@ menuDialog = main
 
     ;Fuel trim composite dialog
     dialog = inj_trim1TblTitle, "Channel #1"
-        panel = fuelTrimTable1Tbl,      { fuelTrimEnabled }
+        panel = fuelTrimTable1Tbl,      { fuelTrimEnabled && nCylinders >= 2 }
```
noisymime, line 2919: Is the `nCylinders` check needed for trim table 1?

noisymime:
This looks good,. The Trim code is node something that has had much widespread use, so I suspect things like this have been missed.

PR #680, opened 2021-10-07
Variable Launch control

Fera31:
this is my first pr. I hope there are no problems ..
variable launch control
lnchSoftLim .The soft limit is this many rpm below the hard limit.
lnchHardLim. became the minimum rpms parameter
VarLaunchMax. is the maximum rpm parameter
VarLaunchPin. It is for the pin pot for the variation of the rpm
when you don't want to use. You must set minimum rpm and maximum rpm in the same number of rpm
each new implement has its description

sorry for my bad English
Sin título


Fera31:
> does it require a lot of work to add additional input for rolling launch switch that will hold at that rpm when switch is on ?

that was my first change. But then I came to the conclusion that it would not be so necessary. once the launch rpm has been set with the potentiometer.

for what reason would you put an additional for the change between fixed and variable?

MauriceRivers:
Hey, how well does this work?

PR #682, opened 2021-10-10
Added selectable pin function for idle air valve

emretunali:
Added selectable pin function for idle air valve.


adbancroft:

PR #684, opened 2021-10-11
Advance calculation improvements

DeionSi:
Fixes locked timing and cranking for multiply and added secondary table. The corrections were not applied last thus the degrees did not match.
Corrections are only applied once for add and multiply secondary table. This will probably mean those tunes will need to be modified, however I couldn't get the multiplied and added modes to work correctly on the previous build.
Advance1 and advance2 are only retrieved/calculated if required (meaning the table that isn't used will show 0 degrees in tunerstudio).
Performance improvements in some circumstances (unchanged when table 1 is active, 5-8% when table 2 is active).


DeionSi:
New pull request on the way

PR #686, opened 2021-10-17
Deduplicate ignition scheduler code

DeionSi:
- Refactor IgnitionSchedule1-8 variables into IgnitionSchedule[] array.
- Add compare, counter, timerEnable, timerDisable, startAngle, endAngle, degreesAfterTDC1 (previously channelXIgnDegrees) and channel to Schedule struct.
- The changes to the Schedule struct has the benefit of deduplicating multiple functions/code segments:
- - refreshIgnitionTiming
- - setIgnitionSchedule
- - ignitionScheduleInterrupt
- - checkIgnitionSchedule
- - calculateIgnitionAngle

This is a direct continuation of #648 and that pull request is used as base. It is also related to #578.

This change more or less only refactors code. The logic changes are that the scheduler callbacks are only set once in initialiseAll and a small optimization into calculateIgnitionAngleWithSpecificEndAngle for rotary.

ignitionSchedule[i].member calls seem to be slowed than ignitionSchedule->member calls. This is why most functions pass a pointer to the array member rather than the number of the ignition channel.

Tested on mega2560. Loops/sec reduced by 0,5% (from Tunerstudio). Size reduced by 1,8% and memory by 0,72% (PlatformIO compile info). I've done some basic testing on the bench and the ignition events appear to happen at the correct time (scope tested).

There are other areas where code can be deduplicated as a continuation of this change.

### DIFF
--- a/speeduino/board_avr2560.h
+++ b/speeduino/board_avr2560.h
@@ -101,24 +101,23 @@
   #define FUEL7_TIMER_DISABLE() TIMSK5 &= ~(1 << OCIE5C); //
   #define FUEL8_TIMER_DISABLE() TIMSK5 &= ~(1 << OCIE5B); //
 
-  //These have the TIFR5 bits set to 1 to clear the interrupt flag. This prevents a false interrupt being called the first time the channel is enabled.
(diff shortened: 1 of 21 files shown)


DeionSi:
Fixed the decoder and scheduler tests. Both are successful.

adbancroft:

PR #687, opened 2021-10-17
Correct COMPARE_TYPE and COUNTER_TYPE

DeionSi:
I believe these types are incorrectly set and prevents #686 from building on Teensy and F407.

### DIFF
--- a/speeduino/board_stm32_official.h
+++ b/speeduino/board_stm32_official.h
@@ -21,8 +21,8 @@
 */
 #define PORT_TYPE uint32_t
 #define PINMASK_TYPE uint32_t
-#define COMPARE_TYPE uint16_t
-#define COUNTER_TYPE uint16_t
+#define COMPARE_TYPE uint32_t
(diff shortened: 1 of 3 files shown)


VitorBoss:
Please don't touch STM32 code, the timers are 16 bit, the ones that are 32bit is used in 16bit mode to keep compatibility. That gave a lot o headache in the past and the current code works flawlessly.

noisymime:
> From what I can gather the counters and compares of STM32 and teensy 3.5 boards are 32 bit.

This is not correct at all for the Teensy 3.x boards. All the Flex timers on them are 16-bit in hardware. The likely reason that the variables you're seeing are 32-bit is because they have a feature to chain 2x 16-bit timers into a single 32-bit timer. 

I understand what you're getting at, but Speeduino is built around using 16-bit timers. If you change the values to 32-bit then you're going to cause all sorts of issues as the expected overflows won't work.

PR #688, opened 2021-10-19
TPS or MAP load for AFR Target table

shiznit304:
Allows the user to select MAP or TPS as the load source for AFR target table. It uses a single bit.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -580,7 +580,7 @@ page = 5
 #endif
 
       rpmBinsAFR  = array,  U08,   256,[   16],    "RPM",   100.0,     0.0,   100,   25500,      0
-      loadBinsAFR  = array,  U08,   272,[   16],    { bitStringValue(algorithmUnits ,  algorithm) },        2.0,      0.0,   0.0,   {fuelLoadMax},      0
+      loadBinsAFR  = array,  U08,   272,[   16],    { bitStringValue(algorithmUnits ,  afrLoadSource) },        2.0,      0.0,   0.0,   {afrLoadMax},      0
 
 ;--------------------------------------------------
 ;Start page 6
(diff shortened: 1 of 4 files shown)


ric355:
What is the reason for wanting the AFR Target to use a different load axis to whatever is set for the VE table? On the face of it, it just seems like an unnecessary complication that will be confusing when trying to tune an engine unless both are using the same load (which is how it works now).

HWright9:
> @ric355 when used with include afr target, it can work like power enrichment, a feature found on other ECUs

Incorporate AFR already works like this abit using the same axis as the main table. The "include AFR Target" you reference is really a proportional only O2 feedback comparing measured O2 to this table. But since it has no time delays, sensor checks or any other limitations it is extremely unreliable way to tune an engine. (IMO this "feature" should be removed)

I can see that if O2 feedback was disabled having the AFR table on a different axis would provide the power enrichment function you talk about. But I think for clarity it would be easier to add this as a seperate single axis table. i.e. Above TPS XX take the minimum of regular AFR table and [Insert new value or table]. 
Also For everyone that has O2 enabled and Incorporate AFR on (which IMO is the "correct" way to tune) it would be confusing to have AFR on a different axis to the main VE Table.

HWright9:
I am referring to the definition in speeduino.ino Where "Include AFR" uses the O2 sensor value.

`  if ( (configPage2.includeAFR == true) && (configPage6.egoType == 2) && (currentStatus.runSecs > configPage6.egoStartdelay) ) {
    iAFR = ((unsigned int)currentStatus.O2 << 7) / currentStatus.afrTarget;  //Include measured O2 AFR from sensor (vs target) if enabled
  }`

and "Incorporate AFR" Uses the table vs the stoic ratio (no O2 sensor feedback)
`  if ( (configPage2.incorporateAFR == true) && (configPage2.includeAFR == false) ) {
    iAFR = ((unsigned int)configPage2.stoich << 7) / currentStatus.afrTarget;  //Incorporate stoich / target AFR, if enabled.
  }`

The confusion you have is exactly why we should remove the "include AFR" method.

PR #693, opened 2021-10-22
Add MPX4100A pressure sensor;Fix Bosch 3 Bar TMAP baro setting

DeionSi:


DeionSi:
You are correct, thanks for the due diligence. I've made the change.

dantob:
We should rename a few of these while here

these have totally different transfers (I get -11 and 10 for min)
[MPX4250](https://www.nxp.com/files-static/sensors/doc/data_sheet/MPX4250.pdf) -> [MPX4250A](https://www.nxp.com/docs/en/data-sheet/MPX4250A.pdf)

minor, but non _**A**_ variants don't exist
MPX4115/MPXH6115/KP234 -> MPX4115/MPXH6115**A**/KP234
MPXH6300 -> MPXH6300**A**
MPX6400 -> MPX6400**A**

PR #696, opened 2021-11-03
Change loopsPerSecond to uint16_t

adbancroft:
It's currently uint32_t - seems a bit ambitious!

And the INI file specifies it as uint16.


noisymime:
Believe it or not there is a reason for this. Teensy 4 overflowed a `uint16`and hence the change to a 32-bit value, though it was always capped to 16-bit when sending to TS.

PR #698, opened 2021-11-05
don't manually set values for ADCFILTER defaults, use defines

dantob:
Fixes a few inconsistencies where defaults are set to different values, rename flex filter to match other filter ADC naming, expose flex filter adjustment in tunerstudio


noisymime:
The `#defines` for the default values are a nice addition, however the reason the filter name was different for flex is because it's not an ADC filter. Flex is a frequency based pulse signal, not an analog one, so it doesn't use the ADC. That's why the variable name for it didn't match the others.

dantob:
Is it still worth exposing the flex filter adjustment slider in tunerstudio? Also, what is the reason for limiting filters to 240 (F0) and not 254 (FE)?

VitorBoss:
The flex slider is on flex dialog, it don't need to be added on other place as it isn't analog but a frequency input

noisymime:
Vitor is correct with where the slider is located already. I don't see a need to add it with the ADC filters. 

The filters are limited to 240 as beyond that the response curve changes dramatically, to the point where it is largely useless even in the most extreme cases.

dantob:
> The flex slider is on flex dialog, it don't need to be added on other place as it isn't analog but a frequency input
Somehow I missed that it was in there already, should be good now

PR #702, opened 2021-11-07
Sum ign corrections and add to base advance

DeionSi:
This changes the way corrections are applied. Essentially the corrections are added together before being applied to the ignition table advance value.

PR #701 is a prerequisite of this PR as this changes IAT retard in the same way.

This change doesn't really fix anything except how calculations are performed at the int8 overflow.

If this is applied with #684 the tempAdvance2 overflow line in getAdvance() should be removed.


DeionSi:
New pull request on the way

PR #703, opened 2021-11-07
Ignition advance tests

DeionSi:
This is a series of tests for the ignition advance calculations. When it comes to secondary tables I've made the assumption that those values are to be applied/used before any corrections are made. The code comments explain the basic idea on how these tests work. The purpose of these tests is not to test individual corrections but rather the overall calculation.

Currently 397 of the 1244 tests fail on master. With PR #684, #701 and #702 there are 0 fails (unless something has gone wrong when splitting the work into these PRs).

The max/min values in this PR uses the values changed to in #701.
The commit "Compatibility before PR #684" should not be merged if PR #684 is merged.

### DIFF
--- a/test/test_advance/test_advance.cpp
+++ b/test/test_advance/test_advance.cpp
@@ -1,0 +1,29 @@
+#include <Arduino.h>
+#include <unity.h>
+#include "test_advanceDC.h"
+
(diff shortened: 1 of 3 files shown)


DeionSi:
New pull request on the way

PR #708, opened 2021-11-13
Encapsulate decoder initialization

adbancroft:
This is the first of multiple PRs to separate out the encoders from each other. This will make adding new encoders easier and minimize memory (.bss) usage.

Encapsulate decoder initialization into the existing `triggerSetup_xx()` functions. This essentially moves code from `init.ino` to `decoders.ino` so each decoder self configures it's internal and external state.

- 20 byte reduction in RAM
- 2688 byte reduction in Flash
- No loss in performance. Tested on Ardustim using 4G63, Missing Tooth 60-2 & Basic Distributor. Comparison logs below (bright green is master, dark green is this PR):

4G63:
4G63 at 4500 rpm

4 cylinder distributor
Basic Dizzy at 4500 rpm

Missing Tooth 60-2:
Missing Tooth 60-2 at 4500 rpm

### DIFF
--- a/speeduino/decoders.ino
+++ b/speeduino/decoders.ino
@@ -40,13 +40,6 @@ A full copy of the license may be found in the projects root directory
 #include "crankMaths.h"
 #include "timers.h"
 
-void (*triggerHandler)(); ///Pointer for the trigger function (Gets pointed to the relevant decoder)
-void (*triggerSecondaryHandler)(); ///Pointer for the secondary trigger function (Gets pointed to the relevant decoder)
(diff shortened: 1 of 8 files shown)


speeduino/decoders.ino
```diff
@@ -170,32 +166,34 @@ static inline void addToothLogEntry(unsigned long toothTime, bool whichTooth)
   } //Tooth/Composite log enabled
 }
 
+static inline bool isTrggered(const decoderISR &decoder, const uint8_t triggerState)
```
DeionSi, line 169: Misspelled?

adbancroft: Sharp eyes! Fixed

PR #714, opened 2021-11-17
Replace toothLastMinusOneToothTime calculations with lastGap

DeionSi:
*work in progress (only Missing Tooth and Dual Wheel decoders updated, more coming)*

toothLastMinusOneToothTime is only used to calculate the gap between the last two triggers/edges/teeth and this is done in quite a few places throughout the code. Use lastGap variable instead and set it to the already calculated curGap.

Bench-tested 0.4.3d Missing Tooth 60-2 gives 4% loopsPerSec increase at 6000RPM.


DeionSi:
Replaced by #721

PR #718, opened 2021-11-18
return early if sync loss 

rerpha:
or weird case rather than set tempRPM then execute the rest of the code

I don't really think this has any tangible benefits other than maybe a tiny amount of execution speed but I think it makes it more readable.


noisymime:
The way the code is now is done intentionally to comply with MISRA 2012 guidelines. Specifically R15.5:

> A function should have a single point of exit

Issue #716, opened 2021-11-18
Feature Request: Radiator fan indicator


github-account-0:
I am working on a pi dashboard using tuner studio and noticed there is no indicator available to see when your radiator fan is running.  It would be nice to have that feature.

I am also curious about adding buttons on the tuner studio dash to turn on your radiator fan and condenser fan.

npmstart-pray:
Since it's possible to have speeduino be a pwm controller for electric cooling fans, might I suggest that along with coolant temp displayed, we can see the duty cycle of the cooling fans? It might make config of Pwm controller a tad more accurate and faster.

pazi88:
> Since it's possible to have speeduino be a pwm controller for electric cooling fans, might I suggest that along with coolant temp displayed, we can see the duty cycle of the cooling fans? It might make config of Pwm controller a tad more accurate and faster.

PWM fan control and duty display here: https://github.com/noisymime/speeduino/pull/637

github-account-0:
> > Since it's possible to have speeduino be a pwm controller for electric cooling fans, might I suggest that along with coolant temp displayed, we can see the duty cycle of the cooling fans? It might make config of Pwm controller a tad more accurate and faster.
> 
> PWM fan control and duty display here: #637

I don't have a strong need for a pwm controlled fan. I just want to show if my rad fan is on or off. I might tackle this in a couple months when I have more free time.

npmstart-pray:
I’ve always felt my fan comes on too late and has to run too strongly because of it. Being able to adjust on/off and duty cycle would be awesome for me. Glad someone has requested it before me

github-account-0:
In testing the fan indicator on the dash shows on and off as expected alongside the radiator fan switching on and off.

npmstart-pray:
Nice!  Now an ammeter reading alternator current output reading would be sweet to verify charging system load.

mike501:
We already have a gauge that shows car system voltage. Doesn't that cover
the alternator? If it's not 14.5 (ISH) volts it's not working properly. If
you wanted a light you could do programmable output to turn on a pin at
lower than X volts.

npmstart-pray:
Not voltage, amperage. The current passing through the electrical system. You know the alternator can deliver 125A or whatever at “12vdc”, but that’s max output. If you put an ammeter on the system you can tell how hard it’s working…like when your electric fans are drawing 30A on high

github-account-0:
[...]

PR #719, opened 2021-11-20
Add Radiator Cooling Fan indicator

exploitagency:
This PR fulfills the Feature Request in issue #716

This PR adds a radiator cooling fan indicator to tuner studio and adds support to the speeduino code to do so along with cleaning up some of the formatting(spacing deleting tabs etc) of the fan control portion of auxiliaries.ino


github-account-0:
`
indicator = { fanOn       }, "Radiator Fan OFF",     "Radiator Fan ON",     white, black, red,      black
`
Was thinking of simplifying this PR, would this also work?

DeionSi:
What's the correct term? Cooling fan or radiator fan?

github-account-0:
> What's the correct term? Cooling fan or radiator fan?

In my build I have to say radiator fan. I have another indicator for my ac consensor fan.

exploitagency:
I renamed everything as "FAN".  This allows future proofing of adding a "FAN2" or "FAN_2".

The user is also able to change what the indicator says within Tuner Studio so I think this is the best solution.  Let me know if I need to make anymore changes.

exploitagency:
I just added datalogging of the radiator cooling fan status.

I can revert the changes though if for some reason it is unwanted and I can also make any additional changes as needed.

pazi88:
There is already "fanOn" bit on the currentStatus -struct. Why didn't you use that?

exploitagency:
I didn't want to break anything but I think this way will be ok.

Using currentStatus.status4 I believe will actually save 7 bits of memory because a bool is actually a byte.

exploitagency:
I pulled the latest PR with the PWM fan control and am now even with master and believe I have made all the necessary changes.

exploitagency:
I did some formatting and fixed a mistake where I missed my () somehow during a copy/paste.

I tested it before the PWM Fan PR but it needs to be reviewed and I need to test it again and if someone running the PWM Fan could test it as well that would be great.

pazi88:
With quick glance, looks ok now.

exploitagency:
Thanks.  If I upgrade my ECU I may try your new changes out.

As long as I didn't screw anything up doing all the formatting changes I think its ok too.

exploitagency:
PR is even with Master with no conflicts and should be ready to merge

I tested it on a MEGA and it works as expected.  I have not tested it with PWM fan control but I do not see a reason why it should not also work as expected.  It would be nice if someone could test it with a PWM fan control enabled ECU with PWM fan control enabled.  But again I can't see a reason why it should not also behave as expected.

noisymime:
Tested and working OK here by the looks. Nice clean PR!

noisymime:
Fixes #716

PR #721, opened 2021-11-20
Replace toothLastMinusOneToothTime calculations with lastGap

DeionSi:
toothLastMinusOneToothTime is only used to calculate the gap between the last two triggers/edges/teeth and this is done in quite a few places throughout the code. Use lastGap variable instead and set it to the almost always already calculated curGap.

All patterns have been tested for basic sync/rpm readings with the following exceptions:
Jeep 2000 not in ardustim, untested
Harley EVO not in ardustim, untested
36-2-1 not in ardustim, untested (but similar decoder 36-2-2-2 works)
Mazda AU does not sync on master

Bench-tested 0.4.3d with Ardu-Stim:
Basic Distributor gives ~0,65% loops/sec increase across the RPM range
Missing Tooth 60-2 gives loops/sec increase of 0,8% at 1000RPM and 2,4% at 6000RPM.

4 bytes less memory but 66 bytes larger firmware.

### DIFF
--- a/speeduino/crankMaths.ino
+++ b/speeduino/crankMaths.ino
@@ -40,7 +40,7 @@ unsigned long angleToTime(int16_t angle, byte method)
         if(BIT_CHECK(decoderState, BIT_DECODER_TOOTH_ANG_CORRECT))
         {
           noInterrupts();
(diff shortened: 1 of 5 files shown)


adbancroft:

PR #722, opened 2021-11-21
A flat shift for AT transmission implementation.  

XPbIM3:
This algorithm will detect if AT tranny in upshift event and set ignition to a certain retarded values.  A limitations will not be applied until ATFS_RPM_MIN reached,  then advance will be set as an interpolation between adv_when_min:adv_when_max in a ATFS_RPM_MIN:ATFS_RPM_MAX range.  above ATFS_RPM_MAX hard cutoff will be done.   To enable this in GUI a bit in configPage6 is used, the former  VVT_on/off bit.    Also 7 bytes in configPage9 is used to setup MAP/KPA/RPM thresholds.

This algorithm is used by me (XPbIM3) to drive an m104 3.2 mercedes engine paired with MB 722.4 automatic transmission.  Works nice.


XPbIM3:
Things to consider doing:
- Add a valid default values to base tune. Set AT flatshift to disable.
- Double check for interoperation with launch start.
- Merge Enable bit for MT flat shift and enable bit fot AT flat shift into one single dropdown.

nuosu:
Can an update be made for the latest firmware? I tested it on my Aisin 4AT. perfectly worked

XPbIM3:
Glad to hear the feedback!
Yes, sure ,   I will do an update,   stay tuned.

пн, 21 мар. 2022 г. в 08:50, nuosu ***@***.***>:

PR #723, opened 2021-11-22
Improve missing tooth decoder

DeionSi:
A set of improvements for the missing tooth decoder.

Bench-test with 0.4.3d and ardu-stim 60-2 pattern. 0,84% additional loops/sec at 1000RPM, 1,27% at 3k, 2,81% at 6k. 236 bytes smaller firmware.

The changes look bigger than they are, but the individual commits should clarify things.


noisymime:
The speed improvement here I'm guessing is primarily coming from reduced number of checks for the missing tooth? Whilst I get what you're aiming for, you're only looking for the missing tooth where you already expect to find it. In a perfect world this will work, but you'll miss situations where a pulse has been missed at some point during the cycle. 

Has this been tested on an engine or only stim?

DeionSi:
Fewer checks is probably where most of the speed improvement is coming from. It has only been bench-tested.

I see a possibility where if you miss a tooth there will be a larger gap which would be interpreted as the missing tooth. This could cause early ignition events leading to engine damage and other issues. Not accepting an early missing tooth seems like a safer option.

noisymime:
Closing this as I don't believe the benefits outweigh the increase risks

PR #724, opened 2021-11-24
Add tacho sweep function

wonder99:
This enables the "Tachometer sweep on startup" in TunerStudio.  The target RPM is slowly ramped at startup for the first second, then holds that level for the next half-second.

Slight tweak to the TCNT2 preload value, based on timing measured on an actual board.  Loading the timer at the start of the loop eliminates jitter because varying ISR execution time is eliminated.

Also defines the loop counter terminal counts to be prime numbers, to limit the occurrence of multiple bits being set per loop, which should reduce overall jitter.

Sorry there are few little things in this one commit.  I can break them up if needed.

### DIFF
--- a/test/test_misc/tests_init.cpp
+++ b/test/test_misc/tests_init.cpp
@@ -1,6 +1,7 @@
 #include <globals.h>
 #include <init.h>
 #include <unity.h>
+#include <storage.h>
 #include "tests_init.h"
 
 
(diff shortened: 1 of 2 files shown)


shiznit304:
I'm running it and it works nicely

wonder99:
That's great, thanks for testing!

PR #727, opened 2021-11-27
Consolidation of setIgnitionSchedule functions

TBAMax:
Combine setIgnitionSchedule functions into a new generic function. This reduces code duplication in this area quite a lot.

This is a continuation/alternative from #686 and #578 . Only taking some(better) parts from the both and keeping changes in the manageable level. Only the code parts that are in the main loop are modified, interrupts are left as is at the moment, because do not want to add extra function calls to the interrupts. And also this would be to probably too much for single PR.
It was discovered that ignitionSchedule[i].member calls seem to be slowed than ignitionSchedule->member calls so this uses the latter variant. Also this seems to be the safer option.


TBAMax:
Replaced by #804

PR #728, opened 2021-11-27
Assignable auxiliary timers / PWM fan for Mega

DeionSi:
Request for comments.

Change the Boost/Idle/VVT timers into general "auxiliary" timers to be used for any feature. Timers are only assigned to enabled features. This makes a timer available for PWM fan on Mega. Currently WMI uses VVT timer but should be freed from that constraint and included in this auxiliary timer functionality. This will be cross-MCU and those with more timers will be able to run more timer features at the same time.

This is just a proof of concept but the code is working for mega. Needs more work, especially in updating the board files. Tunerstudio needs to warn if too many timer features are enabled. I'll continue development if there is interest.

Question 1: Currently xxxTimer pointers are checked before accessing (except for in interrupts). This could cause issues if this check is missed. Executing a nullptr callback probably causes issues? For a small memory tradeoff of one dummy AuxTimer assigned to unused feature could be used instead of all the checks.

Question 2: I know there were some discussions on Slack about using some kind of software timer which can trigger multiple timers. Perhaps that is a more future-proof solution if it's not too expensive in terms of performance/memory usage.

### DIFF
--- a/speeduino/auxiliaries.h
+++ b/speeduino/auxiliaries.h
@@ -64,13 +64,11 @@ long boost_pwm_target_value;
 long boost_cl_target_boost;
 byte boostCounter;
 byte vvtCounter;
-#if defined(PWM_FAN_AVAILABLE)//PWM fan not available on Arduino MEGA
 volatile bool fan_pwm_state;
 unsigned int fan_pwm_max_count; //Used for variable PWM frequency
 volatile unsigned int fan_pwm_cur_value;
 long fan_pwm_value;
 void fanInterrupt();
-#endif
 uint32_t vvtWarmTime;
 bool vvtIsHot;
 bool vvtTimeHold;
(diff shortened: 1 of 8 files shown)


VitorBoss:
Did you consider the frequency limitations? Currently both VVT/WMI runs at the same frequency, some PWM can controllers use a fixed low frequency input and I know VVT valves use a lot in most cases

DeionSi:
The "assignable timers" definition might be a bit misleading as it's actually "assignable compares/counters/interrupts".

The timers actual resolution (uS/tick) is unchanged. 16 for avr, 32 for teensy 3.5, 2 for teensy 4.1 and TIMER_RESOLUTION for stm32. These are the same per platform for all the boost/vvt/fan/wmi/idle pwm timers.

The output frequency is controlled by the xxx_pwm_max_count variables which are calculated from the uS/tick and frequency settings in tunerstudio.

PR #731, opened 2021-11-27
Clean up initialization of PWM max count

DeionSi:
The initializations of the xxx_pwm_max_count variables are spread out throughout the code and in the board files. Only set the MCU specific timer resolution in the board files and set the variables where the features are normally initialized.


pazi88:
Pretty sure the reason why those initializations currently are like they are, is because Teensy requires those to be like that to work properly. Have you verified this PR works with Teensy?

DeionSi:
I bench tested this on a Speeduino 0.4.3d. I don't have access to a Teensy for testing.

The xxx_pwm_max_count variables are only set in the board files, never used. I don't see why Teensy would require a different solution to AVR or STM32. Please elaborate.

adbancroft:
Closing - duplicated by #1537

PR #732, opened 2021-11-28
Dual O2 closed loop

HWright9:
Changes to allow dual closed loop from two independent O2 sensors. 
Includes all changes from PR #650.

Also has the fix for injector open time and structure and efficiency improvements for independent cylinder trim functions.

Select which injectors are on bank1 (EGO1) and which are on bank 2 (EGO2) to align with the feedback.
O2 control settings are the same for each bank.

Example of new injector bank select. Note this is related to the number of **injectors** not commonly used **cylinders**
DualO2_Injectors

**Example of AFR setup for narrowband**
DualO2_AFRConfig


HWright9:
To reduce EEPROM usage by 10bytes. I removed the tables and used 5 constants. The new constants scale the agressiveness of adjustment between 0% (no adjustment) and 100% (reaching the target in one step - at least theoretically). This should be easier for the novice to tune and allows the EEPROM reduction.

Calibration parameters now look like this:
image

HWright9:
Hey All, I updated this to be aligned with the master after #756 was added.
But also pointing out that we are out of memory in config page 9. With this plus DFCO changes. I had to cut back the coolant protection bins to 3 from 6 to make this fit. We should decide if this PR is going to make it or not.

Tjeerdie:
Hi HWright. Did you notice the PID ego control with wideband lambda actually does not work at all during your investigations/testing? I made a PR to fix this. But i will not go further with improvements on EGO PID until this PR is either abandoned or in the master to prevent a host of other issues.

couple of questions for this:
Why the complexity in one big PR? (making it impossible to review). I would have proposed the following 4 PR split up. It is just a suggestion no guarantee it will be merged if you do.

Following PR's
1. Fix any issues with trim corrections on whole PW and not effective opening time. Making an effective PW and only adding inj opening time at the end was a good call  ;-). 
2. Improve EGO closed loop control with the 1 sensor. Make your controller as a third option in controller strategies (Original simple controller, Original PID controller, New PI controller? ). This would keep it backwards compatible for users having tuned their PID or simple controller. My simple controller settings would now be gone and i would be left without a working EGO after upgrading? I could not find where in updates.ino the simple controller are transferred/translated to your new version? maybe it still is there?
3. PR to add the cycle rich lean for catalytic converters. With an option to disable it for users without a 3 way cat but still want to use EGO control. (Adding/subtracting some correction % + AFRtarget value every x seconds. This should work irrespective of controller used).
4. Add the 2 bank controllers. Use the same controllers (Simple, PID, new strategy) but separated over 2 banks. (just instantiate two controllers, one for each bank, there should not be any interference between the 2 bank controllers anyway. Apply the corrections according to the injector pairings set in TS after each controller is done calculting). 

I could imagine @noisymime does not want to add this PR because it seems to break current installs (I could not find where the simple controller is "transformed" into the new one. And the PR it is huge! So very very hard to review at once. Hard to see if this unintentionally influences other parts of the system. And it removes the original PID controller completely as far as i can see?

From a control perspective. Do you use fix size integral steps depending if the error >X AFR or <X AFR? A usual PID controller in speeduino case will do ITerm += (ki * error);. So if the error is small the steps taken are also smaller proportionally. Your controller seems to be an adaptation of the simple controller more than adaptation of a PI controller? 

Attached a screenshot of the EGO PID controller with mainly I gain. As you can see the controllers steps quickly towards target when the error is "big". and lots slower when nearly at the target. At the moment the controller is severely limited by the resolution of both the AFR and Ego corr resolution. 
EGO control

Please correct me where i might misunderstood some things. Its just such a huge change i have a hard time following the intentions and changes of this PR.

HWright9:
Hi Tjeerdie,
Thanks for the feedback. You are mostly correct in your assumptions. First on the Large PR. I originally had this as a smaller change to just single bank O2 (pr650), but based on some slack discussion there was a lot of interest in dual O2. To efficiently integrate the dual O2 the software needs the concept of "engine bank 1 and 2" to align with "O2_1 and O2_2" But this didn't exist, trying to jam it in the existing fuel trim area was creating a huge mess, so it was better to just neaten it up. But while doing that I found all the issues with injector open time. So the reason it's so large is I only tested this as a whole package, we could break up the pr's as you say no problem.

I don't mind doing the work and even suggested breaking it up before, but I basically had no feedback from @noisymime or really anyone else until now and I wasn't going to do all that work for no reason. I have 6 other open PRs ranging from a very simple and clear improvement, to large ones like this. None have been pulled and worse, sometimes other PRs that half fix the issue and use the memory I was planning on using are pulled in ahead of mine, making maintainance a lot of work. So the motivation on my side is low unless we can work on a plan to get this in together.

In answer to some of your specific questions.
1. No worries, we could do this.
2. The issue is a lot of RAM, EEPROM and processing efficiency was saved by removing the old Arduino generic PID controller, there are fundamental reasons it doesn't work properly with this O2 control related to it's fixed time steps vs this variable time step control. I don't see any reason to keep the old controller. It is possible to calibrate this new controller to replicate the simple controller. I would rather create some instructions in the Wiki on how to do this rather than keep the old code if it creates inefficiency.
3. You can currently disable the swing around stoich by setting it to 0 or integrator delay to 0. The reason I didn't use changes in AFR target as the control is that not all users use AFR target correctly (incorporate AFR = On), it would be a lot simplier if we just removed the option to calibrte without AFR but I don't think that's going to get approved. This way works for both methods. (also it's the same as some other aftermarket ECU brands and the way OEM ecu's do it so should be familiar).
4. That's pretty much how it works. The 2 controllers are just in the same function to share/re-use some variables and save RAM. The compensations are done in a very similar manner to injector trim (but you have to move injector trim otherwise it's a mess as mentioned above).

For tuning and backwards compatabiltiy. I haven't created any update.ino yet. Waiting on direction, I think the best option is to set up update.ino to replicate the simple controller but with O2 control disabled by default. Once users enable it, they should be very quick to set up the new controller. Old PID does not translate to this controller, but this controller is much faster to set up.

For control theory. This is still an integral only controller. You don't need P or D for closed loop control because there is no system 'inertia'. i.e. in a perfect condition if you were 20% lean and you add 20% fuel in one step with the correct transport delay time then you are at target with no overshoot. Pure integral.
Now in real life with measurement uncertainty (noise) we will take smaller steps, so that's what the adjustment factors are. If you set them all to 100% you will take one step to the target, 50% moves up 50% at a time etc. Since you have two adjustment points you can make it react faster when a long way from the target and slower when closer. It should be very simple to tune this way and prevents users having to deal with any complicated control theory. I really want to write a nice Wiki article to go along with this PR.

I don't think adding more resolution would solve the issues with AFR noise you are seeing. And all of the compensation functions have 1% accuracy so making O2 more accurate probably won't do a lot in the bigger picture. All I will say is that my AFR sensor (Motec PLM) is not nearly as noisy as that, have you confirmed if it's electrical noise or if it's really the sensor seeing that?

Tjeerdie:
General. The simple and PID controller do work. (When all bugs are fixed). But Improvements can be made. And yes PR are kept open to long without feedback.. I stopped making bigger changes like this PR because of the hassle. Its an energy drain. 

Point 1: Nice! That would make the code a bit clearer and easier to read. 
Point 2: Lets agree to disagree. Breaking current installs is a big fat dis-qualifier for PR if you ask me. You even removed the conditional enables like TPS% etc. 
Point 3: Agree to disagree. Separating feedback control from the stoic swing for 3 way cats would be easy to do and would make the controller a lot less spaghetti code. It could be implemented as a separated new correction type. 
Point 4: Again lets agree to disagree. I do not clearly see 2 controllers in the code. Only one intertwined controller with 2 corrections in one function. Making more spaghetti code

`but this controller is much faster to set up.` For you maybe? There are a lot of variables you need to tune, transport delay + sensor delay + 4! gains? Update rate engine cycles high and low. Lets say its subjective how easy it is to tune.. 

For the control theories part:
You right in theory the (varying) time delay of the O2 sensor can indeed be a problem. It limits the maximum gains possible. But somehow it does not seems so bad in practice with small gains? (See my attachment with PID controller).

If i understand you correctly you made a time delayed version of “Multiply by ratio of AFR to Target AFR”? This code is where the current AFR reading is taken and applied as [current Afr reading/Target AFR]*Effective PW. So when current AFR is 10% lean of target apply 10% more effective PW. Not sure who made it but it seems to follow your logic. 

You are smarter and put a gain on it of say 80% for noise and such. So your configPage9.egoInt_Lean1 = 80% you get a 8% effective PW compensation for that. (10%*80%). And this is only applied, cumulative (integrated), every X time delay depending on your sensor+transport delay?

If I look at you code I can not find that actually?

For the integral part
Line:726: `ego_Integral = ego_Integral + (int8_t)(table2D_getValue(&ego_IntegralTable, O2_Error) - OFFSET_AFR_ERR); `

Integeral seems to read from a fixed table with the error as index: The table is populated in line 674 with fixed values from the config. 
```

egoIntAFR_Values[0] = configPage9.egoInt_Lean2; // Corresponds with -3.0 AFR Error
egoIntAFR_Values[1] = configPage9.egoInt_Lean1; // Corresponds with -0.3 AFR Error
egoIntAFR_Values[2] = OFFSET_AFR_ERR; //offset value is 0 adjustment
egoIntAFR_Values[3] = configPage9.egoInt_Rich1; // Corresponds with 0.3 AFR Error
egoIntAFR_Values[4] = configPage9.egoInt_Rich2; // Corresponds with 3.0 AFR Error
```
and line 282: `ego_IntegralTable.values = egoIntAFR_Values;`

I do not see where it is proportional to the error. I was looking for a line similar to this:
egocorrection =  ego_Integral  * AFRerrorPercentage;  

But i could not find where it is. I assume you tested this and that I am wrong :-D. Could you please share the results of these tests with some logs? 

Indeed PI only controller (even a I only controller) would suffice.

And my sensor noise is not caused by the limited 1% steps. But it ALSO limits my control over AFR. There is areas where it is a lot less noisy:
image

HWright9:
Hi,

Ok lets go through it.

>Point 2: Lets agree to disagree. Breaking current installs is a big fat dis-qualifier for PR if you ask me. You even removed the conditional enables like TPS% etc

Totally dissagree, Keeping these functions causes us to use additional Memory and EEPROM which we have precious little of and causes a long term effect of splitting the code into multiple methods of calibration.  If it's not possible to change the way a function is calibrated once it's added what do you think the code and the functions will look like after a few years of this? In all of my experince of powertrain development, software and hardware re-factoring has been constant. I think using the default values a reasonable closed loop calibration can be created in about 15mins on a warm engine mostly at idle with a short drive to confirm. That's a small price to pay for having the additional functionality.

>Point 3: Agree to disagree. Separating feedback control from the stoic swing for 3 way cats would be easy to do and would make the controller a lot less spaghetti code. It could be implemented as a separated new correction type.

Each O2 bank are independent for feedback (i.e. once can be rich swing while the other is lean swing), so how would you suggest we implement this? Right now there is only 1 AFR target signal. We would have to duplicate the AFR target for each bank add the swing and change the main PW calculation to calculate for each bank using the corresponding modified AFR target. This is in addition to the g_ego corrections which are also happening for each bank. I think this would add extra computation time and more variables to watch. The current method of just having Gego and Gego2 toggle when near stoich works pretty well and has minimal impact to the rest of the code. This also works with AutoTune, which as far as I know can only read the AFR target from the target table.

>Point 4: Again lets agree to disagree. I do not clearly see 2 controllers in the code. Only one intertwined controller with 2 corrections in one function. Making more spaghetti code

The reason it's that way is to re-use some local variables and save RAM. The two controllers are one after the other re-using a lot of the same enable conditions etc. It can totally be split up into more functions, but only at the expense of more RAM (about 8 bytes) and I would have to pass a few pointers.

>If i understand you correctly you made a time delayed version of “Multiply by ratio of AFR to Target AFR”? This code is where the current AFR reading is taken and applied as [current Afr reading/Target AFR]*Effective PW. So when current AFR is 10% lean of target apply 10% more effective PW. Not sure who made it but it seems to follow your logic.

Nope. That code is a proporional only controller. It has no integral function. My code has an integrator. I think where you are getting confused is that the implementation of the integral step is not the classical IntegralGain*Error. Rather the error is fed into a lookup table and the integral step is the lookup value. The reasons to do this are so you can safely bound the integral step to the x axis range of the table (so very large errors dont make huge integral steps) and easily create a step which varies in a non-linear way with the error. I would normally have this as a simple 2D table and probably a few more axis points but because of the EEPROM limitaitons I cut it down to just 4 points with a hardcoded axis and instead of the integral step in units of Gego I scaled them into 0-100% step values to make it easier to calibrate. 

>And my sensor noise is not caused by the limited 1% steps. But it ALSO limits my control over AFR. There is areas where it is a lot less noisy

Yes I get it and I can see the benifit. But fundamentally most of the speeduino corrections are only 8 bit math with 1% error. We could change that for AFR and closed loop but then the next function would be the limitng factor for fuel control accuracy etc. Comes back to my original point about limited EEPROM and Memory. Are people willing to lose some accessory functions so we can spend the memroy and EEPROM to have better primary functions? Probably we won't get everyone to agree on what to remove.

atomspring:
I'm interested in this capability as I have a dual exhaust 4cyl bike I'm working on. Are we for sure sticking with these hardware limitations long term? I obviously defer to you guys, but... a lot of this back and forth centers on memory constraints.

I'm using a teensy4, so it'd be interesting to possibly enable features which are more memory intensive based on the core hardware being used.

HWright9:
> I'm interested in this capability as I have a dual exhaust 4cyl bike I'm working on. Are we for sure sticking with these hardware limitations long term? I obviously defer to you guys, but... a lot of this back and forth centers on memory constraints.
> 
> I'm using a teensy4, so it'd be interesting to possibly enable features which are more memory intensive based on the core hardware being used.

Hi, a couple of things. First this PR is pretty old. Once I saw that it was unlikely to be pulled in I stopped maintaining it. I do have a compliled version of the code running in my own car with this change and my other PR's that haven't been pulled in, I have diverged somewhat from the main branch, but if you want to try it it's located here:
https://github.com/HWright9/speeduino

The dual O2 functionality works fine and I think achives a reasonable solution with minimal memory footprint and code overhead. It could be refactored to be a bit clearer but as of now it works. I drive with it reguarly in my own car using a Mega + NO2C.

The main issue that needs to be resolved is currently speedy doesn't have a clear structure for individual cylinder and "bank to bank" fuel modifications. I know there is something in there for individual cylinder trim but its not convenient to break into and edit with new modifier functions and there are issues with injector dead time being in the wrong place etc as discussed above. Any dual O2 code first needs to address this that's what I attempted to recify with this PR.

Your welcome to develop your own PR or sequence of PR's and try and get it in.

PR #733, opened 2021-11-29
Add possibility to read any value from Speeduino using OBD2

pazi88:
This simple PR adds functionality to read any of the values that are available in TS via OBD2. The standard OBD2 PIDS are quite limited of what data can be read from Speeduino. But with this PR, all the data is available if the OBD2 reader/dash etc. is configured properly. I have tested this with Torque Pro and Car Scanner Android apps using elm327 Bluetooth dongle. Here is example CSV for the Torque app using the Custom PIDs:
[SpeeduinoExamplePIDs.csv](https://github.com/noisymime/speeduino/files/7618067/SpeeduinoExamplePIDs.csv)

image


shiznit304:
Wow this is pretty neat. Gives me a reason to wire up my OBD2 wires. I wonder if I can get it to read from an iPhone with BlueDriver scan tool.

pazi88:
> Wow this is pretty neat. Gives me a reason to wire up my OBD2 wires. I wonder if I can get it to read from an iPhone with BlueDriver scan tool.

Yeah if it can read OBD2 data, then it should work. I just haven't used that particular tool, so don't know if you can add custom log items in that (which is mandatory feature for this PR to work). Otherwise you are limited to standard OBD2 items only.

shiznit304:
will this work on mega2560 based speeduinos?

PR #734, opened 2021-11-29
Fix bi-linear fixed point rounding

### LINKED ISSUE #726: 3D table bilinear interpolation can return values outside the table
The fixed point math used by the bi-linear interpolation always rounds down. This can result in a value that is outside the table range.

adbancroft:
1. Fix #726 
2. Add unit test for this bug
3. Add `native `platformio environment - can be used to run table tests, including `test_all_incrementing`.
4. Correct the test data so that `test_all_incrementing `passes.


adbancroft:
Could the native unit test runner be added to the CI build? It includes a longer running test that isn't normally run, which is how I found this bug.

adbancroft:
Code changes included in #735

PR #736, opened 2021-12-05
Improve closedloop boost control

Tjeerdie:
Improved closed loop boost control

Implemented improved boost control. It is fully backwards compatible with the original closed loop. The added functionality is:
1. Selectable closed-loop enable boost pressure (defaults back to barometric pressure as switch point, so it is backwards compatible).
2. Selectable fixed boost controller duty cycle. This sets the boost valve duty cycle when the closed loop controller in not enabled as point 1. (set to zero  for backward compatibility)
3. Boost valve duty cycle look-up table for desired boost. Depending on RPM for more controllability. (Table filled with all 50% DC for  backward compatibility).

The rational behind each point:
1. The point where the closed loop controller should enable is when the boost is just below the boost pressure only on wastegate. (this is determined by the spring stiffness of your wastegate actuator). Below this boost level you can set the duty cycle of the boost control valve to every possible value but it has no effect. So the controller should be enabled only when it can actually control boost.
2. When boost is still building up to the point where the waste-gate opens on the spring you want the waste gate to be shut as tight as possible and not leaking a little bit. So a fixed 99% duty cycle will help build boost as quick as possible when the controller is not enabled. 
3. If a system is designed and adjusted correctly lets say a 10% duty cycle on the boost control valve should increase the boost to 10% above the waste-gate spring level of boost. 20% for 20% above that level etc. In this lookup-table the desired closed-loop boost is on the Y axis and the RPM in the X axis. So for a desired boost level the controller looks in this table (LUT) to get the estimated duty cycle to get to that boost level. This ensures the system is already in the ballpark. So the controller must react less to get to the desired outcome. The controller still works on top of this LUT value to increase accuracy of the total system. This is called feed forward. And the user must tune the LUT to get close to the desired boost without controller. (Tune the LUT with all gains zero for example)

Tested in my car and on the bench. performs great for my setup at least. maybe @pazi88 can test or review it too.


jackscr12:
Yees, been waiting for this since we got closed loop boost,  I wish I could test this now but  winter has set in  :(
Did you manage to keep it compatible with the boost by gear?

Tjeerdie:
> Yees, been waiting for this since we got closed loop boost, I wish I could test this now but winter has set in :( Did you manage to keep it compatible with the boost by gear?

I am not sure, the boost by gear was not merged when i did this work i think? At least i didn't see (and thus touched) anything related to it. I will take a look if i might have mist something related to this

Tjeerdie:
From first glance the boost by gear only changes the boostTarget (Setpoint). So it still should work with boost by gear. Any piece of extra functionality that only changes boostTarget should work fine.

Tjeerdie:
Ok this is not easily to merge anymore because of the changes made in 4 different PR that interfere with the closedloop boost control function. My merging tool does not understand and neither do i

Tjeerdie:
@pazi88 I do not think you have tested this? Any interest? It works perfectly fine for me, still running in my car at the moment.

jackscr12:
I can test this  now, can it be updated to current master?

Tjeerdie:
@jackscr12 unfortunately Josh does not seem to be interested. I resolved merge conflicts for two months (dec to jan) and asked many times if this is going to be merged to Josh. The last time i asked the reply is that it needs thorough testing because one of the pages is extended to fit the LUT table (couple of months ago). But It is still open without any comment. 

I think the eeprom location intended for this function is now used by some newer PR but i will see. I try to make it mergable with the latest master. If it takes me more than tonight to resolve the issues it is sad but this one will never be merged. Feel free to use the idea of this function your self and make a PR. It does work great to get closed-loop boost going, the closed-loop only is just not usable at all (at least for me).

Tjeerdie:
This now has a fully untested merge. Not even bench tested. Try at your own risk.

jackscr12:
OK will do!

jackscr12:
So, the results after a short test : 
The valve duty below Threshold  stays at 99% (as set) until map reaches 95 and than goes to 0 for 0.2secs and up to whatever value is in the lookup table 
Map follows the Lookup Table duty cycles and seems to not be influenced at all by the closed loop PID (I set it in simple mode to reduce variables)
Boost by gear limits are ignored completely in Multiplied mode.(haven't tried constant limit as I don't see a point in that)

Tjeerdie:
Hi JAck

Thanks for testing! 

1. Did you also set the threshold to fixed instead of baro? and set the switch point to 95kpa? Or is it still at Baro switch point (the orginal hard coded in firmware) and is the baro reading 95kpa? In both cases this is indeed intended as it should be. The dip to 0 for 0.2sec is not ok. investigation needed. 
2. MAP should not follow the lookup table? Not sure how that came to be? the LUT should be RPM vs boostTarget, not MAP vs RPM. The boost target table should be MAP vs RPM. The LUT (lower table in TS) for me is 0% duty = 1 bar boost. 10% duty is 1.1bar boost 50% duty is 1.5 bar boost. etc. 
3. This PR should not influence the boostTarget (setpoint) in any way. So that must be a bug introduced by the merge? Maybe some old code has been added during the merge? You can see what code has changed. 

Maybe i will take a look at the bench again. everything described above did work perfectly fine originally. But the PR is open so long that merges means re-testing that takes hours of work to iron out bugs introduced. even for a small PR like this.

jackscr12:
1. the set point is defapt to baro. did not touch it. and yes a i realise  it is suposed to work that way I  wanted to say that it works as intended except the 0.2s drop. 
2. I think i worded it wrong there, i meant to say the actual map reading is what i would expect to be acording to the duty values i entered în the LUT  and for me that is 50% duty  1.4bar aprox.
3. I will check and see if i spot something off, and maybe reflash the firm.

jackscr12:
I will retest it in a few hours again. if there is something else you think i should look out for let me know. thank you

jackscr12:
OK  went out again  this time I changed the Boost control trigger to FIXED to 95kpa.
1.the boost by gear limits are ok don't know what happened the first time around
2.PID closed loop does absolutely nothing the duty cycle to the waste gate is the same one set in the LUT.
3.The dip to 0% duty  is behaving very weird... as in if  the boost slowly rises from 95-130kpa the dip is present the whole time(ex 2secs)until the TPS is reduced and the map drops below 95kpa and than it's back to 99%. So it is not 0% for a fixed time but varies according to something...seems like MAP but I could not find a correlation between a map value and the dip returning to a the value in the LUT.
And I was always under the impression that other ecus do it a bit differently : the duty is 99% when the map goes over 100 and stays there until a set point like 20kpa below the target MAP value is reached and than the PID takes over . 
I mean I do not see a point to run  the valve at 99% during cruising when there will not be enough pressure to open the gate anyway, or anything below 100kpa for that matter.

jackscr12:
[DataLogs.zip](https://github.com/noisymime/speeduino/files/8730333/DataLogs.zip)
ok I attached the log and tune.
But I just noticed I forgot to set the vertical axis in the LUT and it was the default 0-100... I'm a moron... this might be the reason the closed loop was inactive as well ...

jackscr12:
I set the boost threshold to a more reasonable(for my WG)  160kpa and I set the vertical axis on the LUT table as it should be from 0 to 250kpa. and going back for a test drive .
Sorry I wasted your time I was so eager to test this I messed it up.

Tjeerdie:
> I set the boost threshold to a more reasonable(for my WG) 160kpa and I set the vertical axis on the LUT table as it should be from 0 to 250kpa. and going back for a test drive . Sorry I wasted your time I was so eager to test this I messed it up.

Let me know if this works better. I hope the controller is actually working and controlling the pressure now. 

I would recommend to start the LUT at your threshold value (so 160kpa in your case) up to around double the WG boost pressure (so i guess around 220 kpa in your case?).

jackscr12:
> > I set the boost threshold to a more reasonable(for my WG) 160kpa and I set the vertical axis on the LUT table as it should be from 0 to 250kpa. and going back for a test drive . Sorry I wasted your time I was so eager to test this I messed it up.
> 
> Let me know if this works better. I hope the controller is actually working and controlling the pressure now.
> 
> I would recommend to start the LUT at your threshold value (so 160kpa in your case) up to around double the WG boost pressure (so i guess around 220 kpa in your case?).

Ok will start at 160  but my usual max running MAp is 250Kpa.. I am stretching this turbo as far as I can with pre and post turbo meth injection to keep some efficiency at high compressor rpm. The turbo is small but I need it that way for fast response , I do not care if it doesn't last long. witch is why I need  your PR as the regular CL Boost does not work for me because I need the waste gate duty to have a very steep increase after 5500rpm and the PID does not like that at all.. wasted allot of fuel trying to cooperate but it just was overly twitchy at low rpm or did not get to the high WG duty  at high rpm.
Open loop works great, but boost by gear not so much for the same reason as above. I think the code could be modded to work ,that's next maybe if CL does not cooperate.  :))

jackscr12:
I think this is working , I mean the it doesn't stay on the target very well but that is the lack of tuning for the PID . 
The testing wasn't ideal , i didn't have enough room to do a good pull in 4th but there is clear evidence the PID is trying to keep it on target . 
Tomorrow I will give it another shot and try to set  the Full PID parameters a bit more aggressive.
[DataLogs.zip](https://github.com/noisymime/speeduino/files/8732169/DataLogs.zip)

Tjeerdie:
Wel thats good to know. So with some LUT tuning and than add some gain it should work pretty well i hope. Let me know if you get it working or that some tuning parameters might be missing.

jackscr12:
YES ! This works great! I can finally have boost control that does what I tell it to do :)) I didn't even have to tune PID just the sensitivity slider and a good LUT. 
There is small issue with big implications :) and that is the boost cut did not engage like it should have and I don't know wth is going on because last I checked there was no code change there...or did I miss something?

Tjeerdie:
Boost cut is indeed completely separate. But there have been issues in the
past. And with updated read speed of various sensors this might be an issue
again. Please put it as possible issue in discord channel. 😃.

Happy that this works! Hopefully more people adopt this and it will be
added to the main branch.

jackscr12:
Yep this is a great addition hope it's not missed by @noisymime for the next firm release.

noisymime:
I've run through testing on this today and agree it looks good. The biggest problem I have with it by far though is the increasing in size of the memory page. This will break all tunes when loaded and will mean everyone needs to reload their tune via TunerStudio. 

I've tried in the past moving all the tables following the increased one back by the relevant number of bytes, but it's extremely messy and caused issues for many who upgraded. 

I think the alternative here is relatively simple though. Instead of adding 80 bytes onto table 6, a new page can be added with more than enough space for the LUT (Maybe 240 bytes). We've needed a new memory page for a while anyway and doing it this way will prevent a lot of issues. 

Thoughts?

Tjeerdie:
Hi Josh

If you take a good look the addressing is still consecutive. It's not
shifting any other pages in memory so i expect it not to have problems with
current tunes...?

The pages are getting a bit mangled address wise.

But if it is still problematic a new page is even better (And much cleaner
with all variables related to boos close together

noisymime:
> Hi Josh If you take a good look the addressing is still consecutive. It's not shifting any other pages in memory so i expect it not to have problems with current tunes...? The pages are getting a bit mangled address wise. But if it is still problematic a new page is even better (And much cleaner with all variables related to boos close together

Ahhh I hadn't seen they've you've split the page up in EEPROM. I'm actually amazed that even works and I'm pretty sure it will be breaking things like the fast EEPROM loading. Either way, I'd REALLY prefer to have all the page data contiguous where in EEPROM possible, so I think the new page is the way to go. It really shouldn't be a big change either, the page just needs to be defined and the table moved in the ini and globals. 

Do you want me to create the new page in a separate commit and then have this PR updated to move it over to there?

Tjeerdie:
I just stole the concept from another part in eeprom that worked the same
way 😃.

I am on holiday at the moment so earliest i can take a look at making a new
page is this Monday. Not sure what your release time line is at the moment?

Tjeerdie:
And we have merge conflicts again... I am getting fed-up changing configuration bit and bytes around because other PR are added breaking the config i planned to use. This time its the tacho setting... @noisymime is there any plan to prevent this in the future? At least add PR's the order they came in? 

I mean i spent at least twice the time fixing merge conflicts and testing again vs adding and testing the actual improvement! That is just a waste of time that could haven been prevented.

Tjeerdie:
Ok i think i am all done. A new page is added (Page 15) to eeprom and ini etc. This "should" work. But a lot has changed so please check carefully.

Corey-Harding:
A non-functional change could be updating the comment at the top of storage.h with the current map of memory to reflect the new page 15 changes.  When I was looking at a PR that also adds a new page 15(AC PR) I remember there also being several comments that were outdated.  I think one outdated commented was referring to the "current" eeprom version somewhere.  I know it technically doesn't matter and I am sure there are several comments that could be updated or removed.  I can't test this PR with my setup but it would be nice seeing a new page get added.

Tjeerdie:
Added the new page to the memory layout in the comments. Unfortunately i see i did not add the CRC32 for page 15. Also i see obvious room for the 4 bytes needed for this. @noisymime please advice. 

BTW why is adding the map on the end like the TRIM tables 4..8 such a bad idea? I think it nicer than adding a second page with some boost config + having the old boostVVT config page also still there? Anyway please advice on where to put the CRC32 in. (or maybe using update to swap the old locations to new to make room?)

noisymime:
I'll run through testing on this tomorrow and pull it in if no issues are found. It was the last big PR I was hoping to get in before next release, so the timing is good. 

> f the PR's are pulled in FIFO style that would also solve it. 

This is nearly impossible unfortunately as 80+% of PRs require reworking and changes between when they're submitted and when they get pulled. Then there's also the chance that they end up not getting pulled at all (For any number of reasons). 

I understand that the ini memory space conflicts are annoying as a result of this, but they're usually not too difficult to resolve. Conflicts in the actual code are more difficult, but also far less common thankfully. 

Great work on keeping at it with this one, having it in will be a really nice outcome.

Tjeerdie:
@noisymime thanks. Don't forget to add the 4 bytes of CRC32 for page 15. I try to test it this evening a bit better too to see if any merges did make some unexpected changes.

PR #738, opened 2021-12-07
Performance: speed up table I/O (optimized division)

adbancroft:
### Overview
This PR speeds up 3 operations:

1. EEPROM write: 17%
2. Send to TunerStudio: 11%
3. CRC calcs: 26%

It does so by optimizing the table axis value division using the [libdivide](https://github.com/ridiculousfish/libdivide) library

### Details

- During table I/O (write to EEPROM & send to TS), we compress the table axis values to 8-bits by dividing by a constant. 
- There is a well known trick of turning constant division into multiplication by the reciprocal: compilers apply this automatically in many cases
- **GCC will only apply this optimization when both operands are of register width or narrower.**
- **So on AVR, constant division of anything other than (u)int8_t is not optimized**
- **On AVR hardware, division is especially slow. There is no hardware divider; the compiler calls various division methods implemented in software (E.g. __divmodqi4)**
- This PR replaces the slow built in division routines with calls to `libdivide` using pre-computed parameters (since the axis I/O dividers are known compile time constants).


adbancroft:
There are likely many other parts of the code where we could apply this optimization:

Any 16 or 32 bit division by:
1. A constant
2. A 'semi-constant'. I.e. a divisor that is computed at run-time, but doesn't change much.

DeionSi:
Is it possible to "simply" overload the division operator with this library for the whole of speeduino?

adbancroft:
[Example showing the difference in 8-bit & 16-bit division](https://godbolt.org/z/1sP3d8fWh)

Note the call to `__divmodhi4 `for the 16-bit case, versus the multiply & shift operations for 8-bit.

adbancroft:
> Is it possible to "simply" overload the division operator with this library for the whole of speeduino?

Not effectively.

I see 3 cases, all probably need assessed case by case:
1. The divisor is an obvious compile time constant. E.g. `a/17`. We could find & replace these, but there is a small cost in Flash usage, so not worth doing if the code is rarely used. E.g. decoder setup functions.
2. The divisor is a compile time constant, but is passed into other code (E.g. as a function parameter). This was the case with the axis I/O. These aren't obvious and will need familiarity with the code and refactoring to apply libdivide.
3. The divisor is computed at runtime. Setting up the libdivide parameters has a runtime cost, so it's only worth doing when the divisor is computed much less often than it's applied.

noisymime:
This looks really nice (The C++ flavour of it aside).

Thinking this could be used for a number of the functions in `maths.ino` as well?

adbancroft:
> Thinking this could be used for a number of the functions in `maths.ino` as well?

Yup, `divu100()` & `divs100()` could be replaced completely. I already tested & using narrower types + `libdivide` is much faster.

adbancroft:
I do have a plan to use clang + python to find where division is used in the code base & then figure out if it's worth updating.....but that's longer term.

PR #741, opened 2021-12-11
Added loops per revolution gauge and adjustments to loops per second gauge

DeionSi:
Max values for gauges are set so the minimum values can be seen. Very high loop count is good but it's not really useful to visualize.


VitorBoss:
Just a thought, don't limit the maximum loops per second, it might not make sense for Mega2560 dut it does for Teensy and STM32, on my BlackPill I got 5900 loops/s on idle with a 60-2 wheel

DeionSi:
I am aware. My justification is that once loops/s goes above a certain limit there's not really any interest in how many loops/sec there is. It is mostly only interesting below 1000?

Issue #740, opened 2021-12-10
There is more than 1 ignition for 1 trigger input for 1,2,5 cylinder using basic distributor and single coil output. 


JuanNavarroS:
In basic distributor mode + single coil output, for 1, 2, 5 cyl, there is a OK spark in synchro with the trigger event (equivalent to breaker point lobe), but one extra spark KO in the middle between 2 trigger events.

jonbills:
I think there are two code issues involved here:

1) in init.ino, only the 3 cylinder case recognises that IGN_MODE_SINGLE is 'sequential' and sets for 720 degrees of crank. The others leave CRANK_ANGLE_MAX_IGN=360
2) 2 cylinder setup has 'even fire' defined as 180 degrees (which is confusing with boxer twins, which are 360 degrees apart and you'd expect that to be 'even fire')

so... 3 cylinders doesn't exhibit this 'issue' because it's explicitly coded as sequential, and 4,6,8 cylinders don't exhibit this 'issue' because the outputs are all overlaid on the single ign channel and the extra/redundant spark triggers coincide with legitimate spark triggers for other cylinders. 
I guess this isn't a practical problem even with 1,2,5 cylinder engines, because they must use a distributor to route the output sparks and the redundant sparks happen when the distributor rotor won't be connecting to an output plug lead.
Maybe this results in excess dwell on the coil for those sparks which can't be discharged?
I would think the 'proper' fix would be to code all IGN_MODE_SINGLE same as IGN_MODE_SEQUENTIAL for  all cylinder counts (and change 'even fire' for 2 cylinder to 360).

JuanNavarroS:
In answer to
> I guess this isn't a practical problem even with 1,2,5 cylinder engines, because they must use a distributor to route the output sparks

(Always talking of 4 stroke and even firing)
2 cylinder + single coil output = Yes it needs a distributor
2 cylinder + wasted spark = No it doesn't need a distributor, and should only work at 1 channel output (just 1 coil for 2 cyl). The problem is that with this setup speeduino is giving output both to channel 1 and 2, that is wrong. Additionally both channels are suffering the issues indicated in a comment by @mike501 in https://github.com/noisymime/speeduino/pull/578#pullrequestreview-660263652_

In fact 2 cylinder + single coil output should behave identically as 2 cylinder + wasted spark, the coil should fire as many times for both using just 1 ign output, not 2.

[...]

PR #744, opened 2021-12-12
fix even fire, four stroke, 2 cylinder crank angle 

jonbills:
ref issue #740
for four stroke engine, even fire is 360 degrees, and for two stroke it's 180 degrees.
This might break backward compatibility for anyone using a four stroke 180 degree crank, but they can easily fix with setting it (correctly) to odd fire.

### DIFF
--- a/speeduino/init.ino
+++ b/speeduino/init.ino
@@ -450,7 +450,8 @@ void initialiseAll()
         channel1IgnDegrees = 0;
         channel1InjDegrees = 0;
         maxIgnOutputs = 2;
-        if (configPage2.engineType == EVEN_FIRE ) { channel2IgnDegrees = 180; }
+        if      ((configPage2.strokes ==  TWO_STROKE) && (configPage2.engineType == EVEN_FIRE )) { channel2IgnDegrees = 180; }
+        else if ((configPage2.strokes == FOUR_STROKE) && (configPage2.engineType == EVEN_FIRE )) { channel2IgnDegrees = 360; }
         else { channel2IgnDegrees = configPage2.oddfire2; }
 
         //Sequential ignition works identically on a 2 cylinder whether it's odd or even fire (With the default being a 180 degree second cylinder). 


noisymime:
The problem is that in a wasted spark configuration, 2 cylinder even engines can be 360 or 180 crank angles. Both are even and both are correct. 

I can see your argument here, but a 180 degree 2 cylinder is still even fire in wasted spark (At least from the ECUs perspective), even though the actual power strokes have a gap. It's a difficult one given that this change will break existing tunes, which is always to be avoided unless it's impossible to convert.

jonbills:
Yeah, I thought backward compatibilty might be an issue. 
its easy enough for those with 360 cranks to see that as odd fire, even though imo its not logically correct.

JuanNavarroS:
Even firing engine is the one that have sparks (thus cycles for each cylinder) evently spaced (stagered) for all cylinders.

Even fire in 2 cyl engine, means that a complete cycle 720º / 2cyl = 360º firing, one spark at 0º, the other at +360º next at +360º = 720º=0º, cycle completed. Even spacing of 360º between sparks. That is called parallel twin
2 cyl parallel

Not to confuse 2 cyl boxer (aka flat twin or opposed cilinder engine) where "crank angle" between cylinders = 180º, with the "180º firing" naming
As you know, any engine should be "straigthened" to understand the real firing order, and a flat twin, when "straightened" (that is, to imaginary twist 180º the 2nd cylinder and its crankpin over the first cylinder to make it straigh), results in a straight twin cylinder with the two crank pins aligned, that is a 360º firing even engine.
2 cyl boxer

180º firing 2 cyl engine is per definition odd, as there is a spark at 0º, next at +180º, next at +180º=360º, next at +540º = 720º, cycle completed, sparks are not evently spaced (180º, 540º) that is called inline twin
2 cyl 360 parallel vs inline 180

If you put a wasted spark coil in a 180º firing 2 cyl engine you will have 2 wasted sparks, one at BDC end of admision (I would not recomend a spark with cylinder filled of mixture, even without compressing), the other one is not as worriying, as is BDC begining of exhaust, I don't know of any engine on this wasted spark configuration, but I don't know all cases.

While if you put a wasted spark coil in a 360º firing 2 cyl engine, you will have the 2 wasted sparks both at TDC end of exhaust (no flamable mixture), it's known to be commonly used, proved and tested, in fact is what happens in cylinders 1 and 4 (or 2 and 3) in a straight 4 cyl engine.

The tunes will brake for updating the firmware under many circumstances, may be, but who's updating the firmware for a perfectly running engine, should be aware of the setoffs vs improvements, and the ones that are updating the firmware because of bugs, are specting bugs like this to be resolved, so all shoul'd be aware and happy of the improved result of upgrading.

Thanks for reading

JuanNavarroS:
As it's already closed, see my comment here https://github.com/noisymime/speeduino/issues/740

Issue #673, opened 2021-10-03
Aux Inputs Not Working


Corey-Harding:
I am unable to get aux inputs working as expected using the latest ini file and compiling the latest firmware (1d3a30ab4c08b146dd92ea21c55c9f138f808be8).  I have also applied fix #655 and no difference.  I first noticed this issue when trying to test PR #586.   In regards to #654 the inputs as listed under programmable outputs still start at 1 instead of 0 when using the [current ini file](https://raw.githubusercontent.com/noisymime/speeduino/master/reference/speeduino.ini), this is fixed in my ini file but still does not work as expected.  I submitted my updated ini as PR #676 

There is one sequence I can run through to get the inputs to behave as expected which I will post below but the problem is back as soon as I power cycle and the input is unresponsive again.

A8 is hooked up to my ac switch and always works with idle up setting to increase and decrease idle but the two programmable outputs relying on a8 as an input to turn on compressor and fan only work when I follow the sequence below.

> 
> Then there is a series of events i can do to make it work briefly
> 
> I go to Secondary Serial IO Interface
> Enable Second Serial
> 
> Then go to local auxillary input configuration
> change aux input 0 to external source
> click burn and power cycle
> 
> then i go back to local auxillary input configuration
> change aux input 0 to analog local a8
> click burn but do not power cycle
> 
> during this period the input works as expected but as soon as i power cycle it is back to not working

Attaching images of my settings and ini file/relevant info

![adapter wiring tuner studio settings](https://user-images.githubusercontent.com/91717711/135763780-86bfd8d6-9700-420f-8a87-04871beb1ad9.jpg)

![adapter wiring cuts](https://user-images.githubusercontent.com/91717711/135764011-34df0e09-7070-4d49-acdb-cc722f573c57.jpg)

![adapter wiring schematic](https://user-images.githubusercontent.com/91717711/135764026-6154d614-8554-4f3c-9be2-4a6dda3b2ac8.jpg)

shiznit304:
This is the fork I currently run in my 2000 Miata, feel free to give it a try. My aux inputs work fine. https://github.com/shiznit304/speeduino/commits/mycar5 
It's the main branch from August 20, 2020 plus a few PRs such as #608 #549 #586 #607 #649 and one PR i haven't published yet for idle up with closed loop

Afroboltski:
[...]

PR #745, opened 2021-12-14
Change Aux in from 1-16 to 0-15 under programmable outputs

Corey-Harding:
This is the new #676 changing the branch name i had to make a new PR

Changes "Aux in" 1-16 to 0-15 under "programmable outputs" menu to match the pin configuration page found under "local auxillary input channel configuration" menu as suggested by @afroboltski


github-account-0:
This PR has worked good on my setup for a few months using it to drive the ac condensor, fan, and cel

Corey-Harding:
@pazi88 @Afroboltski @shiznit304 
Changing the branch name closed my old PR this is the new PR for Change Aux in from 1-16 to 0-15 under programmable outputs

noisymime:
Just to confirm, does this resolve #673 ?

Corey-Harding:
Yes although to do exactly what I wanted to I also had to also do this

https://github.com/Corey-Harding/speeduino/commit/e11864160506a487f099fc83d86c1fd4795423fd

And disable the pinisinput check

Corey-Harding:
Well in case you dont read comment edits and just use emails then I guess technically no this PR doesnt fix that issue entirely because one of the pinisoutput checks is causing me an issue for my use case but it fixes the ordering issue.

Im using prog outouts pin A8 to control my ac fan and the same pin for idle up so it fails unless i disable the check on that pin.

shiznit304:
I had to revert this PR to allow my Local Input Auxiliary Channel Configuration to work again. I believe the pins weren't initializing or something. All the pin values were 0

pazi88:
Are you using digital inputs? This PR should only affect those (fix bug that has been there for lomg time). Also the programmable output related naming is the only other thing changed.

shiznit304:
Yes I am using digital inputs. Although my firmware has several other pr.
I’d have to test with master alone to verify.

Corey-Harding:
You need to choose the new aux number that matches? Before your aux input
was 1 now it is 0, 2 is now 1, etc.

shiznit304:
Yes I tried adjusting my programmable outputs but that didn't work. All my
digital pins (I didn’t check if it affected analog) were stuck at 0.

Corey-Harding:
I was just going off what others had said they needed to do with the 63+1. I only use analog inputs. But the labeling did need to be changed to match for sure.

You changed the aux in 1 to aux in 0 though on the programmable outputs menu in far left side of tuner studio where you set up the conditions to match labeling?

Corey-Harding:
It matches the settings now in speeduino ino you can try 127+1 though.

I also think pina might need to be changed to 63 from 127.

shiznit304:
Yes the labeling matched, aux 0. My condition was checking if the aux input
is a 1 or 0. I also have hard coded some other things outside of
programmable outputs that check the pin status. But the pin never
initialized and stayed stuck at 0.

Corey-Harding:
I have to do this with my personal vehicle
https://github.com/Corey-Harding/speeduino/commit/e11864160506a487f099fc83d86c1fd4795423fd

shiznit304:
does #803 the issue caused by this PR?

Corey-Harding:
It does not. I think the issue you are facing is the same issue I had after this PR.

#398

PR #749, opened 2021-12-16
Update pin mappings for JuiceBox board.

joesbox:
Pin mappings updated for future injection enabled board. Doesn't break current hardware compatibility.


joesbox:
Removing this in lieu of adding a completely new board definition.

PR #752, opened 2021-12-23
can#.write func can exit without returning a value

dantob:
It doesn't look like speeduino actually checks whether can#.write actually succeeds anyway, and it already covers the channels that should be in use.


pazi88:
There is already new better library in pr #725, so I wouldn't bother changing anything in the old.

dantob:
fixed by #725

PR #754, opened 2021-12-30
2nd fuel table add ve analyze, 3d map and table generator

sixten9999:
adds ve analyze, 3d map and table generator to 2nd fuel map

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -1788,6 +1788,7 @@ menuDialog = main
    menuDialog = main
       menu = "3D &Tuning Maps"
         subMenu = veTable1Map,    "Fuel Table"
+	subMenu = fuel2Map,       "2nd Fuel Table"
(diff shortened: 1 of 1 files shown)


noisymime:
This looks like it has all the main pieces, but 1 thing I think needs to be looked at is having a filter or condition to indicate to TS which table is currently in use. There's no documentation I can find, but there's 2 places I can see to do this:

1. Use a regular filter. Eg `filter = std_Expression, "Table Inactive"    , { !(fuel2Active) }, false`
2. Use the `activeCondition` value on the `veAnalyzeMap` definition. I haven't played with this at all, but should be straightforward using `fuel2Active` I think?

The 1 other question this poses is what the 'correct' behaviour should be when a hybrid table is in use (IE both fuel tables are being used rather than switching from one to the other). Should VEAL only ever modify table 1 or are there conditions where the 2nd one should be used?

PR #756, opened 2022-01-02
Performance: optimize fuel trim PW calcs

adbancroft:
#### Summary

- Replace the use of a fixed divisor (100) when applying fuel trim correction to PW with a call to `div100()`.
- ~8% improvement in loops/sec

#### Detail
Also

- Simplified the optimized `div100` family of functions
- Refactored the fuel trim correction code into a shared function
- Slight increase in available RAM and decrease in flash :-)

Bright green is master, light green is PR:
PW libdivide 4G63 at 3700


HWright9:
Hey, If you have a look at #732 I also had a look at restructuring these fuel trims. The main reason is the correction also includes injector open time since the open time correction is done before fuel trim. This makes a 10% fuel trim add additional fuel compensation ammount depending on the injector open compensation time currently used. Also it would be best to seperate fuel delivery calcuations from injector timing calculations. Any thoughts about merging the structure I put together in #732 into this PR? Because it was a lot to have in one PR for me. I would rather #732 focused on dual O2 feedback only.

adbancroft:
> Any thoughts about merging the structure I put together in #732 into this PR? Because it was a lot to have in one PR for me. I would rather #732 focused on dual O2 feedback only.

This is a very limited PR, just focused on performance - I'd like to keep the narrow focus.

I agree that splitting up #732 would be a good path - better chance of acceptance. I think others are working on fuel delivery calcs? Have you asked on slack?

noisymime:
Any reason why you removed the ability to switch between division methods with `USE_LIBDIVIDE`?

 It was useful in doing back to back testing and debugging was all.

adbancroft:
> Any reason why you removed the ability to switch between division methods with `USE_LIBDIVIDE`?
> 
> It was useful in doing back to back testing and debugging was all.

It seemed unnecessary: now that we're using libdivide in a few places I'm thinking we're committed to libdivide; plus it wasn't applied consistently (the axis I/O code didn't apply it).

Just realized the last rebase from master included code that still uses it! (crankMaths.ino)

Let me know which direction to go in (`USE_LIBDIVIDE` or not) & I'll fix this PR up.

noisymime:
I think there is still value in including the `USE_LIBDIVIDE` macro check as it gives the flexibility to turn this on/off based on the platform in use. ALL my testing of `libdivide` so far has been its impact on the mega2560 platform, not any of the 32-bit ARM archs. There also isn't any on hardware CI/CD testing for those architectures, so if there ever was an issue there it will not be picked up. 

I think it's therefore safer to leave the 32-bit ARM platforms using the built-in divide instructions as they really don't need `libdivide` anyway. 

I haven't looked at it closely, but if `libdivide` does some internal checks for the architecture already, then perhaps it doesn't matter.

adbancroft:
`USE_LIBDIVIDE` is now back in place.

speeduino/maths.ino
```diff
@@ -23,74 +27,21 @@ int fastMap(unsigned long x, int in_min, int in_max, int out_min, int out_max)
...
-  else if( numerator <= UINT16_MAX ) { return FAST_DIV16U((uint16_t)numerator, 200); }
-  else { return libdivide::libdivide_u32_do(numerator, &libdiv_u32_200); }
-  
-  //return libdivide::libdivide_u32_do((y * x), &libdiv_u32_200);
+#ifdef USE_LIBDIVIDE    
+    return libdivide::libdivide_u32_do(y * x, &libdiv_u32_200);
 #else
-  return (y * x) / 200;
-#endif
+    return (y * x) / 100;
```
noisymime, line 43: Should this be 200 rather than 100?

adbancroft:
Argh! I did run the unit tests & they all passed (as did the build). 

But that was with `USE_LIBDIVIDE` in place, which is (and always has been) forced on in `math.h`. Forcing it off does cause a unit test failure. Fix uploaded.

This highlights the risks of having alternate code paths.

speeduino/maths.ino
adbancroft: Sharp eyes. Fixed

PR #762, opened 2022-01-08
Minor decoder related improvements

DeionSi:
Theoretically angleToTime and timeToAngle could calculate incorrectly for decoders that change triggerToothAngle.

The other two commits improve performance by ~0,3%.

### DIFF
--- a/speeduino/crankMaths.ino
+++ b/speeduino/crankMaths.ino
@@ -31,9 +31,10 @@ unsigned long angleToTime(int16_t angle, byte method)
         {
           noInterrupts();
           unsigned long toothTime = (toothLastToothTime - toothLastMinusOneToothTime);
+          uint16_t tempTriggerToothAngle = triggerToothAngle; // triggerToothAngle is set by interrupts
           interrupts();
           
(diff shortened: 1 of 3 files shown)


noisymime:
I'd need to do some testing to check the impact of this, but you're changing the value of the missing tooth end tooth calculation in some sequential instances. 

In the current code, line 622 in `decoders.ino` has:
`tempIgnition1EndTooth = (triggerActualTeeth + toothAdder);`

Where `toothAdder` is the total number of teeth that would be on the wheel if there were no missing ones. So in the case of a 36-1 wheel, `tempIgnition1EndTooth` would end up being 35 + 36 = 71

In your code you've replaced this with:
`tempIgnitionEndTooth = triggerActualTeethAdjusted;` 

Where `triggerActualTeethAdjusted` is 2x `triggerActualTeeth`. So in the above case of a 36-1 wheel `tempIgnition1EndTooth` would end up being 35 * 2 = 70

As mentioned, not sure of the impact as I'd have to test it, but seems questionable at first glance that it's giving different results to the current code.

PR #765, opened 2022-01-08
Fix full tune/large chucks of data for STM32. Next to this fix the co…

Tjeerdie:
Next to these changes the Arduino core needs to be adjusted. The receive buffer size for the CDC usb serial interface needs to be larger. 

In the Arduino core for STM32 change line 57 in cdc_queue.h to: #define CDC_RECEIVE_QUEUE_BUFFER_SIZE ((uint16_t)(CDC_QUEUE_MAX_PACKET_SIZE * 8))


Tjeerdie:
Another fix will solve it better/neater.

PR #781, opened 2022-01-26
New advance ignition calculations

DeionSi:
This is a combination of #702, #703 and #684 with additional code to support calculating in the old method.

Increases firmware size by 194 bytes. Performance for single table is down by ~0,25%. Performance for multiply (and probably added mode too) is up by ~5-3%. Tested on 0.4.3d board with Ardustim, seems to be working well.

The base tune must be updated to set "Corrected multiply/added mode" to "Yes".

### Changes

The advance calculations are rewritten and moved into one function for table 1 and 2 (adds getAdvance, removes calculateSecondarySpark). All advance related functions are moved from speeduino.ino and secondaryTables.ino to advance.cpp (except ignition corrections).

currentStatus.advance1 is only set when the secondary table is not in use, otherwise it is 0. currentStatus.advance2 is only set when the secondary table is in use, otherwise it is 0. currentStatus.advance is always the current advance.

The correctionsIgn function has been rewritten to not overflow at 127 degrees. Also it returns whether the resulting degrees are fixed or relative. (ref1)

Fixed advance corrections will now be honored for multiply and added secondary table. Previously this value would be used afterwards in calculations removing the fixed value. (ref2)

Multiplications above 127% no longer overflow to 0. All the way up to 215% will work. (ref3)

### Corrected multiply/added mode
The secondary spark table option ”Corrected multiply/added mode” refers to if the new mode (Yes) or old mode (No) is used. The only reason for old mode (No) is for the advance to not change on old tunes.

A new eeprom version is used to set ”Corrected multiply/added mode” to false to tunes utilizing the secondary spark table in added or multiply mode, otherwise it will be set to true.

- **Yes:** Corrections are only applied once.
  - **Multiply calculation:** ((table1+corrections)*(table2))/100
  - **Added calculation:** table1+corrections+table2

- **No:** Corrections are applied twice.
  - **Multiply calculation:** ((table1+corrections)*(table2+corrections))/100
  - **Added calculation:** table1+corrections+table2+corrections

  This option is intended to mimic the behavior of the multiply/added mode calculations before this pull request. However the fixes listed above also affect these settings (ref 1,2,3). The validity of this mode has been confirmed by comparing outputs of the tests below between master and test-groups 16 and 17. The values that differ are all due to the fixes (ref 1,2,3).

### Ignition advance tests

Test are included for the ignition advance calculations. The code comments explain the basic idea on how these tests work. The purpose of these tests is not to test individual corrections but rather the overall calculation.

All 1120 tests pass with this change.

The tests can be run on master by changing the references to configPage10.spark2correctedMultiplyAddedAdvance in test_advanceDC.cpp. The first should be commented and the second replaced with the rvalue of the commented line.


adbancroft:

PR #785, opened 2022-02-02
Coolant/RPM Protection

opelpanfan:
Coolant Temp based RPM limiter

image

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -1027,7 +1027,15 @@ page = 9
       boostByGear6               = scalar, U08,     161, { bitStringValue(boostByGearLabels ,  boostByGearEnabled) },   2.0,   0.0,  0.0,  {boostTableLimit},      0
 
       PWMFanDuty                 = array,  U08,     162, [4],       "%", 0.5,    0,    0,    100,      1
-      unused10_160               = array,  U08,     166, [26],       "",       1, 0, 0, 255, 0
+
+      coolantProtEnbl   = bits,   U08,     166, [0:0], "Off", "On"      
+      coolantProtRPM    = array,  U08,     167, [6], "RPM",     100.0,       0.0,   100.0, 25500,    0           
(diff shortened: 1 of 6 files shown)


Corey-Harding:
Why is the max F temp 215?

Is there a reason the min/max F temps for everything in the ini is capped the same as C? (-40, 215)

I have cars that run 210 when warmed up.  This would be more useful in my unit of measurement if I could increase these limits.  I see the line in the ini but wanted to ask before I go too far making changes.

opelpanfan:
[Corey-Harding](https://github.com/Corey-Harding) U08 max value is 255 so if we allow -40 to 215 then we get 255 in total.
Possible to change to U16 I think or make 0 to 255  in F range.

pazi88:
Internal calculation is in C so 215C is max, which is 419F. In there are IF -ELSEs for using C or F and the value limits just need to be set there differently if Fahrenheits are used.

opelpanfan:
Everywhere in TS.ini file for F used  "F",        1.8,    -22.23,    -40,    215,      0
So maybe need to be changed in all TS.ini file?

Corey-Harding:
I understand now where the 255 comes from 215+40. But in locations where F is commonly used -40 seems extreme.  Maybe a shift to -10 and 245 would be more useful?  I'm not the authority on the weather but having the max value a bit above operating temp seems more useful than the super low min value.

pazi88:
The limits for F are bit overlooked because as far as I know the guys mostly working on FW use C. But it's should just require tweaking the limits in the ini for more sensible. The Celcius side limits are just what the 1-byte variable allows with 40 offset. So the max values for F should be -40C and 215C converted to F.

Corey-Harding:
I didn't look at all of the temp settings in depth yet on the ini. I wanted to ask before I wasted a lot of time but I believe I will change mine.  Thanks for the all of the info.

PR #788, opened 2022-02-07
Fix building for STM32F407

pazi88:
Use EEPROM_LIB_H instead of <EEPROM.h>


noisymime:
I got lazy when I implemented this, the EEPROM stuff really should never have gone into `newComms` in the first place. I'll push a change shortly that performs all the EEPROM stuff from within `storage.c` and this will resolve the build issue for stm32f407

pazi88:
Fair enough😄

PR #790, opened 2022-02-08
angleToTime tests

DeionSi:
Testing for the angleToTime function. test_misc was running out of memory so I added test_misc2.

Also changes a (from what I can tell unused) tooth interval calculation to not be subject to integer rounding errors.

### DIFF
--- a/speeduino/crankMaths.h
+++ b/speeduino/crankMaths.h
@@ -19,13 +19,8 @@ unsigned long angleToTime(int16_t, byte);
 uint16_t timeToAngle(unsigned long, byte);
 void doCrankSpeedCalcs();
 
-volatile uint16_t timePerDegree;
-volatile uint16_t timePerDegreex16;
-volatile uint16_t degreesPeruSx2048;
-volatile unsigned long degreesPeruSx32768;
-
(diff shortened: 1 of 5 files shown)


noisymime:
These are good tests to add, but I'd really strongly suggest just having individual tests for each case. I get what you're doing with the looping through them, but for clarity both of making the tests purpose simple to read and being able to track it down if/when it does fail, why not just have each unit test as a separate function like normal? 

It has the other advantage of not requiring big calls to things like `snprintf`

DeionSi:
I made the test this way to reduce duplication/increase maintainability. It makes it very clear which parameters are being tested by being shown in a "table". I think the output accurately identifies which testdata is being tested.

Since these are just tests (where performance is almost irrelevant) I don't see why any heaviness of snprintf is an issue.

Nevertheless if splitting the code is a requirement for inclusion I will do so, please let me know.

noisymime:
I had a bit more of a look and a play with this and I think it's probably OK in its current form. What would definitely help is some comments that describe the functionality of each test and describe what a failure of the test would mean etc, but overall these work well enough.

PR #792, opened 2022-02-14
Better gear detection for A/T

rafolg:
Previous gear detection logic kept lagging 1 gear behind. This code has been tested  on an automatic transmission and works very well.

I have not tested on a manual transmission but should also work.


noisymime:
Closing this as I don't believe this is resolving an actual issue. Cause of this more likely appears to be hardware related (Eg TC slip)

PR #800, opened 2022-02-23
Change the newComms to wait for eeprom write to finish.

pazi88:
This small PR changes the the newComms behavior so that the `SERIAL_RC_OK` return code is only send out once all the data has been actually burnt to eeprom. Without this, large tune burns appear to take only short time, but actually the speedy is writing the tune to eeprom for some time after that. Which is ok, but if speeduino is turned off during this time, the tune will corrupt. The change in this PR makes the "sending tune" -dialog show in TS for the whole time when tune is being burnt. Removing risk of corrupt tunes.

Tested on Mega 2560 and STM32 with internal flash eeprom emulation.


noisymime:
Closing as this is no longer required after b7901568a01d098d58e6883e10206d4b9c765931

PR #803, opened 2022-02-28
Update Auxpina from 127 to 63

Corey-Harding:
In speeduino.ino comments regarding Auxpina/b says 127 is for dev test use so in sensors.ino I changed the values from 127 to 63.


Corey-Harding:
I tried to rebase it but I made it worse, its only two lines....  Will resubmit.

PR #804, opened 2022-02-28
Ignition  and Fuel schedulers

TBAMax:
Upgraded ignition and fuel schedulers.
This PR includes:
1. Get fuel and ignition schedules to work as originally intended: dwell timing and fuel injection pulsewidth directly applied to the timers. Apparently there were some computational hardships and problems with having all integer types correct, to have integer math on the timers work correctly (regarding need to take account that all math must work correctly with timer counter that rolls back to 0 periodcally). That is all honed out by now.
2. No change in loop speed on atmega2560 or only a very slight drop.
3. Noticeable dwell timing accuracy improvement, observeable with logic analyser. No observeable random extra ignition events visible during testing. Tested with all speed ranges from 60rpm to 16000rpm.
4. Smooth, no misses, no glitches, transition from cranking to running ignition dwell and timing tested with simulator and logic analyser.
5. Replaces #727 and #578 and #653 , related to #686 
6. New ignition mode also working and tested. With 60-2 trigger (harder case) seems to be working fine.
7. on Atmega3560 compared to master:133 Bytes less RAM used, 4324 bytes less flash used.
8. Updated ignition dwell limiter.
  update information 3.2023: Dwell limiter works also against possible race contitions when port register read-modify-write 
  type writes are done in some other parts of the code. Fixed possible race condition in dwell limiter itself when next ignition pulse falls within dwell limiter range, to prevent dwell limiter affecting the next ignition pulse at high RPM.
9. New Unity tests for the new schedulers (468 tests total for schedulers on atmega2560, more for boards that have more channels)
10. (update information 3.2023) Fuel injection duty cycle range extended to 100% (previously limited to 95%). This means continuous ON fuel flow from injectors at extreme cases.


pazi88:
Sounds promising 🙂

noisymime:
I've added comments for a few things I've spotted whilst reviewing this, mostly where something is fundamentally changed from the current functionality. 

Have tested through some of the most common setups and things generally seem to be OK. The resulting code is certainly cleaner, but there's quite a few areas where edge case checks have been removed, which makes me a little nervous. I'll need to do some more testing around these though

Here's a few more general comments anyway:

* The cleanup of the scheduler code is certainly overdue, but does come with some (Probably inevitable) downsides such as the increased time spent within ISRs and the additional context switching that will be required due to the extra function calls. The scheduler code in particular sees quite a number of function calls added to the ISRs, which is something that will need to be considered (Noting that the calls won't be inlined)
* Pretty much all the unit testing is broken in this PR as it stands at the moment. This will need to be fixed before pulling this in and obviously all the checks will need to be verified also. 
* I'm seeing some weirdness on the fuel pulses at high rpm/load (Beyond the duty cycle limiter) where pulses are either not starting or not completing correctly. Need to dig into this more though.

TBAMax:
@noisymime thanks for the comments. Today I managed to reorganize things so that I got rid of the small helper functions, and also got rid of the need for using the function pointers to the timer helper functions in the structs.
  The result is that there is less calls in the interrupt code (checked that from the disassembly listing). Also the code is even more clean now. And still can use the generic function for dealing with the schedulers. 
  The good thing is that now calls(previously) are inlined as much as possible, so I consider there is no increased time in the ISR from this PR at the moment. This all for today. Looked with logic analyser, all seems to be operating as expected at least on atmega. 
  Fuel duty cycle limiter behaviour still needs to be adressed and unit tests.

noisymime:
@TBAMax Re-reading my comment from the other day and I realised it probably came across a bit negative. Wasn't my intention as this is a really good PR. Just want to make sure it's not causing any issue when pulled in given it's size and how it's touching on so many critical areas.

TBAMax:
@noisymime no problem. I hope I did not ruin the PR with the latest addition of structs with virtual functions. I researched into it and it still seems good solution in my eyes. Specially for the solution against having function calls that can not be inlined. With having virtual functions in the main struct and then overriding them in the derived child struct(class) allows those functions to be inlined, and it is being confirmed by looking at reduced amount of instructions (and icalls) in the dissassembly listing. So functionally and reduced codewise it looks good, but it is essentially not so conservative C coding, but rather C++.

adbancroft:
@TBAMax I may have a way to remove the virtual function calls. Is this PR still being actively worked on?

@noisymime Is this PR likely to be accepted?

TBAMax:
> @TBAMax I may have a way to remove the virtual function calls. Is this PR still being actively worked on?

@adbancroft Bring on the info about alternative ways! Currently this PR is included to the https://github.com/TBAMax/speeduino/tree/mainbranch (with some other stuff) for testing (some lag of maintaining mergeability here). And I have been using it on the car to get the feel how it is behaving. Although some recent changes about the full/semi sequential switching not included there also. It was little hard to understand why some of the scheduler swithing was moved to the init.ino recently, and I need to chew that logic out yet, to bring this up to date.

adbancroft:
@TBAMax How would I bench test changes using ArduStim? Is there anything in TunerStudio to log?

adbancroft:
@TBAMax I'm working on a patch for you. Question: the new fuel scheduler adds a new check when setting the next*Compare members: there has to be 400 timer ticks remaining in the current injection pulse. Why is this needed?

TBAMax:
> @TBAMax I'm working on a patch for you. Question: the new fuel scheduler adds a new check when setting the next*Compare members: there has to be 400 timer ticks remaining in the current injection pulse. Why is this needed?
@adbancroft This was needed to make 100%duty cycle possible, to cater every possible case.
 Indeed the hardcoded constant is not nice. Maybe it can be calculated as following:
`if ((timeout < MAX_TIMER_PERIOD) && ((COMPARE_TYPE)(targetSchedule->endCompare-targetSchedule->getFuelCounter())>(uS_TO_TIMER_COMPARE(INJECTION_REFRESH_TRESHOLD)))`
need to give it a thought, and then test everything again...
When injection pulses overlap (injector always "ON" condition), then fuelchedule status is always RUNNING so all schedule setting in the main loop is actually done via nextSchedule. So the nextschedule update can not also be too close to the interrupt. Maybe it can be done simpler also. I was observing the duty cycle limit. It was previously set capped to 95% because schedulers could not take 100% and would have quite random behaviour at more than 95% (because of fluctuation of 'timePerDegree' causing overlap sometimes). So this was what I came up for now.

noisymime:
Most of the questions I posted above look like they still stand, even after the changes. Some look like they might just be copy/paste type things, but others might be a bit more fundamental.

TBAMax:
Just a note here that recently  merged  changes from #876 probably affect this too, because ignition or fuel timer macros are changed there. So when rewriting this at some future point, need to look at those changes there.
Currently I still consider the ignition and fuel schedulers presented here to be superior to those of the current speeduino/master. But it is not possible to maintain such a big change reliably as a mergeable PR, as it is practically maintaining a separate fork. So fresh rewrite starting from the master is needed for this, to avoid accidentally reversing some bugfixes that have been done meanwhile.

adbancroft:
@TBAMax Teemo - I agree with this approach to the timers and the need to recode against latest master. What can I do to help?

A different branching strategy could make maintaining this easier: in your separate branch in your repo, *rebase* from master regularly. This keeps the both commit histories easily visible, rather than bulk merges that contain multiple commits under one "Merge branch 'master' into IgnAndFuel" comment. It may complicate in other ways though.

TBAMax:
> @TBAMax Teemo - I agree with this approach to the timers and the need to recode against latest master. What can I do to help?
> 
> A different branching strategy could make maintaining this easier: in your separate branch in your repo, _rebase_ from master regularly. This keeps the both commit histories easily visible, rather than bulk merges that contain multiple commits under one "Merge branch 'master' into IgnAndFuel" comment. It may complicate in other ways though.

@adbancroft  Some points with that was that it breaks the automated (unit)tests for the schedulers. I was uncertain how to deal with that. Either make new tests, or try to modify the code so that it can be squeezed to satisfy the existing testing pattern..., or combined both.
Other issues I consider pretty much fixed (irregular pulses beyond the duty cycle limiter, function calls in the ISR inlining).

TBAMax:
Overdwell protection currently under inspection here.
Objectives: 
1.Gather all information about current status of the overdwell protection and its working mechanisms.
2.If missing pieces or places of possible improvement found, make nessesary changes.
3.Update documentation of the PR about the topic.

noisymime:
Please do look at the other comments I made in the review above. These may have been addressed already, but there were a few things there that looked questionable.

speeduino/speeduino.ino
```diff
@@ -890,399 +844,134 @@ void loop(void)
...
-          //*********
-
-          refreshIgnitionSchedule1( uSToEnd + fixedCrankingOverride );
-        }
-  #endif
         
 #if IGN_CHANNELS >= 2
-        if (maxIgnOutputs >= 2)
+        if (maxIgnOutputs >= 2 && !BIT_CHECK(curRollingCut, IGN2_CMD_BIT))
         {
```
TBAMax, line 930: Something is strange with the rolling cut here! Why the master do not have it any more?

speeduino/speeduino.ino
```diff
@@ -1424,7 +1111,7 @@ byte getAdvance1(void)
   else if(configPage2.ignAlgorithm == LOAD_SOURCE_TPS)
   {
     //Alpha-N
-    currentStatus.ignLoad = currentStatus.TPS * 2;
+    currentStatus.ignLoad = currentStatus.TPS;
```
TBAMax, line 1115: Seems like accidental reversal of totally unrelated thing! How can this happen?

speeduino/speeduino.ino
TBAMax: Ok, checked that: change view picks that up from different place. In reality there is no issue. Rolling cut is still in place. So all ok with that.

speeduino/speeduino.ino
TBAMax: Fixed that in the nex commit.

speeduino/scheduler.h
```diff
@@ -44,247 +33,247 @@ See page 136 of the processors datasheet: http://www.atmel.com/Images/doc2549.pd
...
 
-inline void refreshIgnitionSchedule1(unsigned long timeToEnd) __attribute__((always_inline));
+void setFuelSchedule (struct Schedule *targetSchedule, unsigned long duration);
+void setFuelSchedule (struct Schedule *targetSchedule, int16_t crankAngle, int16_t injectorEndAngle, unsigned long duration);
 
-//The ARM cores use separate functions for their ISRs
+
+//The ARM cores use seprate functions for their ISRs
 #if defined(ARDUINO_ARCH_STM32) || defined(CORE_TEENSY)
-  static inline void fuelSchedule1Interrupt(void);
```
DeionSi, line 113: The argument should probably be void on these. Some MISRA rule I believe.

TBAMax: Yeah, fixed in the next commit here.

DeionSi:
The tests are no longer passing. It seems they need to be updated.

TBAMax:
> The tests are no longer passing. It seems they need to be updated.

Is there a technical reference of how to run those tests?

DeionSi:
> Is there a technical reference of how to run those tests?

https://docs.platformio.org/en/stable/advanced/unit-testing/index.html

In vscode platformio
image

DeionSi:
Have a look at @adbancroft 's [PR](https://github.com/TBAMax/speeduino/pull/11). It seems to have a lot of testing stuff in it.

TBAMax:
Update

@DeionSi @adbancroft  I have now upgraded the tests here , and managed to run them on atmega physically also. I must say this is quite fun! Took the more advanced route and included the crankmath into the test, so the schedules can be tested at various RPMs.

@noisymime, when looking over the checklist now : 
- The function calls in the interrupts are minimized. The virtual functions are here to stay. I looked the alternative of using function pointers or references to the `setCompare(x) `and  `getTtimer()`, this was not good, and resulted in indirect function calls from the interrupt routine vs. with virtual function style as now all is inlined.( `avr-objdump -d -S firmware.elf >firmware.lst`)
  So all that is left from the function calls inside the interrupt routine is `startFunction()` and `endfunction()` (indirect calls) but nothing to do about those because we need to be able to reconfigure what those point to. 
- Dwell limiter is now  fixed and reintroduced. Though there is possibility to add corresponding unity test for it in the future. Current present day status of this is that I have not tested it on the real hardware.
- Fuel schedules performance should be resolved with the allowing of the PW up to 100%. So that it do not have any more weirdness at near or even over 100% duty cycle. I remember doing extensive testing on those with the real board and Ardustim, so it should be fine now.

So thats all for today. Thanks for watching! Remember to click "Like" and "Subscribe"!

adbancroft:
@TBAMax 
> The virtual functions are here to stay. I looked the alternative of using function pointers

I ran a quick test and the references/function pointers solution saved 54 bytes and was 2 to 3% faster loop/sec (Ardustim, 60-2 missing tooth, multiple combinations of ignition configurations). I submitted a PR you can test.

Also, I'm seeing some ignition drop outs:
image

TBAMax:
> I ran a quick test and the references/function pointers solution saved 54 bytes and was 2 to 3% faster loop/sec (Ardustim, 60-2 missing tooth, multiple combinations of ignition configurations). I submitted a PR you can test.

Ok, the 54 bytes itself is not critical. But 2 to 3% loops/sec? it should be not noticeable. Possible that it comes from the counter and compare used as pointers to register. I have to check that out, seems good optimization! Timer start and timer stop are still funcitons and so make function calls, but those are not actually used regularily when engine is running.
> 
> Also, I'm seeing some ignition drop outs:

The drop outs are more serious issue to address. What is the configuration of ignition shown in the picture? It seems coils 1,3,4 firing simultaneously and only 2 is shifted? Drop outs may be from the new dwell limiter...

TBAMax:
> 60-2 wheel, rising edges, wasted spark, old ignition mode, Ardustim at 4K RPM.

When you turn off the dwell limiter do you see the dropouts dissapear?

TBAMax:
> Also, I'm seeing some ignition drop outs:

It is a dwell limiter running into the next impulse. This is possible due to he following:
This is due to the nested interrupts, the scheduler interrupt happening inside the timer2 interrupt because the timer2 interrupt is with the argument ISR_NOBLOCK
So the scheduler interrupt happens right in the middle of the dwell limiter. I originally missed to notice that there is nested interrupts on atmega...
When the scheduler interrupt happens right in the middle of dwell limiter, between the limiter comparison and limiter activation(call to the EndFunction), the new pulse gets disabled immediately by the limiter. So kind of a race condition. @adbancroft good that you observed carefully and noticed!
So now this is a known issue. Only to figure out now how to resolve this.

TBAMax:
I have pushed the fix for the dwell limiter. Feel free to check out!

TBAMax:
Oh crap it did not work on stm32. Because cli() and sei() is not applicable there...
So new version.

VitorBoss:
> Oh crap it did not work on stm32. Because cli() and sei() is not applicable there... So new version.

The helper functions have their relative substitutes, if it didn't work isn't because of that.
Keep in mind ARM cores have 15 levels of interrupt nesting.

DeionSi:
Some memory and flash savings in https://github.com/TBAMax/speeduino/pull/13

TBAMax:
> I ran a quick test and the references/function pointers solution saved 54 bytes and was 2 to 3% faster loop/sec (Ardustim, 60-2 missing tooth, multiple combinations of ignition configurations). I submitted a PR you can test.

Just a note here that I have analysed the references/function pointers solution. The 54 bytes are flash not RAM. In the interrupts the references/function pointers version uses about two atmega instruction cycles more than virtual functions version, so practically same. But I did not rush to change yet, because before version is well tested and I do not have the recources at the moment to retest extensively. So I feel keeping this in mind, and when needed, do the change in the future. As for the 2% to 3% loops/sec I do not care at the moment, seems most likely in the range of possible placebo effect, as I can not really see looking at the generated asm where it would come from. Also as for the loops/s I already have some more possibly powerful ideas how to boost that (most significantly by reducing the unnessesary schedule recalculations/refreshes). But I will not add that also at the moment because it is again more of a topic for separate pull reguest to be done after (wanting to push that to the next more extensive testing session).

DeionSi:
@TBAMax Please see my reviews above from last month.

TBAMax:
> @TBAMax Please see my reviews above from last month.

@DeionSi do you mean the "Some memory and flash savings in https://github.com/TBAMax/speeduino/pull/13"  (channelDegrees for injector schedules)? That one I did pull in, and is included at the moment. It was all logical step.

DeionSi:
> > @TBAMax Please see my reviews above from last month.
> 
> @DeionSi do you mean the "Some memory and flash savings in [TBAMax#13](https://github.com/TBAMax/speeduino/pull/13)" (channelDegrees for injector schedules)? That one I did pull in, and is included at the moment. It was all logical step.

@TBAMax I can't find a way to link to those reviews but they are right after https://github.com/noisymime/speeduino/pull/804#issuecomment-1416138142

TBAMax:
It's been quite a lot of time again. I have done some merge commits here to keep this level with the speeduino:master.
So that there should be no major issues with testing it without changing the speeduino.ini. Should be able to test this without any configuration change -- just compile this and upload to the board, and it should work without any more changes compared to the speeduino:master. 
  Only thing is when you do not change the speeduino.ini then you can not set the fuel duty cycle to the 100% and test that. Without switching the `speedino.ini` it allows only up to 95% (same as master).
  
  Recent 
 
> Revert "Merge branch 'master' into IgnAndFuel"
Was simply because I decided to completely not include the recent test changes for the old schedulers on the master. Just for clarificatioin here.

TBAMax:
Trying to find time to build up simulator again for the Atmega and STM32 to test this after this long period of just merging the changes from master. To see if this all still works as expected. No intention to build Teensy test rig because I still do not have any Teensy board. There have been notice by @noisymime that this do not work on Teensy at the moment for some odd reason and the Teensy board just hangs on startup, but I have got the impression that it is a somewhat not related to the topic of this PR.

TBAMax:
Possibly the reported problems on Teensy 4.1 when the board hangs on startup, are related to nested(or not nested) interrupts.
Currently seems that using nointerrupts() and interrupts() in overdwell protection in timers.ino might be the culprit. Because it is unclear how the teensy nested interrupt controller reacts to that.

noisymime:
Current plan is for this to be one of (if not the) first changes merged in the next dev cycle. That will give a long cycle for testing of this given how large the change is to such a significant piece of the code.

adbancroft:
I'm seeing a 5 to 7% slowdown versus the current master? 

Also, some unit tests don't compile & some fail.

TBAMax:
> I'm seeing a 5 to 7% slowdown versus the current master?
> 
> Also, some unit tests don't compile & some fail.

@adbancroft thanks for the review! I fired up the tests, and indeed there were some fails. Some injection schedule tests failed. Turns out the fails were caused by partly by recently changing the `INJECTION_REFRESH_TRESHOLD`, and also for some tests `CRANK_ANGLE_MAX_INJ` was undefined. `INJECTION_REFRESH_TRESHOLD` was not accounted for in the tests and this caused the fails.

`INJECTION_REFRESH_TRESHOLD` I found by testing it is better to have larger than it was, because of possiblity of the scheduler jumping over the cycle when there are fast changes in RPM. And so may cause some fuel cycles to be skipped. (This probably applies to the current master also, not only in this PR). Anyway I included this change in at the moment.

Can not say anything about the claimed slowdown at the moment. Assuming it is on Atmega2560 that you are observing. Only thing that may cause some extra computational load, that I can think of at the moment, is addition of up to 100% fuel duty cycle capability and associated calculations.

adbancroft:
> Assuming it is on Atmega2560 that you are observing.

Yes, I'm running on an Mega 2560: missing tooth, 60-2, crank primary trigger @ 4000 RPM on Ardustim. Consistent slow down across wasted & sequential ignition, with per-tooth timing on or off.

TBAMax:
> > Assuming it is on Atmega2560 that you are observing.
> 
> Yes, I'm running on an Mega 2560: missing tooth, 60-2, crank primary trigger @ 4000 RPM on Ardustim. Consistent slow down across wasted & sequential ignition, with per-tooth timing on or off.

Would You mind including what are using to measure the slowdown? I mean probably TS loops/s? And what exact values are You getting?

TBAMax:
I am tempted to do some optimizations: Particularily modify  `triggerSetEndTeeth_missingTooth()` or `triggerSetEndTeeth` in general, to only calculate new tooth angles in case the ignition advance has changed. This should speed up the loop speed nicely. Of course this actually is not correctly in the scope of this PR.

adbancroft:
> Would You mind including what are using to measure the slowdown? I mean probably TS loops/s? And what exact values are You getting?

Yes, I'm comparing TS loop/s. Here's what I have:
```
Spark Setup	Per Tooth Timing	Master 2023-06-11	TBAMax	Delta
Sequential	Yes			900			835	-7%
Sequential	No			1040			970	-7%
Wasted		Yes			980			925	-6%
Wasted		No			1155			1095	-5%
```

adbancroft:
> Particularily modify triggerSetEndTeeth_missingTooth() or triggerSetEndTeeth in general, to only calculate new tooth angles in case the ignition advance has changed.

FYI, there's an piece of commented out code doing exactly that: speeduino.ino, line 956 on master.

TBAMax:
> Yes, I'm comparing TS loop/s. Here's what I have:

@adbancroft Thanks for the data. I have now verified. Indeed there is the slowdown of the described amount. I have also determined the cause: The cause is how often the `angleToTime()` function is used in the setIgnitionShedule. It is needlessly calculated for some cases. I also already have determined the possible fix, but I want to run a logic analyser test first, it needs some time to be set up again.

TBAMax:
I tested now with the logic analyser that all works as expected (on atmega2560), and pushed some upgrades to get better loop speed. Now it should fall within about 2% compared to the current moment master regarding to loops/s.

TBAMax:
Apparently the updates on the master gone different route. Since this branch is well tested and working well, and functionally superior to the current master, I am keeping this updated, so that it is functionally compatible with the master. At least until changes regarding schedulers on the speeduino/master are reaching at least seemingly finished state, and can be verified against this PR (branch) and then decide if I keep maintaining this further or not.

TBAMax:
I did a short code rewiew at the moment to determine if some of the issues that this PR adressed still stand on the master at the moment. 
The findings:
  Seems like the main motivation that was behind this PR still stands on speeduino/master: To improve ignition timing accuracy  by not calculating the ignition moment via the "startAngle". 

  Situation is that ignition moment is calculated using the following pathway:

- [ ]  Ignition angle(from the table) -> 
- [ ] IgnitionStartAngle(angle domain math). Using engine speed, and round down to the degree accuracy, those are main sources of introduced extra inaccuracies: fluctuating speed reading and subsequent rounding down. ->
- [ ] calculate ignition moment(time domain) from IgnitionStartAngle(angle domain) (using engine speed, and current position, this is second time engine speed is used in the calculations, so any error it has is doubled!)

 Better way to calculate ignition moment (introduced in this PR) is:

- [ ] Ignition angle(from the table) ->
- [ ] Ignition time aka ignition moment(time domain). Calculated using engine speed and engine current position: least amount of computational inaccuracies!
- [ ] back calculate ignition start time(time domain). Using ignition time and dwell.

Same basically applies to the fuel injection timing.
This was the main motivation behind this PR and seems the work analysing and testing done still stands and may have some value to it.

PR #806, opened 2022-03-01
PWM and on/off idle fixes/cleanup

pazi88:
This PR consists some smaller fixes and cleaning up of the PWM and on/off idle control. This will not affect normal operation, but fixes multiple problems in edge cases that I have seen. All 4 modes that are affect by this PR has been tested to work normally.

What has been done:
- Added missing "idle (control) on" -indicator that has been visible in logs, but not in TS
- Unnecessary `currentStatus.idleDuty` and data shuffling related to that has been removed
- Removed unnecessary/duplicate code
- Added proper 100% and 0% duty checks, that take account the idle valve direction and number of idle outputs
- Use same consistent way of setting idle pins high and low.
- Working of the on/off idle output now shows up in "IAC value" in TS. (0 or 100% duty)
- Added "Run before crank" -code for PWM CL+OL -mode, because it was missing from it.


pazi88:
The cranking duty cycle worked before this change and works with this too. The Cranking duty cycle curve is used when RPM is below the "Cranking RPM (Max)(rpm)" -setting. Please note that you need to have sync, before seeing RPM, so during the time that engine actually cranking but there is not yet sync, the idle valve doesn't do anything. To get it running in that phase too, you need to enable the "Run before start" -setting. This wasn't working for "CL+OL" -mode before this PR.

transi01:
@pazi88 : Have includes your changes to my source and using Mega 2560 it is working as it should :-) Well done!

But maybe you have an idea: If I use the code on a teensy 3.5 only Open Loop and Closed loop Idle is working. If I switch over too "CL+OL" it doesn't work... Engine starts and dies immediatly. This problem only persists with the Teensy 3.5 as the Processor. using Mega 2560 everything works fine.
Same behavior without your changes, so it is not resulted cause of your changes :-)

Any idea?

// Thomas

PR #816, opened 2022-03-16
fix typos in code documentation

dantob:


Corey-Harding:
Just an FYI, InitialiSE is the ENGLISH spelling and InitialiZE is the AMERICAN spelling.

digmorepaka:
Just went through the whole thing. 

Right off the bat line 254 corrections.ino enrichments is used as plural so no 's.

There is many words besides Initialize where we need to define what variant of English should the project be in. US English, UK English, etc. Here's the list of what I caught:


US - Initialize
UK - Initiliase

US - fueling 
UK - fuelling

US - behavior  
UK - behaviour

US - targeted
UK - targetted

US - specialized
UK - specialised

US - prioritize
UK - prioritise

US - optimizations
UK - optimisations

PR #820, opened 2022-04-02
Increase dwell limiter in RX-8 type ignition

theKraid:
Currently the dwell limiter calculation considers rotary ignition to have more than 1 pulse per revolution. This only holds true for FC/FD type ignition which is wasted spark correctly resulting in 2 pulses per revolution. RX-8 type ignition however is fully sequential so there is only 1 pulse per revolution. As a result the dwell limiter is being calculated too low.
This change fixes the issue by correctly assuming 1 pulse per revolution in RX-8 type ignition


theKraid:
I personally ran into this issue with my rotary engine by using IGN-1A coils in RX-8 type configuration. These coils use high dwell times (3.0+ ms) and high spark duration (2.9ms) which resulted in the dwell limiter incorrectly engaging early leading to misfires.

Table

By applying this fix the dwell remains stable

Table

PR #825, opened 2022-04-13
Added option to fix tacho signal to coil dwell

pazi88:
In 201903 release there was addition for configurable tacho pulse duration to prevent issues where long dwell times in higher cyl engines caused problems like missing pulses etc. Unfortunately the new method uses 1ms interval in the code for the tacho pulse, causing the tacho pulses to be in 1ms grid. This causes significant accuracy loss for the tacho pulses and lot of jitter in the pulses. 
https://user-images.githubusercontent.com/48950874/163122189-b382abcb-6db0-44e8-8df8-9362f82685c4.mp4

This might or might not be problem depending on the used tachometer, but in worst case the tacho needle can have very jittery behavior.  This PR adds option to use the old method of using coil dwell for tacho pulses that was in use before 201903 release.


sbtoonz:
Finally my RPM no longer wobbles around 4.5-5k :D

sbtoonz (approved):
Great success on current master

Corey-Harding:
I have noticed pretty often when downshifting and engine braking that my tach will drop to zero, do you suspect that this PR will have any impact on that?

PR #831, opened 2022-04-25
Fix for EGO correcting in wrong direction in PID mode and not allowing for correction at WOT. 

mbbrinkman:
Two very simple fixes for pretty important things. 

Currently, AFRvalue (perhaps paradoxically) increases fuel as it increases. The PID loop, which is updating AFRvalue with its output, should then be direct as opposed to reversed.

Secondly, you cannot set your TPS upper bound above 100%, yet currently EGO correction requires it to be explicitly below that bound to function. So if you're exactly at 100% such as in WOT, it doesn't allow you to run correction. Making the < a <= fixes that. 

This is tested and fixes the problems I had with PID EGO correction on my car.

### DIFF
--- a/speeduino/corrections.ino
+++ b/speeduino/corrections.ino
@@ -36,7 +36,7 @@ long PID_O2, PID_output, PID_AFRTarget;
 * Needs to be global as it maintains state outside of each function call.
 * Comes from Arduino (?) PID library.
 */
-PID egoPID(&PID_O2, &PID_output, &PID_AFRTarget, configPage6.egoKP, configPage6.egoKI, configPage6.egoKD, REVERSE);
+PID egoPID(&PID_O2, &PID_output, &PID_AFRTarget, configPage6.egoKP, configPage6.egoKI, configPage6.egoKD, DIRECT);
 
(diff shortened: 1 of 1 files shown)


noisymime:
The O2 PID works off the AFR value, which increases when you decrease the fuel. `DIRECT` on the ego PID would mean that increasing fuel (`PID_output`) would lead to an increase in the AFR value, which is not the case. Have you actually tested this in isolation to see if it's responding the way you want it to?

For the TPS fix though, this is 100% correct. If you submit it separately I'll pull it in.

mbbrinkman:
I'll submit it separately. 

There may be something wrong in my setup because it is working inverted only some of the time now that I'm messing around with it.  I'll look at it more before submitting more on this.  I tried with direct and reversed after you mentioned this and sometimes it worked as expected. 

Perhaps I am misreading or misunderstanding what AFRvalue does in the code.  I didnt give it a through look but in investigating why ego correction is (often) reverse of what is needed I noticed that you commented that in simple mode, an increment of AFRvalue by 1 increases fuel by 1%.

Issue #829, opened 2022-04-21
Warning in TunerStudio after [WIP] CAN broadcasting menu (#639) 


Corey-Harding:
```

1 Warnings:
Warning: Failed to set value for unusedClusterBits .msq value is not valid for current configuration: No valid options found for Bit EcuParameter:unusedClusterBits equal to the proposed 0

```

Corey-Harding:
From my msq tune file
```

<constant name="canBMWCluster">"On"</constant>
<constant name="canVAGCluster">"On"</constant>
<constant name="enable1Cluster">"On"</constant>
<constant name="enable2Cluster">"On"</constant>
<constant name="unusedClusterBits">0</constant>
```

I flashed firmware before this commit
[Remove duplicate updates from last PR](https://github.com/noisymime/speeduino/commit/97f8ef795a514c9a5693e85be6cdb8bdee8ef2c5)

So I assume I need to turn the first 4 values to off?

floods57:
Hi, 
Thanks a lot, that worked for me ;)

PR #844, opened 2022-05-14
Fix Issue #829 Warning in TunerStudio after [WIP] CAN broadcasting me…

Corey-Harding:
One potential fix for Issue #829 Warning in TunerStudio after [WIP] CAN broadcasting menu (#639)


noisymime:
I've fixed this in 9673dd8fe62b423a602f1b0ffd90382cd29cb30f

Rather than removing all the options (Which is technically invalid per the TS spec), I've simply added a valid one ('Off').

Corey-Harding:
Thank you for addressing this.

PR #845, opened 2022-05-15
Moved WMI to VVT2

VitorBoss:
This add the possibility to use single channel VVT and WMI at the same time.

WMI to VVT2 channel and increased PWM resolution to 0.5%.


VitorBoss:
This need to be tested as I can't test ATM.

jackscr12:
Will this work on the mega cpu? I thought the reason for not having VVt and wmi at the same time was because of limited PWM outputs on the mega board.

VitorBoss:
The intent for this PR is to move the WMI implementation to VVT2 channel to be able to enable both on Mega2560!
Timers are a bottleneck indeed, but I made a work around for that, both VVT channels runs on the same timer, the control is handled in code.

The limitation will be the valve frequency, both will be the same.

tuxit09:
> This need to be tested as I can't test ATM.
 If there is a test firmware file for this I'm 100% down to try it.  My particular setup is single VVT1 with WMI.

VitorBoss:
> If there is a test firmware file for this I'm 100% down to try it. My particular setup is single VVT1 with WMI.

Yes, you can go to my PR branch called "WMI-to-VVT2" compile the code and use it.
Just updated to latest release version

tuxit09:
> > If there is a test firmware file for this I'm 100% down to try it. My particular setup is single VVT1 with WMI.
> 
> Yes, you can go to my PR branch called "WMI-to-VVT2" compile the code and use it. Just updated to latest release version

So I was able to get the branch compiled and uploaded to my board. However with the new firmware I get a message when configuring VVT/WMI in tunerstudio that they share a PWM and the 2 cannot be used simultaneously. So while, I am under the impression that you moved WMI to VVT2 function, I cannot enable it in TS.  Any other advise for getting this running on my end?

VitorBoss:
Just pushed the change to be able to do it again

tuxit09:
> Just pushed the change to be able to do it again

Thanks for going to such lengths. I am still having an issue preventing testing of VVT + WMI. from the original problem of the selection for the VVT/WMI output pin define is linked together in TS. so whatever pin is selected for one, the other inherits immediately. VVT and WMI still can not have different output pins defined.

VitorBoss:
Actually it is moved to VVT2 pin

tuxit09:
> Actually it is moved to VVT2 pin

little confused I am. I have WMI on pin5 and VVT on pin 11 physically. Are you saying I set VVT1 pin for 11 and VVT2 pin for 5 and then set the VVT2 table for map-kpa and rpm for WMI functionality from the VVT2 settings/table?

VitorBoss:
The VVT1 will function as normal, the VVT2 will behave like the WMI behave now, the only change is the output pin

tuxit09:
> The VVT1 will function as normal, the VVT2 will behave like the WMI behave now, the only change is the output pin

OK. I tried that. I put the DutyCycle for VVT1 and VVT2 on display in TS dashboard. While I see the duty cycle activated in the dashboard for both VVT1 and VVT2, I am not seeing an activation of the WMI (VVT2) on the output pin (not sure if VVT1 is activated or not). I will verify with certainty tonight and report back. I have an output feedback led on the WMI pump circuit, just not seeing an activation anymore since moving to VVT2.

Just curious, Why not just do an IF WMI and VVT1 are enabled, THEN disable VVT2 settings dialog and use WMI settings for the actual WMI dialog? I don't mind doing it any way that works, just seemed more user intuitive to me.

tuxit09:
> The VVT1 will function as normal, the VVT2 will behave like the WMI behave now, the only change is the output pin

I'm thinking that this could be the limitation and why WMI on VVT output doesn't work for me. 

> Note: Currently Closed loop VVT control for second VVT output is only available for missing tooth trigger pattern with single tooth on the cam.

Since I do not currently use a camshaft position sensor, does this break the functionality of WMI on VVT2 output?

tuxit09:
> The VVT1 will function as normal, the VVT2 will behave like the WMI behave now, the only change is the output pin

Unfortunately I cannot seem to get WMI activation from VVT2 enablement. It shows duty cycle in TS, but VVT2 pin (assigned pin 5) is not going low on activation. 

In addition there is the other issues of other WMI settings being disabled like timing advance on WMI activation. 

Only thing I can say is VVT1 works just fine for me. How else can I help?

tuxit09:
Just wanted to update this. WMI + VVT1 are working simultaneous with this branch. Please try to push to mainstream. I am happy to provide whatever data you need.  WMI table is non-persistent as-is and must send from TS on connect.

tuxit09:
Any momentum on this support request? The patch works great except the WMI table data not being persistent.  I would love to be able to stay current on future speeduino releases.  Can this be rolled in?

VitorBoss:
Yes, this is the queue to be reviewed for the next release, you can follow here: https://github.com/users/noisymime/projects/1

PR #846, opened 2022-05-16
Free up unused function pins for programmable outputs

aaronjonesxyz:
Add function enabled checks to pinIsOutput macro for VVT, Fan and idle, to allow the pins to be used as programmable outputs.


shiznit304:
Looks a lot like #549

aaronjonesxyz:
> Looks a lot like #549

I forgot that existed, still isn't merged into master

DeionSi:
It seems that the code has changed a bit here. Is it possible to get this synced with master?

VitorBoss:
> It seems that the code has changed a bit here. Is it possible to get this synced with master?

Master already have something in place for that, I guess @noisymime forgot to close this PR

PR #852, opened 2022-05-24
Fix possible pin conflicts

pazi88:
This PR mostly affects possible pin conflicts on STM32, but should be good addition on Teensy too.

- Add SD-card SPI-pins to STM32 `pinIsReserved` -list (if used)
- Use `pinIsUsed` -list instead of `pinIsOutput` -list for selectable inputs. (this adds the `pinIsReserved` and mandatory analog sensor pins for the list of pins that can't be selected)
- For features that have selectable output pins, only set as output if needed and check if pin is in the `pinIsReserved` -list

### DIFF
--- a/speeduino/board_stm32_official.h
+++ b/speeduino/board_stm32_official.h
@@ -78,9 +78,17 @@ extern "C" char* sbrk(int incr);
   #endif
 #else
   #ifdef USE_SPI_EEPROM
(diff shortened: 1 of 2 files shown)


DeionSi:
I found the reason for the Injector failures after updating to master, #979

The following failures remain:
test\test_misc\tests_misc.cpp:56: test_initialisation_outputs_V03: Expected 1 Was 0. Fan        [FAILED]
test\test_misc\tests_misc.cpp:114: test_initialisation_outputs_MX5_8995: Expected 1 Was 0. Fan  [FAILED]
test\test_misc\tests_misc.cpp:146: test_initialisation_outputs_boost: Expected 1 Was 0. Boost   [FAILED]
test\test_misc\tests_misc.cpp:151: test_initialisation_outputs_VVT: Expected 1 Was 0. VVT1      [FAILED]

PR #857, opened 2022-05-31
integerPID improvements and negative values on VVT

powermos:
Change the PID implementation, to allow negative values on the input, change the anti-windup machanism, remove the unused and undocumented pOnE condition. This will create a better integral action, for better control, the former implementation did had windup problems.

Allow negative values for vvtCLMinAng and vvtCLMaxAng, switch to signed variables and whole digit precision in tunerstudio

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -1159,8 +1159,8 @@ page = 10
       vvtCLKI             = scalar, U08,     128,      "%",        0.03125, 0.0,    0.0,   7.96,      2 ; * (  1 byte)
       vvtCLKD             = scalar, U08,     129,      "%",        0.00781, 0.0,    0.0,   1.99,      3 ; * (  1 byte)
       vvtCL0DutyAng       = scalar, S16,     130,    "deg",        1.0,     0.0, -360.0,  360.0,      0 ; * (  2 bytes)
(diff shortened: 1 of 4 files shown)


TBAMax:
Looked at this. All this PID thing seems to be kind of raw. Maybe originally the old library. The current standard Arduino library is using foats and so is quite solid but with really bad impact to the performance. There is also unofficial integer version that I have found:https://github.com/mike-matera/FastPID but this is lacking the imprevements, for example "Proportional on Measurement" that seems quite big improvement (http://brettbeauregard.com/blog/2017/06/proportional-on-measurement-the-code/).
So the good solution is write and adjust our own as now.

What catches my eye here at first glimpse is: applying bit shifts to signed variables
`FeedForwardTerm <<= PID_SHIFTS;`
Also for example `output >> PID_SHIFTS;`
Bit shift applied to the signed variable without it being clearly limited to positive values only is just looking for trouble. As the library lets you have impression that negative values are also enabled, while they really are not catered for. Although it is possible to write integer PID that caters for negative values too.

VitorBoss:
The shifts are because the fixed point math, shifting a negative value is possible and the compiler take care of it as only the right shifting need attention (arithmetic/logical shift).
This fear led to the current code limitation for negative numbers on the first place.

As this PR goes it didn't improve anything, it does remove some features introduced before. You don't need negative target if you apply an offset to the values.

adbancroft:
Closing, duplicated by #1450, #1460, #1497

PR #858, opened 2022-06-01
Fix STM32 hanging on boot

### LINKED ISSUE #841: BUG (stm32): Comms/startup issue Master branch
Processor: STM32F407VGT

Symptom:
When plugging board into power it fails to boot. 

Steps to replicate:
Compile master d2b9612409bff3b54a172f77014e4d6ca621db4e
Flash to stm32
unplug board 

plug board in 
board fails to boot speeduino

pazi88:
In current master STM32 will hang on boot if no RTC battery is installed. The boot will either take long time or fails to boot completely. The reason for this is that the RTC will need to use external low speed clock (LSE_CLOCK) to be accurate. But "cold start" of the LSE_CLOCK takes up to 2 seconds and this is what causes the hanging. There seems to be no way around this long "cold start", so this PR basically moves the startup of the LSE_CLOCK to the moment of setting the RTC time. Otherwise internal low speed clock is used for RTC. Which is fast starting but not accurate (and also not powered by the battery).

I have tested that the board doesn't anymore hang up on booth with or without the RTC battery. RTC also keeps time normally with battery installed. Only slight problem is that when first time setting the RTC time, you will most likely get timeout in TS, because the timeout is set to 1500ms, when usually starting the LSE_CLOCK takes longer. But clicking set time again solves this.

Fixes #841

### DIFF
--- a/speeduino/board_stm32_official.ino
+++ b/speeduino/board_stm32_official.ino
@@ -94,7 +94,10 @@ STM32RTC& rtc = STM32RTC::getInstance();
     * Real Time clock for datalogging/time stamping
     */
     #ifdef RTC_ENABLED
-      rtc.setClockSource(STM32RTC::LSE_CLOCK); //Initialise external clock for RTC. That is the only clock running of VBAT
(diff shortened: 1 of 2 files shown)


noisymime:
Would it be worth increasing the timeout for stm32 board in the ini to prevent the problem when setting the RTC time?

Issue #161, opened 2018-09-05
AFR Safety System


noisymime:
Per forum: https://speeduino.com/forum/viewtopic.php?f=12&t=2048

Feature request:
A simple form of afr safety, so you dont blow your engine if a pump or an injector fails. 
Setup would be something like:
1: Above X kpa
AND/OR
2: Above X tps
IF
3:Afr not below X
IN
4: X seconds
DO
Cycle some ignition cuts so the driver feels somethings wrong AND/OR turn of boost control/lower boost

zrparker:
Wouldn't it be better and maybe easier to have something like:
IF AFR>(TARGET AFR+SET VALUE)
 Cut spark

That way it will be checking at all speeds and loads. Or, heck, tossing in a MAP or TPS min would make it so that it couldn't cut under low loads.

ric355:
Must account for DFCO bit being on, where AFR will shoot up very lean. Probably wouldn't make a difference immediately as there is no fuel at that point, but the time period between fuel coming back on and AFRs responding is relevant.  Note also that checks against AFR must account for the time delay which is both load and RPM based and different for every engine.

ric355:
Must also not interfere with the lean spot you tend to get when accel enrichment isn't tuned well enough.

zrparker:
> Must account for DFCO bit being on, where AFR will shoot up very lean. Probably wouldn't make a difference immediately as there is no fuel at that point, but the time period between fuel coming back on and AFRs responding is relevant. Note also that checks against AFR must account for the time delay which is both load and RPM based and different for every engine.

TPS min would account for DFCO. Time delay is already accounted for in the original idea and so is the accel enrich issue with this:

>IN
4: X seconds
DO

ric355:
> > Must account for DFCO bit being on, where AFR will shoot up very lean. Probably wouldn't make a difference immediately as there is no fuel at that point, but the time period between fuel coming back on and AFRs responding is relevant. Note also that checks against AFR must account for the time delay which is both load and RPM based and different for every engine.
> 
> TPS min would account for DFCO. Time delay is already accounted for in the original idea and so is the accel enrich issue with this:
> 
> > IN
> > 4: X seconds
> > DO

[...]

PR #861, opened 2022-06-03
AFR protection function

larmicfre:
As requested in #161, we're now able to avoid melting pistons using AFR protection function.
Function will activate using selected cut method when user specified condition thresholds for AFR protection has been exceeded. Those conditions are following:
- MAP above x kPa
- RPM above x /min
- TPS above x %
- Maximum AFR/lambda deviation (e.g 1.47 AFR or 0.1 lambda)
- Time delay before cut (in seconds)

For reactivation, the following condition has to be met:
- TPS below x %

Since the AFR/lambda deviation value is added to current AFR target value, **a proper AFR target table is required**. Note also that this function **can only be used with a wideband sensor**.


larmicfre:
> Does the merge request mean that we can expect AFR protection in the next FW release?
> Have been looking for this feature for a while!

Hopefully. Still waiting for review.

noisymime:
Looks good and seems to work fine. Nice PR!

Would be nice to have some unit tests added for this at some point, but happy to pull this in now as it's a solid addition

runesmdk:
Can we have a simple option added, like Min MAP to activate, and then just a simple AFR set point, to be reached within X seconds, like the original #161 i suggested?

PR #868, opened 2022-06-09
Crank and ASE Taper Fixes

HWright9:
Two Fixes:

1. Crank taper was using incorrect BIT_CHECK(**TIMER_mask**, BIT_TIMER_10HZ)
Should be
BIT_CHECK(**LOOP_TIMER**, BIT_TIMER_10HZ). The result was inconsitent and delayed crank taper.

2. ASE needs to always run every loop not just at 10Hz to catch the case where it falls back into cranking. Previously it would miss the fall back into cranking and then add ASE on top of crank enrichment.

Broken (before code change):
image

Fixed (after code change):
image


noisymime:
I get what you're doing with the fix, but I'm wondering if there's a better approach as this introduces table lookup every loop. It's not a huge impact, but it is noticeable. 

Would it not be easier to leave the 10Hz check there, but wrap the entire contents of the `correctionASE()` function (Including the 10Hz check) in a check for cranking?

noisymime:
I've just pushed up some changes that should achieve the same fix as this, but without the need to run the table lookup every loop. 

Appears to have resolved the issue in my testing, but confirmation would be good if you're able to test also.

HWright9:
Hi, I haven't tested your code (I assume it will work) but having one loop run at 10Hz and another at 15Hz "feels bad" to me and there are a lot more conditionals in ASE to work though. I worry that will be hard to maintain and debug going forward.

I was thinking instead lets just merge cranking and ASE into the same function, since they run sequentially anyway and have a strong inter-relationship. This should make it a lot easier to keep them in sync and I think we can reduce some RAM and some math overhead.
I will have a go and present my idea in an update to this PR.

HWright9:
Have a look. I still need to clean out the old functions but the new function is correctionCrankingASE().
I like this because it executes all of cranking and ASE together at 10Hz. Also prevents constantly multipling ASE only to divide it back out in crank tapers. We now take the maximum of either which is the same result. I needed one more global variable which is crankCorrection, also this variable correctly shows only crank correction and not merged with ASE which helps debug.

HWright9:
If we go with this or something similar, am I also ok to clean up the old corrections schedulers? They appear to be dead code.

muuris:
I think it's a good idea to wrap conditions of cranking, ASE and tapers of both into one state machine-like function. I also think that crankCorrection variable suggested by @HWright9 would be helpful for both debugging and unit testing.

HWright9:
Just an update on this. I tested in the car today. The code proposed here works well. Also checked on the loops/sec changes between cranking and when in ASE and then no Crank or ASE showing the difference to when no crank or ASE is running between each mode. Note I had a few other changes to add new sensors so don't worry so much about the total value. What we can see is the latest code has the smallest imact to loops/sec in each mode.
Let me know if this is Ok with you and I will do the final cleanup and you can pull it in.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/wrighe/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/wrighe/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">



Loops/sec Changes |   |   |  
-- | -- | -- | --
  | Crank Taper | ASE | ASE Taper | No ASE
Original Code | 780 | 870 | 874 | 902
Delta vs No   ASE | 122 | 32 | 28 |  
Fix ver1 | 719 | 803 | 768 | 834
Delta vs No   ASE | 115 | 31 | 66 |  
Fix One   Function | 808 | 795 | 794 | 814
Delta vs No   ASE | 6 | 19 | 20 |  



</body>

</html>

noisymime:
Closing this as fi for issue went in with 61708a5716e42a4123d0245ff51b591bdd026b5a

If you want to take a further look at improvements around this, please open a new PR or incident to track it separately to this fix.

PR #869, opened 2022-06-12
rev

smb29:
New_board


gdiciocco (commented):
What will cause rpm limit reduction ? Me and probably others are running engines capable of way more than 10k rpms.

smb29:
buono per il motore del gruppo PSA, è più che sufficiente.
Non credo che il tuo motore possa raggiungere i 25.000 giri/min.

smb29:
> What will cause rpm limit reduction ? Me and probably others are running engines capable of way more than 10k rpms.

do you have a formula 1 or does envy eat you?

smb29:
> What will cause rpm limit reduction ? Me and probably others are running engines capable of way more than 10k rpms.

Do you have a formula one?
Or is it just a stock car?

brunob45:
There is this thing called "future proofing", which means that current limitations may not apply in the future.

For example, a use case you didn't think of is: motorcycles! 

Some production bikes can rev up to nearly 20'000 rpm. Your limit of 11'000 rpm is clearly too low for performance bikes, thus cannot be accepted.

PR #872, opened 2022-06-15
Added cranking cold advance option

VitorBoss:
On markets with pure ethanol fuel and cold temperatures the engine requires a LOT of advance to start rapidly.

This PR just add the possibility to use current cold advance added to cranking angle, is just a toggle.


HWright9:
If the problem is on ethanol, should the cranking advance depend on the flex fuel input? i,e instead of "off or "on" it could be a threshold for ethanol % to enable it and 0% would enable all the time.

VitorBoss:
There is a table in place for ethanol % add, but some engines demand more advance when cold, the cold advance don't add to cranking, just after running, this PR adds this possibility

PR #876, opened 2022-06-23
bug fix. A workaround for failing emulated on internal flash writes + a root cause fix

Tjeerdie:
Small workaround for the current issue when using internal flash as emulated eeprom on the STM32. Sometimes writes to emulated eeprom fail writing 255 to the location that has been written to.

This happens because  if somewhere else in the speeduino code. (can be anywhere) a write to a uninitialized pointer is done only on STM32 the internal flash falls into an error state. the first flash write after that error state will fail. 

Because i have no control over all of the speeduino code and writing to uninitialized pointers will find its way in code, the proposed workaround is to check and clear any errors in the flash programming hardware before writing to flash. 

edit:
I found the root cause in this particular case, but this will keep happening in the future so resetting errors before writing to flash is still a good workaround i think.
1. In the running initBoard() all timers of STM32 are setup. When resuming the timers the interrupt fires once, immediately. The next line the timers are stopped again
2. When firing the interrupt the fuel and ignition interrupts are run, but the setPinMapping(byte boardID) is not called yet. so the pins for fuel and ignition are not setup. 
3. during this interrupt routinel we enable/disable coilcharge etc. but the pins and therefore register addresses are not setup correctly yet
4. Writing to *ign1_pin_port &= ~(ign1_pin_mask) for example is writing to uninitialized pointers (*ign1_pin_port) causing the issue.

So the solution is to setup the timers manually with registers directly without relying on the STM32 core. So that the interrupt does not fire once. I am having a hard time getting this to work properly ATM. 

edit2:

Produced a change to fix the root cause issue. 
- Removed enabling the timers for various scheduled functions from the boardinit().
- Add the timer enable in the various IGN1_TIMER_ENABLE() etc macros. So these are only enabled AFTER the pin assignment function
-Added the INJX_TIMER_ENABLE() macros to be run after the fuel schedules are initialized in initialiseSchedulers() function. (So made it to be identical to the IGNx_TIMER_ENABLE macros already in that location.)


amasood1193:
what is the procedure for checking and clearing the errors.

Tjeerdie:
I do not fully understand your question but this is an example of what it does

example (from this PR)

Check flash status register for a write protect error.
if(__HAL_FLASH_GET_FLAG(FLASH_FLAG_WRPERR) != RESET) 

To clear the error (if it indeed is raised) run the following.
__HAL_FLASH_CLEAR_FLAG(FLASH_FLAG_WRPERR);

And that is done for all 5 possible errors of de flash writing hardware on STM32.

PR #879, opened 2022-06-26
bug fix EGO PID controller + some improvements. 

Tjeerdie:
A quick fix for a bug and some other quick improvements to make EGO PID atleast do something usefull. The next improvement steps would be to build a long and short term fuel trim table for Teensy and STM32. Mega probably has too little RAM for this, but who knows.  

Also the accuracy of the controller could be improved if the steps can be <1%. So making the EGO work with 0.1% steps can also be a next improvement step.

What i found and explained in Discord:

> ego PID control is a bit of a mess. I does not work (there is a bug) and probably never did work. The bug makes it so that the gains are varying over time. AFter X seconds the gains are effectively all zero. 
> 
> In PID_v1.ino: 
> Look at line 51. The SampleTime should be calculated here using lastTime and now
> Look at line 71. This is commented out. So lastTime is never updated. (this is done 7 year ago)
> Look at line 37 lastTime is set at the init time and wil keep this value forever
> 
> Look at line 93..96: the SampleTime is used there to change the gains. But now the gains are changing over time because of line 71 being commented out, making the  SampleTime increase with time. 
> 
> So after some time the ego control stops working because gains are zero or very small. It even has the change to be zero when millis() rollover and that would create div by zero errors! Un-commenting line 71 does not fix it, only P gain works then. The I gain remains unable to do anything.
> 
> Because the older version never worked I assume there would not be any setups using it successfully. So there is no point in retaining it?

What has been done is fixing the bugs:
- Make the gains fixed and not vary with Tsample. Changing the gains depending on the sample time (which depends on the RPM because it is calculated every X ignition cycles). The higher the RPM the faster the reaction of EGO can be. So reducing the gains with higher RPM negates that benefit. 

What has been done to improve the controller
- Choose gain scaling such that these fall in the realm of usable gains for a ICE engine. With the original scaling i could only use 1 to 2 I gain. The whole range for i gain is 0 to 200. 
- Disable PID control calculation when DFCO is active. (The out put was already disabled during DFCO but the interator kept running. So when coming out of DFCO the controller was full rich).

### DIFF
--- a/speeduino/corrections.ino
+++ b/speeduino/corrections.ino
@@ -676,7 +676,7 @@ byte correctionAFRClosedLoop()
     if( (currentStatus.runSecs > configPage6.ego_sdelay) || (configPage2.incorporateAFR == true) ) { currentStatus.afrTarget = get3DTableValue(&afrTable, currentStatus.fuelLoad, currentStatus.RPM); } //Perform the target lookup
   }
   
-  if( configPage6.egoType > 0  || (BIT_CHECK(currentStatus.status1, BIT_STATUS1_DFCO) == 1  ) ) //egoType of 0 means no O2 sensor. If DFCO is active do not run the ego controllers to prevent interator wind-up.
+  if((configPage6.egoType > 0)  & (BIT_CHECK(currentStatus.status1, BIT_STATUS1_DFCO) != 1  ) ) //egoType of 0 means no O2 sensor. If DFCO is active do not run the ego controllers to prevent interator wind-up.
   {
     AFRValue = currentStatus.egoCorrection; //Need to record this here, just to make sure the correction stays 'on' even if the nextCycle count isn't ready
     
--- a/speeduino/corrections.ino
+++ b/speeduino/corrections.ino
@@ -676,7 +676,7 @@ byte correctionAFRClosedLoop()
     if( (currentStatus.runSecs > configPage6.ego_sdelay) || (configPage2.incorporateAFR == true) ) { currentStatus.afrTarget = get3DTableValue(&afrTable, currentStatus.fuelLoad, currentStatus.RPM); } //Perform the target lookup
   }
   
-  if( configPage6.egoType > 0 ) //egoType of 0 means no O2 sensor
+  if( configPage6.egoType > 0  || (BIT_CHECK(currentStatus.status1, BIT_STATUS1_DFCO) == 1  ) ) //egoType of 0 means no O2 sensor. If DFCO is active do not run the ego controllers to prevent interator wind-up.
   {
     AFRValue = currentStatus.egoCorrection; //Need to record this here, just to make sure the correction stays 'on' even if the nextCycle count isn't ready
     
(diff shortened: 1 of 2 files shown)


VitorBoss (commented):
Did you mean DFCO == 0?
This will run only at DFCO if the check is 1

Tjeerdie:
Uhh yes. I made a double mistake even:
& (BIT_CHECK(currentStatus.status1, BIT_STATUS1_DFCO) != 1  ) 

Its not OR it must be AND. and indeed not equal to

VitorBoss:
Instead of using `kD * 10` maybe a second derivative can be used for `dInput`?

Tjeerdie:
> Instead of using `kD * 10` maybe a second derivative can be used for `dInput`?

a second derivative would be a completely different metric. If controlling a speed the first derivative would Kd reacting on acceleration. Using the second derivative it would react on Jerk Making it a PIJ controller. One of the reasons this is not used in PID is of the enormous noise amplification it has.  

image

I actually normally have the D gain at zero anyway. It does not add much benefit. The second derivative would do even less benefit. The main gain of the controller is the I gain. And some benefit can be had from the P gain. 

Speed controlled electric drives for electric motor for example employ a PI controller a lot of the times.

ric355:
Does the issue this change fixes also show up in stepper idle control? The reason I ask is that a long time ago I tried to enable closed loop stepper idle control and just couldn't get it to work reliably. I found that it seemed great for a short while (maybe 30 seconds or so) but then it just seemed to lose control of the RPM resulting in a high idle. It was as though it just pushed the stepper to max open and then just gave up. If I restarted the engine it was back to normal again and then would do the same thing.

Tjeerdie:
@ric355 For idle an other PID i used. That does not calculate its own Tsample. So its definitely not the same issue. 

If you are using open+closed loop That particular issue sounds like you are having to small value for the RPM hysteresis (or to large open loop value). If the (target RPM + hysteresis RPM < RPM where the controller hangs) this is probably the issue. 

Lower the the open loop value in the open loop curve or increase the RPM hysteresis. Recently added open+close loop explanations to the wiki. So might be useful to read that.

ric355:
@Tjeerdie Thanks for the info.   The issue I am referring to predates the open+closed loop implementation by a long way. I think it was somewhere around 2019 when I tried it. I was hoping you'd say yes, then I would have at least had something to explain why it didn't work!

What struck me was it seemed to work for a while then it just seemed like it lost control. I've never gone back to it to try it on a more recent firmware but I don't think it has changed much anyway. I'm still on 2020 firmware at the moment - planning to upgrade to the next release provided an issue doesn't pop out that stops me from doing so (had a few of those lately, unfortunately).  I will try open+closed loop after upgrading.

VitorBoss:
It sounds like your steper is working backwards, since 2019 Tjeerd and I had added a lot of fixes and enhancements to idle control, please try again and share your results.

Pure PID control for idle never worked before, it had too much bugs, now is someway usable, but I strongly recommend you use the hybrid control

PR #880, opened 2022-06-26
additional typos found by codespell

dantob:
This does rename one flag, otherwise the rest should be just comment/doc changes
tachoOutputFlag, DEACTIVE->INACTIVE

didn't realise there was a tool for this, but tree is basically clean now, I'll look at setting up a CI


noisymime:
Looks good, nice cleanup!

PR #882, opened 2022-06-27
Add new CANbus display,

sbtoonz:
Add rebroadcasting interface for the Gauge.S which I helped design. This allows Speeduino to rebroadcast various data to this OLED dsplay panel in order for a user to see what is happening live as well as log this data to the SD card inside Gauge.S

[Gauge.S Github](https://github.com/handmade0octopus/gauge.s-sorek.uk)

Signed-off-by: sbtoonz <sbtoonz@gmail.com>


sbtoonz:
I suppose this could get renamed to something like 
"Generic CAN rebroadcasting"

PR #884, opened 2022-07-01
Reorganize Menu's

8SecSleeper:
This is an attempt to fix the mess that has become of the menu's. Appears to be one aspect of this project that hasn't had much love.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -1753,101 +1753,114 @@ page = 15
    ;
    ;----------------------------------------------------------------------------
 
-menuDialog = main
-
-   menu = "Settings"
+menuDialog = main  
+   menu = "System Setup"
       subMenu = engine_constants,   "Engine Constants"
+	  subMenu = triggerSettings,    "Trigger Setup"
+	  subMenu = sparkSettings,      "Ignition Setup"  
       subMenu = injChars,           "Injector Characteristics"
-      subMenu = triggerSettings,    "Trigger Setup"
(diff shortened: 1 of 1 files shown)


DeionSi:
Feels really nice. It can be pretty confusing with how it is now.

The window names should probably be updated to be the same as the menu items.

Maybe use the Setup and Settings terms consistently? It seems to be mixed here and there.

Maybe move the 3D tuning maps to Fuel and Spark?

Maybe Timing Vs IAT/CLT should be named more historically? Like "IAT timing retard" and "CLT timing advance"?

Can Calibrate TPS be moved to Sensor Calibration?

reference/speeduino.ini
```diff
@@ -1753,101 +1753,114 @@ page = 15
...
-      groupMenu = "Engine Protection"
-        groupChildMenu = engineProtection,          "Common Engine Protection"
-        groupChildMenu = revLimiterDialog,          "Rev Limiters",             { engineProtectType }
-        groupChildMenu = boostCut,                  "Boost Cut",                { engineProtectType }
-        groupChildMenu = oilPressureProtection,     "Oil Pressure",             { engineProtectType }
-        groupChildMenu = afrProtect,                "AFR Protection",           { engineProtectType }
-      subMenu = flexFuel,           "Flex Fuel",        2
+      subMenu = prgm_out_config,  "Programmable Outputs"
+	  
+   menu = "Sensor Caibration"
```
DeionSi, line 1772: Misspelling

PR #896, opened 2022-07-21
Fix Jeep2000 decoder

jonbills:
getCrankAngle_Jeep2000() sets crankangle = 146 if toothcurrentcount = 0, because last tooth was cam tooth  which is 146 ATDC, however, all the timing and angle calculations are not done on the cam tooth read, so crankangle should be set to 114 for previous crank tooth.
problem was found by @jaky 47 on Slack, and fix tested and confirmed by him.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -7,7 +7,7 @@
    MTversion      = 2.25
 
    queryCommand   = "Q"
-   signature      = "speeduino 202207"
+   signature      = "speeduino 202210-dev"
    versionInfo    = "S" ;This info is what is displayed to user
 
 [TunerStudio]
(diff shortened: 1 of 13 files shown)


jonbills:
ooh,  should this have been a PR against the 202207-fixes branch instead?

jonbills:
Problem description in #897

JaredSellers4:
@jonbills, the crank tooth degrees in this decoder are rising edge so should it actually be 114 degrees?

jonbills:
It certainly fixed the problem of the duplicate and very early signal. I think youre right that 114 would be more correct. I'll update to that.
Most of the comment is the original authors, I just appended a note on the correction.

jonbills:
redoing as two PRs - one an LTS 'fix' for the original logic error , the other a non LTS 'improvements' (noise filtering)

PR #899, opened 2022-07-23
fix check for DFCO off in correctionAFRclosedLoop()

jonbills:
line 608 in corrections.ino has bit-and of configpage6.egotype > 0  & DFCO off.
it should be logical and &&
I think the effect of the fault is to keep the ego correction in place, so when DFCO goes off, the correction is back on, although the engine is in (potentially) very different conditions at that point so might be too lean or too rich.

### DIFF
--- a/speeduino/corrections.ino
+++ b/speeduino/corrections.ino
@@ -605,7 +605,7 @@ byte correctionAFRClosedLoop()
     if( (currentStatus.runSecs > configPage6.ego_sdelay) || (configPage2.incorporateAFR == true) ) { currentStatus.afrTarget = get3DTableValue(&afrTable, currentStatus.fuelLoad, currentStatus.RPM); } //Perform the target lookup
   }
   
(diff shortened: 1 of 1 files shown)


Tjeerdie:
Correct should haven been a boolean (&&) not a bitwise (&). Still not sure if the effect when running this code is different

The idea behind this is indeed to stop the PID from rubberbanding. Before this fix the PID keeps running during DFCO, reading lean while DFCO is on and trying to correct with enrichment of the mixture. (So integral wind up takes place, or rubber banding, what ever you want to call it). 

So then when going out of DFCO the ego was fully rich instead of at the point of where it was before DFCO. The alternative could be to reset the integral of the PID to 0 after getting back from DFCO.

noisymime:
> The alternative could be to reset the integral of the PID to 0 after getting back from DFCO.

This disadvantage of this is that if the integral has reached a 'good' value, then you lose that when you come off DFCO, which could cause the PID to need to start again. It's unlikely to be a huge difference, but I think simply turning the loop off during DFCO is the better option.

noisymime:
For reference, please raise all PRs against the `master` branch. If relevant, the commit will be cherry picked into `-fixes`

noisymime:
Merged to `master` and `202207-fixes`

PR #901, opened 2022-07-26
cl boost control above baro should use logical and/or

dantob:


Tjeerdie:
Should indeed use boolean AND/OR not bitwise AND/OR. Not sure if the code would execute differently in this case,

noisymime:
Merged to `master` and `202207-fixes`

PR #902, opened 2022-07-26
Volatile decoder variables

DeionSi:
triggerToothAngle is modified by interrupts, hence it needs to be read when interrupts are disabled.

triggerActualTeeth is only set in decoder setup routines, hence is not set in an interrupt and does not need to be volatile.

Firmware size decrease by 68 bytes. Loops/sec is unchanged.


noisymime:
Good pickup. Did you see this actually causing any issues or just a cleanup?

noisymime:

PR #903, opened 2022-07-27
Add OLCL mode to initBoard() for teensy

a32guy:
This fixes an issue of the idle valve not operating on Teensy based (Dropbear) boards if PWM OL+CL IAC mode is selected. I didn't test with Teensy 4.1 but I'd imagine the change would be the same.


noisymime:

PR #904, opened 2022-07-29
Stall reset changes

DeionSi:
Move decoder stall checks and decoder variable resets to decoders.ino as separate functions. Primary purposes are for clarity and to allow for unit testing.

Also prevents interrupts when stall checking and resetting to prevent interrupt race issues.

Working as expected when bench-testing. 4 bytes reduced heap memory and 72 bytes smaller firmware. Loops/sec unchanged.

### DIFF
--- a/speeduino/comms.cpp
+++ b/speeduino/comms.cpp
@@ -244,7 +244,7 @@ void processSerialCommand(void)
       break;
 
     case 'b': // New EEPROM burn command to only burn a single page at a time 
-      if( (micros() > deferEEPROMWritesUntil)) { writeConfig(serialPayload[2]); } //Read the table number and perform burn. Note that byte 1 in the array is unused
+      if( (micros() - lastEEPROMDeferTime > EEPROM_DEFER_DELAY)) { writeConfig(serialPayload[2]); } //Read the table number and perform burn. Note that byte 1 in the array is unused
       else { BIT_SET(currentStatus.status4, BIT_STATUS4_BURNPENDING); }
       
       sendSerialReturnCode(SERIAL_RC_BURN_OK);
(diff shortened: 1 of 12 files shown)


noisymime:
You need to keep the `|| (toothLastToothTime > currentLoopTime)` check in there somewhere. Without it you'll likely get a sync loss and hesitation when `micros()` overflows after around 70 minutes.

DeionSi:
Thanks for the comment, I hadn't really thought about that. However the isDecoderStalled function should be safe for micros() overflow. See https://www.gammon.com.au/millis

I believe `(toothLastToothTime > currentLoopTime)` was for possible interrupts firing between getting the time and comparing as per the comment `toothLastToothTime can be greater than currentLoopTime if a pulse occurs between getting the latest time and doing the comparison`. Since interrupts are now disabled during the check it is no longer a possibility.

noisymime:
The comment isn't entirely complete there. The issue that would be faced is when:
1. `toothLastToothTime` gets set using `micros()` just after approx. 1 hour, 11 minutes, 34 seconds
2. `micros()` rolls over
3. If the next mainloop happens before the next tooth, `currentLoopTime` will be a very small number and `toothLastToothTime` will be very large. This is not handled currently in `isDecoderStalled()` and will cause a very large and incorrect value for `timeToLastTooth`

DeionSi:
I've added tests for isDecoderStalled. To be able to test I've added a "test injection" definition for micros_safe and micros to allow testing with different values. isDecoderStalled is also moved to header file to allow inline usage from test source files.

This pull request now has #976 as a prerequisite.

-----------------------

In response to your comment:

1. Let's assume toothLastToothTime is ULONG_MAX-1 (4 294 967 294)
2. micros() rolls over and is now 4

This code
```
uint32_t timeToLastTooth = (MICROS_SAFE_OR_TEST_INJECTION - toothLastToothTime);
if ( timeToLastTooth > MAX_STALL_TIME ) {
```

Would calculate as
```
uint32_t 6 = (4 - 4 294 967 294);
if ( 6 > 50000 ) {
```
As an uint32_t cannot be negative -4294967290 it will "underflow" all the way around to 6.

The tests test for this condition.

adbancroft:
IMO, with #1528 (and others) we have incorporated the intent of this PR.

PR #908, opened 2022-08-01
Fix for idle PWM on STM32. Enable idle timer when idle pwm is enabled…

Tjeerdie:
Fix for idle PWM issues with STM32.


noisymime:
Ported to `202207-fixes`

PR #909, opened 2022-08-06
Fuel pump control via Hardware test

HWright9:
This PR adds the ability to turn on and off the fuel pump via the hardware test mode. This is a very useful function for injector flow testing.
Also the injector 1 and 2 status variables are included in the datalog when the test mode is active. In this way injector flow testing can be performed on the vehicle using the hardware test mode. For example by lifting the fuel rail off and squirting the injector into a jar then weighing it before and after. The benefit is that this injector test uses all the vehicle hardware in situ so the operation of the fuel pump and pressure regulator is exactly as it will be when the vehicle is running. If your vehicle has a fuel rail pressure sensor this testing can also be done at multiple fuel pressures. 

To get this to work I had to consolidate the fuel pump control into one function. This is much clearer and easier determine fuel pump operation and always aligns the datalogged fuel pump state with the pin mode. I found a few minor bugs such as places where the fuel pump was not using the macro for turning off so this fixes those also. Overall fuel pump operation is the same as before with the following minor changes:

1. The fuel pump function now runs at 10Hz so there will be at most a 0.1sec delay when turning on the fuel pump at controller init and when the engine starts moving. I judged this to be acceptable, but easy to run this function faster if required.
2. I added a small de-bouncing logic to prevent intermittent trigger signals from toggling the fuel pump. If the engine stalls this currently adds a 0.2 sec delay to fuel pump turning off. It could easily be removed if this functionality is not required.

image


adbancroft:
This PR isn't merge-able, but the idea is useful. Partially implemented in #1617.

PR #910, opened 2022-08-09
Disable automatic retransmission on STM32 CAN

pazi88:
I noticed that I have accidentally left the automatic retransmission enabled on the new STM32 CAN library. It has been observed to possibly cause some sync loss issues, so it's better to to be disabled.


noisymime:

Issue #897, opened 2022-07-22
Jeep 2000 decoder, duplicate and early ignition


jonbills:
with the Jeep 2000 decoder, you get a duplicate, early (30° ish) ignition signal on channel 3.
![image](https://user-images.githubusercontent.com/5832708/180442944-b045ff1f-ae86-4262-beb5-53102d0a3dcf.png)
This only happens with ignition advance approx 10° to 40°.
I believe this is caused by miss-setting the crank angle after the cam signal resets the tooth count to zero.
#896 is intended fix.

JaredSellers4:
What software did you use to generate and view that graph?  That would be super helpful for what I am dealing with right now.

jonbills:
I
I believe @jaky 47 captured it with an oscilloscope

PR #912, opened 2022-08-13
Jeep2000 fix

jonbills:
This replaces #896 which included an 'LTS' style fix plus some improvements which wouldn't be in an LTS release. 
(also I'd made the PR out of my master branch, which made things tricky!)
This PR is now ONLY a fix for #897 
the fix is: getCrankAngle_Jeep2000() sets crankangle = 146 + triggerAngle when toothCurrentCount = 0, because toothCurrentCount was set to 0 on seeing the Cam tooth, and the cam tooth angle is 146.
However, the tooth timings that are subsequently used in calculations are from the previous crank tooth, which was at 114. Therefore 114 is the correct crankangle to work with.

### DIFF
--- a/speeduino/decoders.ino
+++ b/speeduino/decoders.ino
@@ -1785,13 +1785,9 @@ void triggerSetup_Jeep2000()
 
 void triggerPri_Jeep2000()
 {
-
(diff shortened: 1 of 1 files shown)


noisymime:

PR #913, opened 2022-08-13
Jeep 2000 noise filtering

jonbills:
The current Jeep 2000 decoder has no code for noise filtering on cam trigger, and the crank trigger filtering can't be used because of the variable gap between teeth.
This contains two changes to address that. 
*  only accept cam tooth triggers when crank tooth is > 11 (i.e. 12 or 13) because a cam tooth at any other time must be noise. 
* setfilter() for the next crank tooth based on scaling the curGap by ratio of current tooth angle and  next tooth angle - thereby increasing the filter time allowance if the next gap is bigger and reducing it if the next gap is smaller.

### DIFF
--- a/speeduino/decoders.ino
+++ b/speeduino/decoders.ino
@@ -1849,7 +1849,11 @@ int getCrankAngle_Jeep2000()
     interrupts();
 
     int crankAngle;
+<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


JaredSellers4:
@jonbills , did you mean to Close this (deleted head repository)?

jonbills:
yes, I replaced it with https://github.com/noisymime/speeduino/pull/943 - because I'd made the original PR against my master branch, which made it hard to do anything else :)
if you comment on the new one, it might get merged.

PR #916, opened 2022-08-21
Fix problem of using an uninitialised variable when initialising the stepper after power on

ric355:
This fixes a bug I found after upgrading from May 2020 to July 2022. I use open loop stepper mode currently. The symptom was that the stepper motor behaviour was basically random. Some starts would result in really low idle RPM (stepper closed off) and others somewhat high RPM (stepper apparently wide open). Altering the curve did not fix it and in any case there should be no reason why the curve should be different between upgrades. 

I discovered that about 8 months ago a change was made to stop the stepper from being re-homed when the idle mode is changed (e.g. from stepper open loop to stepper open+closed). This is because depending on the engine setup it may cause a stall. That is not a real world problem since in normal running it isn't something that will happen, but there's no problem retaining the feature as far as I can tell.
There are multiple issues being fixed here although they all relate to the same underlying cause; 
- going from USB power only to ignition power could result in the stepper being in a random position, making the engine idle point different each time it is started under those conditions.
- initial power on with key will not always re-home the stepper and leaves a random position. Again this results in random idle speed.

This change just alters the method by which the decision is made for re-homing to make it deterministic. It also removes the accessing of the uninitialised variable from the condition, the original purpose of the code being to initialise that variable (compeltedHomeSteps) and others.

Note there is a very slight difference in behaviour in the situation where homing is not completed but the stepper mode is changed. In this scenario, homing will continue until completed whereas previously it would be restarted. The end result should be the same.

I recommend this fix is added to the LTS branch.

### DIFF
--- a/speeduino/idle.h
+++ b/speeduino/idle.h
@@ -69,7 +69,7 @@ uint8_t idleTaper;
 byte idleUpOutputHIGH = HIGH; // Used to invert the idle Up Output 
 byte idleUpOutputLOW = LOW;   // Used to invert the idle Up Output 
 
-void initialiseIdle();
+void initialiseIdle(bool forcehoming);
 void initialiseIdleUpOutput();
 void disableIdle();
 void idleInterrupt();
--- a/speeduino/idle.ino
+++ b/speeduino/idle.ino
@@ -32,7 +32,7 @@ static inline void enableIdle()
   }
 }
 
-void initialiseIdle()
+void initialiseIdle(bool forcehoming)
 {
   //By default, turn off the PWM interrupt (It gets turned on below if needed)
   IDLE_TIMER_DISABLE();
@@ -154,7 +154,7 @@ void initialiseIdle()
       iacStepTime_uS = configPage6.iacStepTime * 1000;
       iacCoolTime_uS = configPage9.iacCoolTime * 1000;
 
-      if( completedHomeSteps < (configPage6.iacStepHome * 3) )
+      if (forcehoming)
       {
         //Change between modes running make engine stall
         completedHomeSteps = 0;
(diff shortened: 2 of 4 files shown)


noisymime:
Would a less invasive method to resolving this simply be to set an initial value of 0 to `completedHomeSteps`? AS far as I could see, that would resolve the issue and could be added easily to fixes without any change in behaviour.

ric355:
No, by itself that wouldn't fix the problem. I did consider it. It only addresses the initial pure power-on scenario by ignition key.

The issue is not only that the variable is uninitialized, but that the condition used in initialiseidle() is basically incorrect in certain circumstances. The condition is determining whether homing has been completed yet, and if homing has not been completed, only then does homing get re-executed.

The problem with that approach is that once homing has been completed, it won't *ever* be re-done when initialiseidle() is called. There are two conditions where re-homing is expected:
1. When the engine stalls. Historically at least, the stepper would be re-homed in that circumstance and I think that should still be done, even though you could argue it should be OK without as the internal state should match the stepper.
2. When power changes from USB power to 12v power. This is absolutely necessary because with only 5v power the firmware will execute homing as soon as USB is connected but this won't actually affect the position of the stepper since it has no 12v supply until the key is turned on. Ergo the stepper is immediately out of sync with the firmware. You would have to be extremely quick with the 12v supply to catch it before homing was completed.

If you only change the initial value of completedHomeSteps to zero (I presume you mean at declaration or before the first initialiseidle() call) then the current test will continue to return false for the two conditions above. Most people plug their laptop in before starting the engine; that's why we had to add that voltage detection difference in the first place.

With (2) in particular, you can get yourself into a real mess if you're trying to tune the stepper for starting purposes. It becomes wildly inconsistent between starts. Just setting an initial value of zero won't change that.

noisymime:
Replaced by #926

PR #920, opened 2022-08-28
Fix improper log entries

scudderfish:
Format typos that cause the data to not be logged


scudderfish:
I spotted that these two values were not logging properly in MSDroid as the '%' was doubled.  Removing the leading '%' results in the values appearing in the log files

noisymime:
Good pickup!

noisymime:
Cherrypicked into `202207-fixes`

PR #926, opened 2022-09-11
Fix for another uninitialised variable issue in OLCL and CL stepper idle control

ric355:
The deprecated CL and the newer OLCL stepper idle PID based control algorithms both use the same piece of code inside idleControl() with some minor variations in logic depending on which is active. 

All of the idle control functions use a 'taper' time which is managed inside idle control when the RPM transitions between cranking and running RPMs. The taper counter is reset to zero when cranking is identified, and then counted at 10hz when running until it reaches the configured time. The taper is used to access the running map, thus giving a more gentle transition between the cranking and running idle valve or stepper position.

Unfortunately the taper control code is missing from the CL / OLCL stepper code. There are two issues with this:
- Firstly the taper counter is never reset to zero, hence when it is accessed later after the engine starts, an uninitialised variable is being used.
- Secondly, the taper counter is never incremented and so the taper will either never start, or never end, depending upon the random value of the un-initialised variable.

This fix has been tested by someone who was experiencing the problem and was able to reproduce it at will on his test bench.

Note that this branch also contains the previous pull request I made regarding the problems around homing.  Both of these changes really need to be pulled in because at present for anyone using a stepper for idle control trying to use OLCL it is very much broken.

I recommend this fix is added to LTS.


noisymime:
Cherry-picked into `202207-fixes`

PR #931, opened 2022-09-17
Jeep4cyl initial

JaredSellers4:
Adding decoder for Jeep 4cyl 1991-1995 YJ very similar to Jeep2000 6cyl decoder but compatible with 4cyl sequential ignition and port injection, therefore also works with paired, single output etc..  Stock had dizzy but this gives option to upgrade to CNP.  16 crank teeth over 720 degrees, in groups of 4. Cam wheel is high for 360 crank degrees.  Tooth number 1 represents the first tooth seen after the cam signal goes high.  https://speeduino.com/forum/download/file.php?id=8470
Early testing has been done by @jaky47 on bench sim and @rowdydouglas cranking real motor without fuel.

Compared to Jeep2000 cyl:
- Only for 4cyl
- Added doxygen markup similar to dual wheel
- removed some effects from TS settings such as + configPage4.triggerAngle from multiple entries
- removed value for TS trigger filtering since doesn't work for varying gap patterns
- added values for sequential such as revolutions and tracking all teeth for 720 similar to dual wheel
- changed tooth angles to use falling edge to match more with cam angle (ex. falling 296 instead of rising 294 because cam rises on 236, probably same results but felt better have dividends of 60 instead of 58)
- Included equivalents from from @jonbills and @jaky47 work and testing on Jeep 2000 6cyl decoder in https://github.com/noisymime/speeduino/pull/913 , that is only allowing secondary trigger when primary tooth count is completed rotation and more filtering work to account for vary gaps on primary trigger


JaredSellers4:
I didn't mean to include that tune file (CurrentTune.msq).

JaredSellers4:
I also meant to give a shout out to @reliant_turbo for initial decoder conversion from Jeep2000 6cyl to 4cyl in the forums https://speeduino.com/forum/viewtopic.php?f=13&t=4710&start=70

JaredSellers4:
It runs! It started right up for the first time in 10 years and idled! This is running LS1 Coil Per Plug with the decoder in this pull request.

JaredSellers4:
I am working to get this 1990s Jeep 4cyl decoder up to speed with the current noisymime master.  I am not sure if I will be able to use this same PR by replacing commits or if it has to be a whole new PR.

JaredSellers4:
I decided on a new PR.  Code changes are done in https://github.com/JaredSellers4/speeduino/tree/masterJeep4cyl and now I need to test on vehicle.

PR #935, opened 2022-10-02
Add Tacho Sweep function at startup

wonder99:
Enables the Tacho Sweep options in TunerStudio.  Ramps the Tach needle upon powerup.


wilmardo:
Followup of #724 I assume? Would be nice to have @shiznit304 report again on this rebased PR :)

wonder99:
Yes it's basically the same, but I separated out another design change which should make this easier to accept

PR #939, opened 2022-10-12
Add deceleration enleanment feature

pazi88:
When reducing throttle, the decrease in manifold air pressure causes fuel that has condensed on the walls of the intake manifold to evaporate and be ingested by the engine. This can result in a temporary rich AFR and to counteract that in speeduino, this PR adds functionality to reduce injector pulse width during deceleration events. This is mostly same functionality as acceleration enrichment but works in deceleration, so it uses mainly same code as AE. The thresholds to trigger decel. are same as in AE and also same code/settings are used for duration. The amount of deceleration enleanment is fixed and can be set in TS. This is really simple but surprisingly effective implementation of decel enleanment and also used in other aftermarket ecus. There doesn't seem to be that much need to have the deceleration fuel amount depend on RPM/load. Even in OEM ecus the load/RPM maps for deceleration fuel amount have miniscule changes, so I didn't think it would be worth the trouble to introduce separate maps for that.

Actual PR might look rather big compared how little actual change there is to functionality, but the changes are mainly there to allow negative tps/mapDOT values. Also some bit of optimization is involved in this PR, making RAM usage less and loops per sec. higher than before. I have tested quite extensively that there is no change in existing AE functionality after this change.

Before this PR:
RAM:   [==========]  95.0% (used 7786 bytes from 8192 bytes)
Flash: [========  ]  84.4% (used 214304 bytes from 253952 bytes)
216 bytes of free RAM when running
954 loops/s @ 1000rpm
816 loops/s @ 5000rpm

After this PR:
RAM:   [==========]  95.0% (used 7784 bytes from 8192 bytes)
Flash: [========  ]  84.6% (used 214954 bytes from 253952 bytes)
218 bytes of free RAM when running
966 loops/s @ 1000rpm
826 loops/s @ 5000rpm


ric355:
What is the reason for changing tpsDot and mapDot to be divided by 10 on the comms? When changes like this are made it affects all applications that consume data through the secondary serial interface as they now all have to be changed to multiply by 10.

pazi88:
> 

The tpsDOT and mapDOT were originally divided by 10 to fit onto byte. But new with negative map/tpsDOT that isn't really feasible, because you only get +/-127 of range there. So map/tpsDOT is now int16, instead of uint8 and WITHOUT divided by 10. So in the cancomms the value is divided by 10 to keep data same as before.

pazi88:
> I would suggest you to keep the previous fullStatus structure and move the new sales to the end to keep backwards compatibility without Givin up on new scales

Only thing that needs backwards compatibility is the cancomms (secondary serial). Adding more stuff (duplicates of what is already there) to fullstatus would increase RAM usage for nothing really.

VitorBoss:
That will break current programmable outputs configurations as the indexes is changing

pazi88:
> That will break current programmable outputs configurations as the indexes are changing

You are right. Totally forgot about that. Easy fix though and I also found mistake and missing items there.

VitorBoss:
If you change only the ini that will fix future configurations, for existing ones it need to be coded in updates.ino

pazi88:
> If you change only the ini that will fix future configurations, for existing ones it need to be coded in updates.ino

Fixed and tested.

TBAMax:
Hi. @pazi88 Can  you be more specific: Are the same settings used for decel that are used in AE(acceleration enrichminet)? 
I mean when I have AE already tuned, can I just activate it and choose the %of reverse AE?

Also as on note here, the other reason for rich condition during reducing throttle is the manifold referenced fuel pressure regulator having a some amount of lag, and that causes momentary increased relative pressure on the injectors (vacuum on the one side of the injectors is increased fast, but the fuel pressure from the fuel rail (on the other side) can not be released back to the tank fast enough).

pazi88:
> Hi. @pazi88 Can you be more specific: Are the same settings used for decel that are used in AE(acceleration enrichminet)? I mean when I have AE already tuned, can I just activate it and choose the %of reverse AE?
> 
> Also as on note here, the other reason for rich condition during reducing throttle is the manifold referenced fuel pressure regulator having a some amount of lag, and that causes momentary increased relative pressure on the injectors (vacuum on the one side of the injectors is increased fast, but the fuel pressure from the fuel rail (on the other side) can not be released back to the tank fast enough).

The settings to trigger AE are also used to trigger decel (dot throshold and min cahnge). And then the accel time is used as decel time too. The actual amount is then just fixed amount that you set in GUI. No relation to AE curve. This is how other aftermarket ecus seem to do this also so I used same method.

Personally I just set the decel fuel amount to 90% and it has been good value for pretty much every setup.

iltHeo84:
Hi @pazi88 and thank you for your work! I was actually planning to mod my fw to add this functionality, so I can't wait to see it merged!

PSIG-dev:
How are the desired results different than leaning/tuning hard decel cells or combining with DFCO, so transition or settle-point Lambda is near-target? _(traditional tuning solution)_

pazi88:
> How are the desired results different than leaning/tuning hard decel cells or combining with DFCO, so transition or settle-point Lambda is near-target? _(traditional tuning solution)_

Well same as with AE, the even though you tune VE map perfectly, AE is still required.  This is to compensate fast changes for short amount of time. Even with VE leaning, the map value (with averaging) is too slow to compensate the effect that this PR is battling. Speeduino is currently only ecu I know missing this feature and you can clearly see in it logs and even hear when stabbing throttle quickly. The rich spike causes small spits on the exhaust when lifting throttle.

speeduino/corrections.ino
```diff
@@ -333,128 +328,150 @@ uint16_t correctionAccel()
...
         currentStatus.mapDOT = 0;
       }
       else
       {
         //If MAE isn't currently turned on, need to check whether it needs to be turned on
-        if (MAP_rateOfChange > configPage2.maeThresh)
+        if (abs(currentStatus.mapDOT) > configPage2.maeThresh)
         {
-          BIT_SET(currentStatus.engine, BIT_ENGINE_ACC); //Mark acceleration enrichment as active.
           activateMAPDOT = currentStatus.mapDOT;
```
noisymime, line 358: Does this need an `abs()` wrapper around `currentStatus.mapDOT`?

noisymime:
Everything tests out fairly well from what I can see. I just added 1 small comment around `acticateMAPDOT` variable. 

Just as a consideration as this came up recently with DFCO, should we be disabling closed loop AFR control when the ACC or DCC flags are set?

pazi88:
> Everything tests out fairly well from what I can see. I just added 1 small comment around `activateMAPDOT` variable.
> 
> Just as a consideration as this came up recently with DFCO, should we be disabling closed loop AFR control when the ACC or DCC flags are set?

Well in ideal case the ACC and DCC corrections are tuned so that AFR doesn't really change from target when throttle moves. But we don't live in perfect world, so disabling closed loop AFR control during ACC and DCC would be beneficial. Only problem is that the effects on AFR are usually seen with small delay from ACC/DCC events. So there would need to be some delay to activate closed loop AFR control after the event.

noisymime:
Yeah, that's probably a side track discussion anyway. 

Are you able to confirm one way or the other around that `activateMAPDOT` comment? I'm happy to pull this in, just wanted to confirm on that one.

iltHeo84:
@pazi88 @noisymime How would such a pull affect the Secondary Serial data stream?
I have a little OLED that reads from Serial3 and it handles TPSdot as an unsigned byte (and multiplies its value by 10), what will change after this?

pazi88:
> Yeah, that's probably a side track discussion anyway.
> 
> Are you able to confirm one way or the other around that `activateMAPDOT` comment? I'm happy to pull this in, just wanted to confirm on that one.

I commented to that already that yes it needs that. Mistake on my part. You don't see that comment?

pazi88:
> @pazi88 @noisymime How would such a pull affect the Secondary Serial data stream? I have a little OLED that reads from Serial3 and it handles TPSdot as an unsigned byte (and multiplies its value by 10), what will change after this?

It should work as before.

iltHeo84:
> > @pazi88 @noisymime How would such a pull affect the Secondary Serial data stream? I have a little OLED that reads from Serial3 and it handles TPSdot as an unsigned byte (and multiplies its value by 10), what will change after this?
> 
> It should work as before.
So the cast to uint8_t will only give positive values? Or range from -128 to 127?
In the latter case, there could be a potential problem with the maximum theoretical value of TPSdot: assuming one could slam the throttle open from 0 to 100% in one single sample, it this would give a TPSdot value of 200 * 30 / 2 = 3000 %/s
It would be sufficient to open 50% of throttle in 1/30th of a second for it to overflow.
Please correct me if I am wrong, I’m not really familiar with casting variables.

noisymime:
> > Yeah, that's probably a side track discussion anyway.
> > Are you able to confirm one way or the other around that `activateMAPDOT` comment? I'm happy to pull this in, just wanted to confirm on that one.
> 
> I commented to that already that yes it needs that. Mistake on my part. You don't see that comment?

Weird, I can't see the comment or any changes to that variable 🤷

pazi88:
> Weird, I can't see the comment or any changes to that variable 🤷

ugh, sorry, forgot to push the change. But the comment shows up like this for me:
image

noisymime:
No idea why the comment shows as 'Pending', it still doesn't appear for me. All good though and PR merged with that last commit.

PR #941, opened 2022-10-13
Stuck ego fix

jonbills:
Ego control gets stuck at 100% when ignitionCount wraps during an DFCO event because the test "if (ignitionCount < AFRnextCycle).." continues to fail until ignitionCount catches up with AFRnextCycle again -  long time.
This fix also checks for ignitionCount < (AFRnextCycle - configPage6.egoCount)

### DIFF
--- a/speeduino/corrections.ino
+++ b/speeduino/corrections.ino
@@ -609,7 +609,7 @@ byte correctionAFRClosedLoop()
   {
     AFRValue = currentStatus.egoCorrection; //Need to record this here, just to make sure the correction stays 'on' even if the nextCycle count isn't ready
     
-    if(ignitionCount >= AFRnextCycle)
+    if((ignitionCount >= AFRnextCycle) || (ignitionCount < (AFRnextCycle - configPage6.egoCount)))
     {
       AFRnextCycle = ignitionCount + configPage6.egoCount; //Set the target ignition event for the next calculation
         


jonbills:
[datalog_2022-10-13_16-24-33.msl.zip](https://github.com/noisymime/speeduino/files/9783958/datalog_2022-10-13_16-24-33.msl.zip)
Attached is the log I used to 'diagnose' this.  I made a special build that logged a 'reason' code to VE2 for when AFRvalue would be 100% unexpectedly. VE2 value of 41 (in this log) occurs when `ignitionCount < (AFRnextCycle - configPage6.egoCount) `

VitorBoss:
The fix seems simple, but why you reuse a branch or based your changes on an used one? 9 commits for a single line of change seems a lot.

Suggestion:
- Save the code you changed on a separated file
- Open the speeduino github folder and open a shell there
- Enter the command `git fetch upstream`
- Enter the command `git reset --hard upstream/master`
- Apply your code changes and commit it
- Go back to the shell and enter the command `git push --force-with-lease origin`
That will sync your branch with master and the commit will be the single one ;)

jonbills:
thanks @VitorBoss , I couldn't complete that sequence - failed at git push --force-with-lease origin with some authentication problems. I guess I need to setup some ssh keys. 
fwiw, it got to this state because I made a previous fix in my master, realised my error and backed the changes out.  so all but one of those commits are are a no-op in aggregate. 
maybe I should just delete my fork and start again.

PR #942, opened 2022-10-14
stuck ego fix

jonbills:
Ego control gets stuck at 100% when ignitionCount wraps during an DFCO event because the test "if (ignitionCount < AFRnextCycle).." continues to fail until ignitionCount catches up with AFRnextCycle again - long time.
The fix adds a check for ignitionCount < (AFRnextCycle - configPage6.egoCount)

### DIFF
--- a/speeduino/corrections.ino
+++ b/speeduino/corrections.ino
@@ -609,7 +609,7 @@ byte correctionAFRClosedLoop()
   {
     AFRValue = currentStatus.egoCorrection; //Need to record this here, just to make sure the correction stays 'on' even if the nextCycle count isn't ready
     
-    if(ignitionCount >= AFRnextCycle)
+    if((ignitionCount >= AFRnextCycle) || (ignitionCount < (AFRnextCycle - configPage6.egoCount)))
     {
       AFRnextCycle = ignitionCount + configPage6.egoCount; //Set the target ignition event for the next calculation
         


Tjeerdie:
Hmm indeed this bug can be triggered in two ways. The first is a very unlikely event but it can still happen. The second one is more realistic to happen. But still a rare event

1. When `ignitionCount` rolls over while `AFRnextCycle` is very close to rollover. Then if `correctionAFRClosedLoop() `is not called in the headroom between `AFRnextCycle` and `ignitionCount` rollover. (if this is for example 1 ignition count this could happen)
2. When DFCO is active in the headroom between `AFRnextCycle` and `ignitionCount` rollover before and prevents `if(ignitionCount >= AFRnextCycle)` is evaluated. 

The alternative fix is to explicitly roll ignition count over at a lower number inside the `if(ignitionCount >= AFRnextCycle)`. This ensures there is always enough ignition events headroom between `AFRnextCycle` and `ignitionCount` rollover.  

```
    if(ignitionCount >= AFRnextCycle)
    {
      ignitionCount &= 0x3FFF;
      AFRnextCycle = ignitionCount + configPage6.egoCount; //Set the target ignition event for the next calculation
```

Anyway I am happy you found it and important to fix this.

jonbills:
Thanks for the confirmation @Tjeerdie  -  I think the 2nd case I have quite regularly  (maybe 1/3 days that I use the car) - when I'm driving in traffic,  I think DFCO may be on as much as 30% of the time, so that will give a 30% chance of this problem happening when ignitionCount rolls over, and it seems to roll over every 30 minutes maybe?

PR #943, opened 2022-10-14
noise filtering for Jeep2000 decoder cam signal

jonbills:
check that we're at an acceptable crank tooth before acting on cam trigger pulse.

### DIFF
--- a/speeduino/decoders.ino
+++ b/speeduino/decoders.ino
@@ -1818,7 +1818,10 @@ void triggerPri_Jeep2000()
 }
 void triggerSec_Jeep2000()
 {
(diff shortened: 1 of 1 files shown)


noisymime:
Just to check, has this been tested on an enigne? The PR seems OK, but I just want to make sure it's not falling into the trap of only ever looking for the cam signal at exactly where it's expected to be found. I think it's probably safe, but just wanted to confirm this has been tested.

jonbills:
I believe @jaky 47 is using it- I've asked him to confirm here.

JaredSellers4:
@jonbills, PR913 had this "cheap filter" AND a gap comparison.  Why did you take out the gap comparison?  I am trying to update Jeep 4cyl to match latest noisymime master and want to include whatever you recommend/have success with regarding filtering for trigger patterns with varying gaps angles..

PR #944, opened 2022-10-15
Update decoders.ino

rafolg:
Fix an issue where the current logic cause erroneous crankAngles for ignition by tooth whether the trigger angle is negative or positive.

Moving the ignitionLimits() to inside the if statement after the 360 deg addition to first revolution check as well as inside the else statement before calling checkPerToothTiming() fixes the issue.


noisymime:

PR #954, opened 2022-10-27
fix check for engine rpm below target

jonbills:
In correctionIdleAdvance(),  idleAdvActive is false until the engine RPM is > (target RPM - 200) , which means that on start up, you have to manually (with your foot!) raise the RPM to the target speed before idle advance control engages.
Which is really annoying if you want to be able to just start the car with the key and have it run. 

so I changed it so it is consistent with the comment "// Active only after the engine is 200 RPM below target on first time"  which works now for me.

### DIFF
--- a/speeduino/corrections.ino
+++ b/speeduino/corrections.ino
@@ -783,8 +783,17 @@ int8_t correctionIdleAdvance(int8_t advance)
     else { idleAdvTaper = 0; }
   }
 
-  if ( !idleAdvActive && BIT_CHECK(currentStatus.engine, BIT_ENGINE_RUN) && (currentStatus.RPM < (((uint16_t)currentStatus.CLIdleTarget * 10) - (uint16_t)IGN_IDLE_THRESHOLD)) ) { idleAdvActive = true; } //Active only after the engine is 200 RPM below target on first time
-  else if (idleAdvActive && !BIT_CHECK(currentStatus.engine, BIT_ENGINE_RUN)) { idleAdvActive = false; } //Clear flag if engine isn't running anymore
+/* When Idle advance is the only idle speed control mechanism, activate as soon as not cranking. 
+When some other mechanism is also present, wait until the engine is no more than 200 RPM below idle target speed on first time
+*/
+
+  if ((!idleAdvActive && BIT_CHECK(currentStatus.engine, BIT_ENGINE_RUN)) &&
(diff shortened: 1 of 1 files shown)


wonder99:
I think it's actually correct the way it was--the idea is not to engage the advance until the RPM is above (target-200).  The word 'below' in the comment refers to the "-200", and the intent is only engage ignition advance after the RPM has gone above that.  So essentially, don't engage the advance algorithm until you're with 200 RPM of your target.

jonbills:
@wonder99 thanks for reviewing.  Why would we want it to not attempt to raise the idle to target speed on startup?
also, the opposite seems a good question too - why would we want it to not attempt to bring the idle speed down if it is too fast on startup (as it could be with my PR)?
the more I think about it, the more I think all the logic with idleAdvActive is unhelpful and should be removed and just let idle advance control work straight away on startup. I'm probably missing something!

VitorBoss:
There are some scenarios where the idle advance must be disabled, one is on low temperature long cranking, another is on gear shifts.

If you need to push the TPS to get to target you might need to increase the idle valve % to get some overshoot, that will avoid the RPM to sink when idle control isn't running.

AlessandroEmm:
@VitorBoss thats not easily possible for people on ITBs (@jonbills and me), hence our reliance on Advance Idle Control.

jonbills:
I think the long cranking case is covered by checking for the engine being in 'RUNNING' state - and gear change could be covered by the activation delay timer?
I don't think the one-time check for being above (target-200) helps with either case, does it?

VitorBoss:
I talked about the delay because you sayd to remove all the activation code.

About the `RUNNING` flag, the way it currently works in speeduino is just 1 RPM above `CRANKING`, that is tricky because sometimes the RPM oscillates on that region before truly starts, that will trigger the idle advance and make the advance very high, making the engine stalls. That is the main reason I made that code, it was happening a lot here, making the cranking vey long and "heavy" on the starter.

Maybe is better to include an activation bit for this, or compare the idle valve type, if no idle valve are in use enable it right away.

jonbills:
I've revised the PR in line with @VitorBoss suggestion. 
The logic is now as it was in the cases where configpage6.iacAlgorithm  is not zero, but where it is zero, idle advance is activated right away. 
Vitor, please could you review again?

PR #958, opened 2022-10-29
Lower max on IAT and CLT gauges for celsius

DeionSi:
Lowers the maximum for Celsius on the IAT and CLT gauges in TunerStudio to a more reasonable value.


DeionSi:
(and fuel temp)

jonbills:
I think 110 is too low for CLT on some modern cars, e. g. I had a mini where the fan didn't cut in until 115.

DeionSi:
It seems some Minis have 112c for the high fan. Change both Celsius and Fahrenheit to max 120c/248f.

brunob45:
Make it configurable in TS like the rpm gauge?

noisymime:
> Make it configurable in TS like the rpm gauge?

I think this would be a good addition if anyone feels like adding the local setting. For now I'll pull this in though as it is an improvement over what's there.

Issue #960, opened 2022-10-31
Clutch input is floating when set to pullup


shiznit304:
Firmware 202207.1 and master as of 10/31/2022 have a bug where Launch Control feature isn't setting the clutch input pin to use internal pullup resistor

Steps to reproduce:

1. Flash 202207.1 to Mega2560 using Speedyloader
2. Create new project and use 202207.1.ini definition file
3. Load speeduino generic base tune
4. Set board version to UA4C
5. Enable launch control, clutch input board default, low, pullup
6. Configure Auxillary input config to digital, pin 37
7. Change any Tunerstudio gauge to show your configured auxiliary input channel
8. Power cycle board
9. Watch the gauge float between 0 and 1

shiznit304:
Performed the same test on a few other pins (30, 53) and they all are stuck floating. It seems to work when I switch the the Aux input for the pin to Off. I have a dashboard for troubleshooting that uses the aux input channel config feature to monitor the pin status. Seems like whenever I monitor the pin, it goes floating.

PR #961, opened 2022-11-02
Fix for Issue #960

shiznit304:
Fix for issue #960 . Switching this line of code allows me to use Auxiliary input channel to "monitor" the status of a digital pin while the pin is configured for a feature (launch control pin, WMI, etc)


pazi88:
This isn't really fix for anything. Just introduces back the old bug.

shiznit304:
Perhaps so, but I get the results I expect from it. If anyone else needs
the expected behavior they can try this workaround.

VitorBoss:
If you got the results you expect from a bug maybe you shouldn't

PR #965, opened 2022-11-14
Faster Flash writes, both on die and SPI

VitorBoss:
On die writes got a small change 38 to 34us, SPI flash got a big improvement, average byte write changed from 144 to 93us.

Tested on STM32F407VE board on both cases.
Need to be tested on Teensy.

[FlashAsEEPROMTest.zip](https://github.com/noisymime/speeduino/files/9998938/FlashAsEEPROMTest.zip)
This is @Tjeerdie project used to test the speed, **_this will wipe all the eeprom/flash content!_** **Test on an unused board**


DeionSi:
I get a few compiler warnings with this when compiling for mega. I know this doesn't apply to mega but I guess it shouldn't generate warnings for it.

PR #967, opened 2022-11-29
MISRA check: add XML output

adbancroft:
### What
Add a -x (or--xml) option to the MISRA script so it will write the output as an XML file.

### Why
It's useful to have the MISRA script output a machine readable format. Since cppcheck supports XML output, XML was the obvious choice. This can easily be converted to other formats. E.g. JSON with xmltojson.


noisymime (commented):
Looks good to me. I'm assuming no changes are needed to any of the automated scanning as this is only an optional output?

adbancroft:
> Looks good to me. I'm assuming no changes are needed to any of the automated scanning as this is only an optional output?

Correct - no change to Github actions. Although I'm struggling with line endings (CR vs CR LF) - might be good to add a call to dos2unix?

adbancroft:
@noisymime - latest commit found close to 1000 additional type mismatch violations: the 'byte' type wasn't visible to cppcheck & it's unsigned on AVR. Which has flushed out many C2012-10.4 violations ("The target of an operation must be of an appropriate type. Eg No implicit conversion of signed/unsigned")

Good or bad - I'll let you choose ;-)

noisymime:
10.4 is a hell of a requirement. There's already a stack of those that need resolving but ultimately it's a fair requirement. Comparing sign vs unsigned can be problematic, although typically only when they're of the same size. If they're real violations, they should at least be counted. 

I was going to pull this in the other day but then noticed the changes that were added with c65734abef96f546e06b1ece6d33fbddda0a43d6 . Is that all still 'safe' for the automated CI?

adbancroft:
> Is that all still 'safe' for the automated CI?

Should be - it's still just running cppcheck as it always has, only with a much simpler script.

I'm running it using:
1. WSL 2 - Ubuntu 22.04, cppcheck 2.7, python 3.10.6.
2. Git Bash - Windows 10 x64, cppcheck 2.9, python 3.11.0

I just pushed a couple of changes that will help the automated CI:
1. Add a "dos2unix" step to the GitHub action: vscode keeps converting the line endings, so this is "just in case"
2. Restored the rename steps for the libdivide folder: on Git Bash, the scan was taking forever (though it ran quickly on Ubuntu under WSL).

So we should be good to go now.

adbancroft:
@noisymime I noticed the GitHub action failed on master - the fix is to add `sudo apt-get install dos2unix` to misra.yml. I can put in another PR if that's the easiest solution?

PR #970, opened 2022-12-17
Race conditions by direct register access pin manipulations

Tjeerdie:
There is some serious race condition issue in the speeduino codebase. There are a lot of potential race conditions! I never had encountered issues with this before but now i have run into one with the fuel pump and noticed there are many many more! (that could explain my random ignition issues)

For a short explanation what this is and why it is and issue see youtube video: [Race conditions](https://www.youtube.com/watch?v=JjnyYf4BIco)

**Introduction**
When developing for my new STM32F411 board I noticed some rich spikes every 5 minutes or so. When testing on the bench I found indeed the same issue. 

Below you see the logic analyzer output showing one of the glitches on the injection channel. The issue appeared on both the ignition and injection channels. And more importantly the over-dwell protection does not catch it! So it is a potential serious issue. 

RaceCondition

A zoomed in version shows exactly what is going on.
RaceConditionZoomedin

I traced it down to the fuel pump pin being on the same port as the injector and ignition pins.  Where a noisy battery voltage adc (unrelated issue) makes that it wants to prime the the fuel pump very often. Causing this issue to show its head now. But even with a noisy vbat reading this problem should never be possible to occur. And more importantly there are a host of other pins that are manipulated the same way. That can have the same issue! basically all pin manipulation suffer from this potential issue, because there is no guards against this very well known race condition.  

**Problem description**
A race condition occurs when asynchronous code access the same variables non atomically.  So when interrupt routines and main loop code both interact with the same variables that take more than 1 instruction cycle to handle (non atomic). 

For speeduino the way the MCU pins of injectors, fan etc are manipulated  The same port register is used within main loop and interrupt routines. (If the FAN or BOOST and INJECTOR pin share the same MCU port for example). This can cause race conditions.

_Example:_
Example prerequisites. Fuel pump pin and injector1 pin are both on PORTB

If the main loop tries to write the FUEL pin high the following code steps are executed:
1a. Read current register status of PORTB
2a. Mask the current port status with the FUEL pin mask
3a. Write the masked value to register PORTB

If now this routine is interrupted by the fuel schedule interrupt after point 2 a race conditions occurs.  Because in the interrupt routine tries to write INJ1 pin HIGH:

1b. Read current register status of PORTB
2b. Mask the current port status with the INJ1 pin mask
3b. Write the masked value to register PORTB

So after the interrupt routine is done The code of the main loop continuous. But it will write the OLD register PORTB value, because it has read the value before the injector interrupt occurred. The same for IGN and IDLE PWM pins or any other combination of pins. 

**Solution**
The easy solution is to disable interrupts before and enable after all direct register manipulations in speeduino.

I tried to find all of them and enclosed them in noInterrupt(); CODE interrupt(); That is included in this PR. This might impact performance a bit. especially on the Mega. So proper timing tests on the Mega are advised. The impact on the faste MCU should be really neglect able.    

The best solution would be to manipulate the ports only from one routine (one interrupt, or only the main loop code). But that would require a whole concept change for speeduino where the code only sets flags and there is only one routine that reads the flags and actually manipulates the port registers. But this has al lot of potential timing issues, and needs a lot of changes to the code! So the quick and dirty one should suffice. 

**Further actions**
_1. testing_
I did test this with my stm32F411 on the bench  and it solves the issue with the fuel pump, and has no direct visible side effects. But that is not really exhaustive testing. So more people should test this on all platforms. Especially the arduino Mega as this has a potential influence on main loop speed. 

_2. fuel and ignition schedule_
This fix should also include the fuel and ignition schedules. So at the start of a port manipulation a noInterrupt(); should be added and closed by a interrupt(); again after its done. Alternatively the whole interrupt routine can be encased in a  

noInterrupt(); 
INTERRUPT FUNCTION
interrupt(); 

edit: I added some changes to that should fix it for the injection and ignition too. But proper testing is still required!


pazi88:
I have encountered similar issues with some ports on mega2560 speeduino too. I have never been able to trace down where the issues come from and I have complaimed about that in slack too. Bit this seems plausible explanation for that. Should this fix for mega2560 too, or only stm32 (don't have time to check the changes now).

ric355:
Obviously the fuel pump pin is only pulled low when the engine is first started, and after that it is not changed, so any race condition with the injector outputs is not really a significant problem except in a scenario where you have a system fault causing that pin to be toggled while the engine is running.  Do you have any better examples of where there is a significant issue in normal running?   I have in the past had unexplained hesitations which might have had this sort of scenario as their root cause as they a rare, so it isn't a bad tune.

Tjeerdie:
> Obviously the fuel pump pin is only pulled low when the engine is first started, and after that it is not changed, so any race condition with the injector outputs is not really a significant problem except in a scenario where you have a system fault causing that pin to be toggled while the engine is running. Do you have any better examples of where there is a significant issue in normal running? I have in the past had unexplained hesitations which might have had this sort of scenario as their root cause. They a very rare, so it isn't a bad tune.

Sorry i disagree, This should never be possible to happen. potential race conditions are usually rare events but should be eliminated by design and not by change. 

Even if the it can happen only 1 in a 100.000km event it must be fixed. The other pins like AIRCO, FAN, VVT, BOOST etc have similar potential issues, but i never observed them. But these would not show up very regularly as it is a combination of circumstances

Tjeerdie:
> I have encountered similar issues with some ports on mega2560 speeduino too. I have never been able to trace down where the issues come from and I have complaimed about that in slack too. Bit this seems plausible explanation for that. Should this fix for mega2560 too, or only stm32 (don't have time to check the changes now).

This is a general and very well known programming issue in asynchronous code (Like interrupt code and threading). The proposed fix works for all platforms. inclusing Mega

ric355:
> > Obviously the fuel pump pin is only pulled low when the engine is first started, and after that it is not changed, so any race condition with the injector outputs is not really a significant problem except in a scenario where you have a system fault causing that pin to be toggled while the engine is running. Do you have any better examples of where there is a significant issue in normal running? I have in the past had unexplained hesitations which might have had this sort of scenario as their root cause. They a very rare, so it isn't a bad tune.
> 
> Sorry i disagree, This should never be possible to happen. potential race conditions are usually rare events but should be eliminated by design and not by change.
> 
> Even if the it can happen only 1 in a 100.000km event it must be fixed. The other pins like AIRCO, FAN, VVT, BOOST etc have similar potential issues, but i never observed them. But these would not show up very regularly as it is a combination of circumstances

You really need to read what I write more carefully.

I did not say it should not be addressed. Nor did I say it should never happen or that your fixes were somehow inappropriate. I am well aware of what a race condition is and how they can be extremely bad in software.

I asked you if you had any "better" examples - by which I mean things that are _more likely_ to happen in a system that doesn't have a battery adc fault that causes the fuel pump to be switched off and on while the engine is running. I asked this because I am interested in understanding if events I have had in the past could be explained by a similar race condition between some other pins.

VitorBoss:
@ric355 Fuel pump also switch off when a lost sync happens

Tjeerdie:
> > > Obviously the fuel pump pin is only pulled low when the engine is first started, and after that it is not changed, so any race condition with the injector outputs is not really a significant problem except in a scenario where you have a system fault causing that pin to be toggled while the engine is running. Do you have any better examples of where there is a significant issue in normal running? I have in the past had unexplained hesitations which might have had this sort of scenario as their root cause. They a very rare, so it isn't a bad tune.
> > 
> > 
> > Sorry i disagree, This should never be possible to happen. potential race conditions are usually rare events but should be eliminated by design and not by change.
> > Even if the it can happen only 1 in a 100.000km event it must be fixed. The other pins like AIRCO, FAN, VVT, BOOST etc have similar potential issues, but i never observed them. But these would not show up very regularly as it is a combination of circumstances
> 
> You really need to read what I write more carefully.
> 
> I did not say it should not be addressed. Nor did I say it should never happen or that your fixes were somehow inappropriate. I am well aware of what a race condition is and how they can be extremely bad in software.
> 
> I asked you if you had any "better" examples - by which I mean things that are _more likely_ to happen in a system that doesn't have a battery adc fault that causes the fuel pump to be switched off and on while the engine is running. I asked this because I am interested in understanding if events I have had in the past could be explained by a similar race condition between some other pins.

The fuel pump pin is the one that only hinted me to the issue. That is why i described it. I thought you mentioned it not being a problem because the fuel pin is only toggled when engine is not running. (So i wrongfully assumed you think it is no potential issue?). 

For your info it seems the speeduino code tries to start a already running fuel pump more often than only when a tooth is seen, also when running. But that's not really an issue because the fuel pump should already be running turning it on again should not give an issue. (same logic, sorry :-D)

Anyway if you read my post in more detail it should be clear that not only the fuel pin has the potential issue. There are a lot more of them. Basically all pins that are set by the code have this potential issue. So yes issues in the past could very well be explained by this.

For example:
BOOST
FAN
VVT1
VVT2
IDLE
N2O
AIRCON
AIRCON FAN

Tjeerdie:
Ow and to be clear the injector pins can influence the ignition pins in the same way if these are on the same port. Both run in there own schedule and thus interrupt. But i did not make a fix for that yet. See the last paragraph with further actions.

pazi88:
Ok. I'll try test this today if this fixes the issue or not. I have issues with few ign pins, but also pwn idle pin issues (same port). The idle pwm issue is much easier to reproduce, so it should be easy to see if this helps.

rafolg:
Is this race issue isolated to bit banging?
Having my speeduino on Teensy platform where Josh modified the auxiliaries.h to use digitalWrite commands instead of bit banging have not seen any problems.

I have extensively tested with scope all my auxiliaries running at the same time and have not seen any issues at all. Not in IGN or fuel either.

Tjeerdie:
> Ok. I'll try test this today if this fixes the issue or not. I have issues with few ign pins, but also pwn idle pin issues (same port). The idle pwm issue is much easier to reproduce, so it should be easy to see if this helps.

Do have a logic analyzer screenshot or can you tell that the issue is exactly the same from my screen shot? The issue will show it self always the same way. When shown with logic analyzer that is not capturing fast enough, it will show as if the interrupt for the falling edge is missed. So the pulse will always extend until the next falling edge. 

If it is capturing fast enough you also will have the very short time the outputs will be low and then become high again. Both are always present if this is your issue. Let me know if this fixes it or not.

pazi88:
> > Ok. I'll try test this today if this fixes the issue or not. I have issues with few ign pins, but also pwn idle pin issues (same port). The idle pwm issue is much easier to reproduce, so it should be easy to see if this helps.
> 
> Do have a logic analyzer screenshot or can you tell that the issue is exactly the same from my screen shot? The issue will show it self always the same way. When shown with logic analyzer that is not capturing fast enough, it will show as if the interrupt for the falling edge is missed. So the pulse will always extend until the next falling edge.
> 
> If it is capturing fast enough you also will have the very short time the outputs will be low and then become high again. Both are always present if this is your issue. Let me know if this fixes it or not.

I have scope to show/see that. It shows up excatly like you described. The pwm signal for the idle valve is sometimes missing the falling edge for one of the two outputs and it makes the idle valve to click fully open/closed and engine makes really short stumbles randomly because of that.

Same does happen for ign outputs on the other design, but it's way more rare and pretty much impossible to catch on scope.

Tjeerdie:
> Is this race issue isolated to bit banging? Having my speeduino on Teensy platform where Josh modified the auxiliaries.h to use digitalWrite commands instead of bit banging have not seen any problems.
> 
> I have extensively tested with scope all my auxiliaries running at the same time and have not seen any issues at all. Not in IGN or fuel either.

It is unrelated to the MC33810 (if that is what you mean by bit banging). The chip select line of the uses MC33810 direct port register manipulation so that one might be affected. 

Anyway i had a bad case and even then it only showed up once every 5 or so minutes. Where one pulse is much longer. It is really hard to pick the offending one out of a long list of perfect ones.

pazi88:
Ok, I tested this. On current master and as much I can remember, the core4 idle PWM outputs have looked like this. Every 5-10 seconds I get missing pulses, and I can hear the idle valve to click fully open/closed:
image

I tried with this PR and yeah. The idle PWM outputs were perfect for 15 minutes that I watched this run on my bench setup. No single missing pulse on scope and didn't hear the idle valve click for even once.
image

So yeah, this fixes the issue that I have been trying to hunt down for long time.

Tjeerdie:
> 

Nice!

noisymime:
Looks good! It's not something I've come across (or at least identified if I have) but it completely makes sense. 

My only comment would be that, from a style ans simplicity perspective, it would be good to include the wrappers all on the `#defines`, including the ones with ternary operations. I'm not sure if there's actually any syntax that allows this within a macro though?

Tjeerdie:
Well when we want it to be the same for all, so also the ones here we use the ternary operators its indeed more tricky. The ternary operators do not allow multi line/command expanding macro's for obvious reasons. So these should then be written as full if then else statements to work. 

Does not matter for me what ever you prefer :+1:

rafolg:
Can it not be:
#define x_ON()        {noInterrupts(); ternary; interrupts();}

Tjeerdie:
> Can it not be: #define x_ON() {noInterrupts(); (ternary); interrupts();}
That does not work in ternary operators. It only allows expressions, no statements.

Tjeerdie:
I changed everything to be more consistent. Only ignition outputs are still not consistent with the rest. Ignition is a lot harder to do, it makes for a lot more changes in the code. Because ignition uses uses define macro functions.

**digitalWrite()**
I did a little more research how the `digitalWrite()` function in the three main Arduino cores are implemented. There is a couple of places in the speeduino code where the `digitalWrite() `is used instead if direct port manipulation. 

_Arduino mega:_	The mega disables interrupts before  writing to the registers, so its safe
_Teensy:_		Uses special hardware registers so the pinwrites are atomic and safe.
_STM32:_	        It is not guarded against the race condition at all. 

Some of these `DigitalWrites()` i already enclosed in the interrupt start/stop to make it less of an issue. Some are still because its only an issue for STM32. Not sure how to handle this best.  

**So the STM32 still possibly suffers from this race conditions when using: (These functions use digitalWrite() )**
Stepper idle control 
Idle Up control
Ignition Bypass
WMI Indicator
Reset Control
Any of the programmable outputs.

pazi88:
Reset control isn't at least used in stm32

Tjeerdie:
> Reset control isn't at least used in stm32

I understand, just wanted to list all digitalWrite() usages that i could find.

Tjeerdie:
> > Reset control isn't at least used in stm32
> 
> I understand, just wanted to list all digitalWrite() usages that i could find.

I also no longer consider it an issue. digitalWrite() is safe for STM32 for the same reason it is for teensy. Atomic register access is used to set or reset port pins.

rafolg:
> Can it not be: #define x_ON() {noInterrupts(); (ternary); interrupts();}
> That does not work in ternary operators. It only allows expressions, no statements.

I’m not suggesting to add the instructions inside the ternary but before and after.

For example inside Auxiliaries.h line 66 there is:
#define AIRCON_ON() with ternary wrapped in brackets and a BIT_SET() instruction at the end outside of the ternary.

pazi88:
Tried to run this on my car with mega2560. The idle PWM signal looks perfect on scope on various test conditions. But while driving I get quite much stumbling. So it might be negatively affecting inj/ign outputs, which needs to be checked.

mike501:
I'm not suggesting this doesn't need fixing and much qdos to finding and
getting this far.

Just wanted to ask if we're going about this technically the correct but
perhaps not ideal way for the mega? My concern is ignition being very time
sensitive and potential impact. I see the timing impacted by 0 to 8us
wobble (acceptable) to 20 to 30us for a couple of ignition events and then
recover. Assuming what I see isn't this (I'm on a test rig with everything
like idle control turned off). I'm concerned slowing the code down would
have a big size effect. Would it make more sense for ignition to remove
everything that could interrupt it and leave it 'free' as now but in its
own dedicated interrupt?

Tjeerdie:
> I'm not suggesting this doesn't need fixing and much qdos to finding and getting this far. Just wanted to ask if we're going about this technically the correct but perhaps not ideal way for the mega? My concern is ignition being very time sensitive and potential impact. I see the timing impacted by 0 to 8us wobble (acceptable) to 20 to 30us for a couple of ignition events and then recover. Assuming what I see isn't this (I'm on a test rig with everything like idle control turned off). I'm concerned slowing the code down would have a big size effect. Would it make more sense for ignition to remove everything that could interrupt it and leave it 'free' as now but in its own dedicated interrupt?
> 

Ignition is already in a dedicated interrupt (just like everything else). Also it is already highest priority i guess? Anyway there is some other issue with it. 

If you look closely to the issue description you can see that the problem is not that the ignition is interrupted. But it can be that the PWM is interrupted by IGNITON. So the interrupt is not stopped to actually save the interrupt where you add the condition. It is to save ALL other interrupts from the that specific interrupt. 

Aka if you remove the 
`{noInterrupts(); Register manipulation; interrupts();}`
from the ignition interrupts you make it so that the ignition interrupt can actually intervene and create the problem for all others (INJ, IDLE_PWM, etc). So you do solve long pulses only for ignition and make it so that ignition can create long pulses with all others. 

In short the only real solution is to fix ALL interrupts to get rid of the problem for once and for all. 
as far as i can understand and see the _noInterrupts();_ and _interrupts();_ are one clock cycle operations on the Mega, so i guess that something else is gone wrong than the timing. I just need some time to figure out what it is.

Tjeerdie:
Ok more info, a new strategy is needed.  
1. Mega has probably no nested interrupts. So only disable interrupts for direct port manipulation in the mainloop is enough.
3. STM32/Teensy have nested interrupts. So a different solution is needed. digitalWrite() on both platforms is interrupt safe and can be used. (It is already used by Teensy). Maybe using the same atomic set/reset registers as used by digitalWrite() is an option, but needs platform specific code all over the place. And that is also no very desirable. 

More to come.

noisymime:
> Mega has probably no nested interrupts. So only disable interrupts for direct port manipulation in the mainloop is enough.

It's a bit of a poorly understood area on the AVRs, but the short of it is that interrupts are disabled by default whenever an ISR is entered. You can manually re-enable interrupts within the ISR, but it's asking for problems and not done anywhere within the Speeduino code. If an interrupt occurs whilst another ISR is already in progress, its flag will be set and its own ISR called once the in-progress one finishes. 
As a result of this, you can never have a situation where one of the PWMs that use these macros (Eg `BOOST_PIN_HIGH()`) would be impacted. Something like the `FUEL_PUMP_ON()` however that is called from the mainloop can be though. 

The solution here might be as simple as swapping out the macros for low speed IO (Such as the fuel pump) to use `digitalWrite` instead?

Tjeerdie:
> > Mega has probably no nested interrupts. So only disable interrupts for direct port manipulation in the mainloop is enough.
> 
> It's a bit of a poorly understood area on the AVRs, but the short of it is that interrupts are disabled by default whenever an ISR is entered. You can manually re-enable interrupts within the ISR, but it's asking for problems and not done anywhere within the Speeduino code. If an interrupt occurs whilst another ISR is already in progress, its flag will be set and its own ISR called once the in-progress one finishes. As a result of this, you can never have a situation where one of the PWMs that use these macros (Eg `BOOST_PIN_HIGH()`) would be impacted. Something like the `FUEL_PUMP_ON()` however that is called from the mainloop can be though.
> 
> The solution here might be as simple as swapping out the macros for low speed IO (Such as the fuel pump) to use `digitalWrite` instead of the direct port manipulation?

To solve the issue for the Mega all low speed stuff done in the main loop as digitalWrite seems reasonable. Just make sure to check there is not one PWMing in there. That would atleast make it so that everything is solved for the mega, and that we have no more mainloop stuff to worry about for the Teensy and STM32. 

But then we are left with the injector and igntion channels that need a solution on STM32. Teensy uses the MC33810 for those so there is no danger there. Only STM is still at risk.

For the aux channels/Boost and idle PWM etc. Just add STM32 to the teensy digitalWrite() macro is a solution.

Tjeerdie:
A new PR with a solution is created for this issue.

PR #971, opened 2022-12-19
Fix decel enleanment in AE Add -mode

pazi88:
This PR fixes the decel enleanment problems in AE Add -mode. Also seems that this will gain ~5 loops per sec on mega in AE Add -mode, because it's not calculating the AE amount on every loop.

### DIFF
--- a/speeduino/corrections.ino
+++ b/speeduino/corrections.ino
@@ -84,9 +84,9 @@ uint16_t correctionsFuel(void)
   if (result != 100) { sumCorrections = div100(sumCorrections * result); }
 
   currentStatus.AEamount = correctionAccel();
-  if (configPage2.aeApplyMode == AE_MODE_MULTIPLIER)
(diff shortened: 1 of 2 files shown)


MykkH:
I think this may need a bit more work, on a datalog the negative TPSdot values at 70% fuel commanded will max out injector duty cycle taking mine from 4ms-9ms up to 38ms of injector PW at 90% DC. Completely flooding my running engine, the opposite of it's intentions. lol

pazi88:
> I think this may need a bit more work, on a datalog the negative TPSdot values at 70% fuel commanded will max out injector duty cycle taking mine from 4ms-9ms up to 38ms of injector PW at 90% DC. Completely flooding my running engine, the opposite of it's intentions. lol

Yes, this is the reason for this PR.

MykkH:
How do we get this PR implemented in the master? Would really like to get this resolved. Thanks

PR #972, opened 2022-12-24
Fix race condition direct port manipulation

Tjeerdie:
This is a new PR based on PR#970. More info and discussion can be found in the closed PR#970. But for cleanliness a new branch and therefor PR is created.

================================================================================================

There is some serious race condition issue in the speeduino codebase. There are a lot of potential race conditions! I never had encountered issues with this before but now i have run into one with the fuel pump and noticed there are many many more! (that could explain my random ignition issues)

For a short explanation what this is and why it is and issue see youtube video: [Race conditions](https://www.youtube.com/watch?v=JjnyYf4BIco)

**Introduction**
When developing for my new STM32F411 board I noticed some rich spikes every 5 minutes or so. When testing on the bench I found indeed the same issue. 

Below you see the logic analyzer output showing one of the glitches on the injection channel. The issue appeared on both the ignition and injection channels. And more importantly the over-dwell protection does not catch it! So it is a potential serious issue. 

RaceCondition

A zoomed in version shows exactly what is going on.
RaceConditionZoomedin

I traced it down to the fuel pump pin being on the same port as the injector and ignition pins.  Where a noisy battery voltage adc (unrelated issue) makes that it wants to prime the the fuel pump very often. Causing this issue to show its head now. But even with a noisy vbat reading this problem should never be possible to occur. And more importantly there are a host of other pins that are manipulated the same way. That can have the same issue! basically all pin manipulation suffer from this potential issue, because there is no guards against this very well known race condition.  

**Problem description**
A race condition occurs when asynchronous code access the same variables non atomically.  So when interrupt routines and main loop code both interact with the same variables that take more than 1 instruction cycle to handle (non atomic). 

For speeduino the way the MCU pins of injectors, fan etc are manipulated  The same port register is used within main loop and interrupt routines. (If the FAN or BOOST and INJECTOR pin share the same MCU port for example). This can cause race conditions.

_Example:_
Example prerequisites. Fuel pump pin and injector1 pin are both on PORTB

If the main loop tries to write the FUEL pin high the following code steps are executed:
1a. Read current register status of PORTB
2a. Mask the current port status with the FUEL pin mask
3a. Write the masked value to register PORTB

If now this routine is interrupted by the fuel schedule interrupt after point 2 a race conditions occurs.  Because in the interrupt routine tries to write INJ1 pin HIGH:

1b. Read current register status of PORTB
2b. Mask the current port status with the INJ1 pin mask
3b. Write the masked value to register PORTB

So after the interrupt routine is done The code of the main loop continuous. But it will write the OLD register PORTB value, because it has read the value before the injector interrupt occurred. The same for IGN and IDLE PWM pins or any other combination of pins. 

**Solution**
The solution is different for Teensy/STM32 and the mega.

1.  For Teensy and STM32 the solution is to use atomic pin manipulations. With this PR these use digitalWrite() for all pin manipulations. digitalWrite() functions uses the BSRR in STM32. This is a method to set and reset pins of ports with 1 instruction. This prevents the race cndition. Teensy uses the same type of register. 
2. For the arduino mega2560 the pin manipulations inside interrupt routine are safe. The mega2560 has no nested interrupts so the port access inside a interrupt routine can not be interrupted halfway thru a write. All manipulations in the mainloop are altered to use digitalWrite() too, the interrupt is stopped before writing the pin when using digitalWrite(). For a few pins that are written inside a interrupt and also in the main loop, like VVT and BOOST, before access from the mainloop interrupts are stopped and restarted after. All interrupts are left as is. 

**Further actions**
_1. testing_
I did test this with my stm32F411 on the bench  and it solves the issue with the fuel pump, and has no direct visible side effects. Tested with 60-2 trigger and 4 cylinder wasted spark. For the mega i did the same with the same setup. But that is not really exhaustive testing. So more people should test this on all platforms. Especially the arduino Mega as this has a potential influence on main loop speed. 
@pazi88 hopefully you have some time to test this and see if your idle PWM problem is solved. (at least on my bench test i could not find anything, but i did not select core4 board).


adbancroft:
Nice find!

```
#ifdef __AVR__
#define no_AVR_Interrupts()  noInterrupts()
#define AVR_interrupts() interrupts()
#else
#define no_AVR_Interrupts()
#define AVR_interrupts()
#endif
```
Could this be modeled after the existing [avr-libc atomic macros](https://www.nongnu.org/avr-libc/user-manual/group__util__atomic.html)? 

They're a well understood idiom AND they take care of restoring the correct interrupt state. I.e nested calls to `no_AVR_Interrupts()/AVR_interrupts()` pairs.

Tjeerdie:
> Nice find!
> 
> ```
> #ifdef __AVR__
> #define no_AVR_Interrupts()  noInterrupts()
> #define AVR_interrupts() interrupts()
> #else
> #define no_AVR_Interrupts()
> #define AVR_interrupts()
> #endif
> ```
> 
> Could this be modeled after the existing [avr-libc atomic macros](https://www.nongnu.org/avr-libc/user-manual/group__util__atomic.html)?
> 
> They're a well understood idiom AND they take care of restoring the correct interrupt state. I.e nested calls to `no_AVR_Interrupts()/AVR_interrupts()` pairs.

I think that is possible, but to be honest i can not see how that is beneficial over the standard arduino interrupt()/nointerrupt() combination. Most of the code is already based of the arduino API anyway. Maybe i am missing something :-D.

adbancroft:
> I think that is possible, but to be honest i can not see how that is beneficial over the standard arduino interrupt()/nointerrupt() combination. Most of the code is already based of the arduino API anyway. Maybe i am missing something :-D.

It's best practice to write code that's hard to misuse. 

The atomic macros save & restore the interrupt state. So there is no risk that calling other functions will silently re-enable interrupts & introduce a hard to find bug. 

With the simple macros it's easier to end up with interrupts in an unexpected state especially as we have multiple devs working on the code over many years and PR's. E.g.
1. My code calls `no_AVR_Interrupts()`
2. My code calls another function written by someone else: `func_other()`
3. As a bug fix, a future PR submitter modifies `func_other()` to include calls to `no_AVR_Interrupts()`/`AVR_interrupts()`
4. Control returns to my code, but interrupts are now on - which wasn't the case before the PR bug fix.

You could argue that the bug fix submitter should check all code paths that call `func_other()` - but why add that friction when we can make the problem impossible to create in the first place?

VitorBoss:
> Nice find!
> 
> ```
> #ifdef __AVR__
> #define no_AVR_Interrupts()  noInterrupts()
> #define AVR_interrupts() interrupts()
> #else
> #define no_AVR_Interrupts()
> #define AVR_interrupts()
> #endif
> ```
> 
> Could this be modeled after the existing [avr-libc atomic macros](https://www.nongnu.org/avr-libc/user-manual/group__util__atomic.html)?
> 
> They're a well understood idiom AND they take care of restoring the correct interrupt state. I.e nested calls to `no_AVR_Interrupts()/AVR_interrupts()` pairs.

As far as I can see those macros disable interrupts **ONLY ON AVR** and do nothing on other platforms, making it irrelevant to the problem.

adbancroft:
> As far as I can see those macros disable interrupts **ONLY ON AVR** and do nothing on other platforms, making it irrelevant to the problem.

I asked if the solution could be **MODELED** after those macros, not that we use the macros as-is. Something like [this package](https://github.com/wizard97/SimplyAtomic).

TBAMax:
Generally a good find, that simple port writes can have a race condition.

Other here discussion about using the atomic macros versus nointerrupts()/interrupts(): I have seen some dead projects using "atomic" macros, so this is not generally a good practice (as lot of macro magic generally). 
`nointerrupts()/interrupts()` is much more Arduino spirit+ the intent of the code is clear, and what is behind it actually is clear.

TBAMax:
Some more notes of though here:
I noticed some simple & commands replaced with digitalWrite(). DigitalWrite is more likely cause the race condition.
Also on 8bit  Atmega the direct bit manipulation should not result in the possibility of race condition because this is done with one instruction (not read-modify-write). If it does still have the problem, it may come from using the pointer to the port instead of direct hardcoded port address or something. Although @Tjeerdie I got the impression you did not test if the problem exists on the Atmega at all? Also did you look at the dissassembly listing for the Atmega, what instructions are used?

TBAMax:
I took the time and looked at the disassembly listing (Atmega2560) for the port read and write, when it is done using "pin_port" and "pin_mask" as speeduino standard. It uses indirect read - modify - indirect write. So it is indeed suspectible to race condition and must be done with interrupts disabled temporarily. 
 Must be noted that this is still probably much better than using digitalWrite() because digitalWrite is known to have some extra checks and things so the interrupts would be disabled for longer and result in accuraccy loss.

Tjeerdie:
> > As far as I can see those macros disable interrupts **ONLY ON AVR** and do nothing on other platforms, making it irrelevant to the problem.
> 
> I asked if the solution could be **MODELED** after those macros, not that we use the macros as-is. Something like [this package](https://github.com/wizard97/SimplyAtomic).

There are multiple ways to solve this issue. So yes it could be modeled after that, but i see very little benefit because at the moment it is simply a 
```
NoInterrupt();
Port minpulation line;
Interrupt();
```

So i might be mistaken but how is that misused and potential misuse solved by adding more complexity in calling different macro functions that ensure you leave with interrupts enabled? I only see a lot of extra layers of abstraction and obscurity added. But that is just my opinion.

Tjeerdie:
> I took the time and looked at the disassembly listing (Atmega2560) for the port read and write, when it is done using "pin_port" and "pin_mask" as speeduino standard. It uses indirect read - modify - indirect write. So it is indeed suspectible to race condition and must be done with interrupts disabled temporarily. Must be noted that this is still probably much better than using digitalWrite() because digitalWrite is known to have some extra checks and things so the interrupts would be disabled for longer and result in accuraccy loss.

1. As said. indeed the assembly of direct port manipulation comes down to a port read -> modify -> write. So it is susceptible to race condition with the mega. (I did check)
2. The digitalWrite(); is to slow with too many checks for the Mega to be timing accurate. But it is ok for other systems like the teensy and STM32. Where both STM32 and Teensy use a special hardware register for digitalWrite(); that is actually atomic. 

And i have seen the race condtion occure in STM32 using direct port manipulation and its solved by disabling interrupts. @pazi88 has seen the race condition in the atmega idle PWM and it was solved by using a different approach. But the latest commit has not been checked by him as far as i know.

adbancroft:
The main reason is they take care of restoring the correct interrupt state:
```
NoInterrupt(); // If interrupts are already off....
Port minpulation line;
Interrupt();   // ...this will unexpectedly re-enable interrupts
```
Compare with:
```
ATOMIC_BLOCK(ATOMIC_RESTORESTATE) {
  Port minpulation line;
} // Interrupt state (on or off) correctly restored.
```

AVR & Teensy both support ATOMIC_BLOCK already. STM is the same CPU as Teensy, so the same code should work: https://github.com/PaulStoffregen/cores/blob/master/teensy3/util/atomic.h

Tjeerdie:
> The main reason is they take care of restoring the correct interrupt state:
> 
> ```
> NoInterrupt(); // If interrupts are already off....
> Port minpulation line;
> Interrupt();   // ...this will unexpectedly re-enable interrupts
> ```
> 
> Compare with:
> 
> ```
> ATOMIC_BLOCK(ATOMIC_RESTORESTATE) {
>   Port minpulation line;
> } // Interrupt state (on or off) correctly restored.
> ```
> 
> AVR & Teensy both support ATOMIC_BLOCK already. STM is the same CPU as Teensy, so the same code should work: https://github.com/PaulStoffregen/cores/blob/master/teensy3/util/atomic.h

Both STM32 and Teensy don´t need the ATOMIC_BLOCK() operation both don´t use the interrupt disable routine with this PR. 
Is there no side effect of using  ATOMIC_BLOCK(ATOMIC_RESTORESTATE)? Like it takes a lot more CPU cycles in time critical operations time?

Tjeerdie:
I did test this with a stm32F407 and stm32F411 on an engine and it seems to work perfectly fine. I have no board with a Mega so that is a little harder to test for me.

@noisymime It seems like a relevant and important bug to fix. Any comments on this PR? How to proceed?

noisymime:
The PR change itself has become a bit messy, but if I'm reading through it all correctly the neatest and simplest option might be to simply add `noInterrupt()` wrappers within the macros themselves (For mega 2560) and use `digitalWrite()` for stm32 as well as Teensy.

There's a tiny amount of overhead adding the `noInterrupts()` on the macro as it means it will get called from the ISR as well, but the difference this would make isn't worth worrying about. Eg

```
#if(defined(CORE_TEENSY) || defined(CORE_STM32))
  #define BOOST_PIN_LOW()     (digitalWrite(pinBoost, LOW))
  #define BOOST_PIN_HIGH()    (digitalWrite(pinBoost, HIGH))
  ...
#else
  #define BOOST_PIN_LOW()     (digitalWrite(pinBoost, LOW))
  #define BOOST_PIN_HIGH()    (digitalWrite(pinBoost, HIGH))
  #define BOOST_PIN_LOW()  noInterrupts(); *boost_pin_port &= ~(boost_pin_mask); interrupts()
  #define BOOST_PIN_HIGH() noInterrupts(); *boost_pin_port |= (boost_pin_mask); interrupts()
  ...
#endif
```

The nice part of doing it this way is that there's basically no change in the code logic itself, only the macro, you never have to worry about people adding some code down the track and forgetting to add the  `noInterrupts()`

Tjeerdie:
> The PR change itself has become a bit messy, but if I'm reading through it all correctly the neatest and simplest option might be to simply add `noInterrupt()` wrappers within the macros themselves (For mega 2560) and use `digitalWrite()` for stm32 as well as Teensy.
> 
> There's a tiny amount of overhead adding the `noInterrupts()` on the macro as it means it will get called from the ISR as well, but the difference this would make isn't worth worrying about. Eg
> 
> ```
> #if(defined(CORE_TEENSY) || defined(CORE_STM32))
>   #define BOOST_PIN_LOW()     (digitalWrite(pinBoost, LOW))
>   #define BOOST_PIN_HIGH()    (digitalWrite(pinBoost, HIGH))
>   ...
> #else
>   #define BOOST_PIN_LOW()  noInterrupts(); *boost_pin_port &= ~(boost_pin_mask); interrupts()
>   #define BOOST_PIN_HIGH() noInterrupts(); *boost_pin_port |= (boost_pin_mask); interrupts()
>   ...
> #endif
> ```
> 
> The nice part of doing it this way is that there's basically no change in the code logic itself, only the macro, you never have to worry about people adding some code down the track and forgetting to add the `noInterrupts()`

I think i tried that one first, I believe this caused some missing ignition and injection pulses on the mega when testing on the bench. But i am not 100% sure.  I will try this again when i can find the time.

I also see a merge conflict fest again. And my previous attempt at a rebase/merge with master does seem like it is not very neat anymore.

TBAMax:
> FYI, calling `interrupts();` inside an ISR prematurely re-enable the interrupts. This can have unexpected consequences.

Indeed. On the Atmega it causes nested interrupts. (Whe have the nested interrupts deliberately at the moment for timer2, with the same fashion. When this happens unintentionally then may not be good indeed).

adbancroft:
```
#if(defined(CORE_TEENSY) || defined(CORE_STM32))
  #define BOOST_PIN_LOW()     (digitalWrite(pinBoost, LOW))
  #define BOOST_PIN_HIGH()    (digitalWrite(pinBoost, HIGH))
  ...
#else
  #define BOOST_PIN_LOW()  \
    ATOMIC_BLOCK(ATOMIC_RESTORESTATE) { \
        *boost_pin_port &= ~(boost_pin_mask); \
    }
  #define BOOST_PIN_HIGH() \
    ATOMIC_BLOCK(ATOMIC_RESTORESTATE) { \
        *boost_pin_port |= (boost_pin_mask); \
    }
  ...
#endif

```

noisymime:
> ```
> #if(defined(CORE_TEENSY) || defined(CORE_STM32))
>   #define BOOST_PIN_LOW()     (digitalWrite(pinBoost, LOW))
>   #define BOOST_PIN_HIGH()    (digitalWrite(pinBoost, HIGH))
>   ...
> #else
>   #define BOOST_PIN_LOW()  \
>     ATOMIC_BLOCK(ATOMIC_RESTORESTATE) { \
>         *boost_pin_port &= ~(boost_pin_mask); \
>     }
>   #define BOOST_PIN_HIGH() \
>     ATOMIC_BLOCK(ATOMIC_RESTORESTATE) { \
>         *boost_pin_port |= (boost_pin_mask); \
>     }
>   ...
> #endif
> ```

Yep, this seems like the best way to go I think. I'll put something together around this and get it in shortly.

noisymime:
Please test latest master following d0286fe02df7b478a96125f271c0478987c27055. 

This issue should be resolved now. I can no longer reproduce here, but if others can confirm I will also get this included in the -fixes branch.

pazi88:
> Please test latest master following [d0286fe](https://github.com/noisymime/speeduino/commit/d0286fe02df7b478a96125f271c0478987c27055).
> 
> This issue should be resolved now. I can no longer reproduce here, but if others can confirm I will also get this included in the -fixes branch.

Should this fix the inj/ign/idle pin problems too? I can test on my ecu at least, because in that the problems are 100% reproduceable with mega.

Tjeerdie:
Hi noisy

I might have missed something because I am checking from my phone. But i
don't see any atomic port manipulation for injector and ignition channels?

If ignition and injection pinport manipulations are not taken into accout
those can still influence other pins because of the NVIC in stm32 and
Teensy.

Regards.

PR #976, opened 2023-01-03
Prevent EEPROM defer comparison fail at micros() overflow

DeionSi:
When micros() returns more than ULONG_MAX - EEPROM_DEFER_DELAY, deferEEPROMWritesUntil will be set to less than micros(), causing the micros() > deferEEPROMWritesUntil check to incorrectly succeed. In practice this issue can occur during one second every 71,58 minutes (micros overflow).
The issue and solution is explained at https://www.gammon.com.au/millis

I have tested on mega by loading different tunes without issues. However I'm not sure of the correct way to test this function.

Firmware becomes 196 bytes smaller. This change also removes a usage of currentLoopTime so it can be removed by #904.


DeionSi:
My proposed solution should not have that behavior. Please see the explanation in https://github.com/noisymime/speeduino/pull/904#issuecomment-1370100062

TBAMax:
@noisymime I reviewed that PR also and it seems legit. When doing timing, it is correct way to calculate interval and compare the intevals, this way it is overflow proof (unsigned integer math). Then when counter overflows math is still correct.

speeduino/storage.cpp
```diff
@@ -28,7 +28,7 @@ A full copy of the license may be found in the projects root directory
 #define EEPROM_LAST_BARO              (EEPROM_CALIBRATION_O2_BINS-1)
 
 
-uint32_t deferEEPROMWritesUntil = 0;
+uint32_t lastEEPROMDeferTime;
```
TBAMax, line 31: Maybe we could use here
 `unsigned long lastEEPROMDeferTime;` 
 or even something like:  `decltype(micros()) lastEEPROMDeferTime;`
 just to make more sure that this variable is same type as `micros()` return value, because when some platforms may have different size the overflow math may not work. Not a critical issue, but just rather for general caution (When some platform have for example 64bit micros() counter the overflow time is anyway too long and it does not matter any more.). I just discovered the `decltype` ,seems to be very convenient.

noisymime:
An mentioned, I think it's incredibly unlikely that anyone would come across this, but happy to include it. 

Did some testing on this and it seems fine, although there does need to be a default value set for `lastEEPROMDeferTime`

I had 1 odd, non-repeatable issue where an initial burn would not taking place at all. I'm fairly sure it was due to `lastEEPROMDeferTime` having a random high value as nothing is set at start time, causing the first burn to be delayed for an unknown period of time.

speeduino/speeduino.ino
```diff
@@ -334,7 +328,7 @@ void loop(void)
       #endif
 
       //Check for any outstanding EEPROM writes.
-      if( (isEepromWritePending() == true) && (serialReceivePending == false) && (micros() > deferEEPROMWritesUntil)) { writeAllConfig(); } 
+      if( (isEepromWritePending() == true) && (serialReceivePending == false) && (micros() - lastEEPROMDeferTime > EEPROM_DEFER_DELAY)) { writeAllConfig(); }
```
TBAMax, line 331:
```suggestion
      if( (isEepromWritePending() == true) && (serialReceivePending == false) && ((unsigned long)(micros() - lastEEPROMDeferTime) > EEPROM_DEFER_DELAY)) { writeAllConfig(); } 
```

noisymime: The cast probably shouldn't be needed as `micros()` is an unsigned long already, but the brackets are better here.

speeduino/storage.cpp
TBAMax, line 31:
```suggestion
uint32_t lastEEPROMDeferTime = 0;
```

DeionSi:
> An mentioned, I think it's incredibly unlikely that anyone would come across this, but happy to include it.
> 
> Did some testing on this and it seems fine, although there does need to be a default value set for `lastEEPROMDeferTime`
> 
> I had 1 odd, non-repeatable issue where an initial burn would not taking place at all. I'm fairly sure it was due to `lastEEPROMDeferTime` having a random high value as nothing is set at start time, causing the first burn to be delayed for an unknown period of time.

In C and C++ all global variables are initialized to 0, so that shouldn't be the cause of the issue. But I've changed it for clarity.

I'll do some burn testing and see what results I get.

speeduino/speeduino.ino
DeionSi: I've added the brackets.

speeduino/comms.cpp
```diff
@@ -244,7 +244,7 @@ void processSerialCommand(void)
       break;
 
     case 'b': // New EEPROM burn command to only burn a single page at a time 
-      if( (micros() > deferEEPROMWritesUntil)) { writeConfig(serialPayload[2]); } //Read the table number and perform burn. Note that byte 1 in the array is unused
+      if( (micros() - lastEEPROMDeferTime > EEPROM_DEFER_DELAY)) { writeConfig(serialPayload[2]); } //Read the table number and perform burn. Note that byte 1 in the array is unused
```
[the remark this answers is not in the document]
DeionSi: Brackets added, cast unnecessary.

DeionSi:
I just did: EEPROM_clear, wait for led, burn speeduino, send current tunerstudio config, wait for EEPROM burn, work offline, connect ts again
10 times for master
10 times for this branch
No errors or strangeness.

TBAMax:
Had a look, seems the changes in the PR is now outdated, because the code have changed so much meanwhile. Although seems the micros() misuse still exist in various places. 
Yet it is completely valid argument not to change this because it is quite unlikely that anyone will do the EEPROM write at exactly 70-th minute after board startup. Still scratches the eye some. 
I guess it is just me, aware of the millis() and micros() overflow preiods, it is so fun again to hear somebody complain "why my project do not work after 1,5 month or after 70minutes, and then tell just that I have no such problem, all my boards still work after those mentioned times too.

adbancroft:
Closing in favor of #1528

PR #977, opened 2023-01-07
Injector characteristic testing

TGYK:
This pull request is to add a feature for injector characterization in terms of flow rate and dead-time.

Changes are as follows:

- Modified speeduino.ini to add injector testing panels to HW testing submenu. This includes a panel for dead-time, a panel for flow rate, and a calculator for flow rate based off of measured volume of fuel from the test with currently set parameters.
- Modified comms.cpp and comms_legacy.cpp to allow for parsing "E" commands with multiple parameters. If a deadtime test is requested, the traditional cmdCombined TS_CommandButtonHandler function can be used, as the dead time is a single byte. If a flow test is requested, extract the parameters from the command payload in low-byte high-byte order and reassemble them into 16 bit words. Store the words in an array, pass array pointer to overloaded TS_CommandButtonHandler function for injector flow testing.
- Modified globals.h and globals.ino to add global variables for injector test parameters and a test status/activation.
- Modified TS_CommandButtonHandler.h and TS_CommandButtonHandler.ino to add functionality for deadtime and flow rate testing. Both of these cases do essentially the same thing: They extract parameters and set global values to trigger the injector test. The flow rate test case called for an overloaded function as all parameters could not be passed within the second byte of cmdCombined. This overloaded function takes the pointer to a word array from comms.cpp or comms_legacy.cpp which contains the parameters for the test.
- Modified speeduino.ino to add the actual injector test routine.. If the test is active and the engine is not running, check that injector 1 status is "OFF" and there are pulses left to send. If true, the fuel schedule for injector 1 is set with parameters from global values, and the total number of pulses left in the test is decremented. If false, check for the end of the test and disable it accordingly.

I have tested this in my own setup and found it to be functional, but am open to suggestions for better practices and code placement! My source for the flow rate equations can be found [here](http://www.megamanual.com/ms2/injectortest.htm).


pazi88:
Sounds really good and a long waited feature. Though need to test this to be able provide any feedback.

mike501:
I suggested doing something like this 12+ months ago on the forum as I
thought it was a great idea & would really help people on initial setup.
The response I got was it was polluting the code from its primary intent,
injector testing risks people spraying fuel over their car and causing a
fire (an infamous you tube video was referenced) and shouldn't be included.
Glad to see someone else has the same thought as me and went ahead with it.
I think it's a brilliant idea.

How can you be certain your calculated values match reality? Do we need to
get a series of tests of different injectors to validate what you've done?
Should we need to include safety warnings on the H/W test page?

TGYK:
> I suggested doing something like this 12+ months ago on the forum as I thought it was a great idea & would really help people on initial setup. The response I got was it was polluting the code from its primary intent, injector testing risks people spraying fuel over their car and causing a fire (an infamous you tube video was referenced) and shouldn't be included. Glad to see someone else has the same thought as me and went ahead with it. I think it's a brilliant idea. How can you be certain your calculated values match reality? Do we need to get a series of tests of different injectors to validate what you've done? Should we need to include safety warnings on the H/W test page?
> […](#)

I can’t be positive that the calculated values are perfect, as I don’t have an oscilloscope with a capture feature to verify pulse widths are what they theoretically should be, nor additional setups to test other injectors against manufacturer provided specifications. I would think that if others can with this equipment could test and verify, it would be ideal.

As for warnings, I put this in the HW test page because it already has a large warning at the bottom about risks. There are also checks within the code to verify that the engine isn’t turning.. I could make these checks a bit more robust, now that I am thinking about it.

HWright9:
Have a look at merging my PR #909 for fuel pump control via the hardware test page into this. You would also need a method to turn on the fuel pump to test injectors.
If you do, you will probably come across the same issues with the fuel pump control I did.

TGYK:
> Have a look at merging my PR #909 for fuel pump control via the hardware test page into this. You would also need a method to turn on the fuel pump to test injectors. If you do, you will probably come across the same issues with the fuel pump control I did.

While not a bad consideration, I don't know if that would reach too far outside of the realm of "atomic" feature additions. I do think it would be great to have in place alongside the injector characterization functions, but I would also think it's outside of the scope of this feature. Don't get me wrong, I am very open to merging your PR into mine, but I don't want to do so without considering the contribution guidelines and getting input from others on whether this sort of thing would be acceptable.

Being my first and only PR to the speeduino project, I don't want to overreach and get a denial because of feature creep!

TGYK:
Converted to draft due to an issue with orders of magnitude on timing values from TS to Speeduino. Will be debugging and testing when able this week.

TGYK:
Utilized a logic analyzer to do testing and debugging. I have confirmed timings are accurate and correct. Going to attempt to get comms_legacy working and do some cleanup before moving out of draft.

73Volvo:
This seems like a great idea and definitely something I'd use. It would really help a lot of people to get their projects started if they had accurate specs for their injectors, especially if they're working with an oddball car.

jackscr12:
oh man this would be a great addition to speeduinos capabilities, great job!

TBAMax:
According to the Speeduino style guide the variables should be "camelCase, lowercase first letter, no spaces or _" (https://wiki.speeduino.com/en/dev/Style_code).

TGYK:
Still would like testers to validate that this works on multiple setups - My hardware and testing capabilities are limited.

> According to the Speeduino style guide the variables should be "camelCase, lowercase first letter, no spaces or _" (https://wiki.speeduino.com/en/dev/Style_code).

I updated global variable names to reflect the requirements. Also cleaned up some of the all-lower-case names in the TS ini. Constants remained as-is per the guide.

PR #978, opened 2023-01-07
Remove half/full sync sequential decision from decoders

DeionSi:
This change replaces hasSync and BIT_STATUS3_HALFSYNC for all purposes except communications. Instead the decoders sets the decoderSync variable to the current decoded resolution. This way the decoders does not have to care about the sequential or not operation of Speeduino.

Firmware size, memory used and performance are unaffected. Bench-tested on mega with a few decoders.


adbancroft:
Closing: not mergeable and most if the changes have been added via other PRs

PR #982, opened 2023-01-12
Added Trigger for the Yamaha XTZ motorcycle

giovanni97bs:
This is a trigger based upon the work of @RempageR1 and adapted for my Yamaha Super Tenerè XTZ750 with EFI.
It is very similar to the Yamaha Vmax one but only has 4 uneven teeth on the crankshaft. I've used it for around 2000km.

For reference the crank has 4 teeth at 90° with uneven lenght, 3 are 7° large and one is 45°. TDC is 10° after the large tooth.

On the image there is the spark signal (original ECU, wasted spark) and the VR output with the low peaks being the 90° spaced tooth start and the high peaks being the end of the tooth (uneven).
CrankOscilloscope

This is my first real commit to the Speeduino project. Please correct me if needed.

### DIFF
--- a/speeduino/board_teensy41.h
+++ b/speeduino/board_teensy41.h
@@ -10,6 +10,7 @@
   uint16_t freeRam();
   void doSystemReset();
   void jumpToBootloader();
+  void setTriggerHysteresis();
   time_t getTeensy3Time();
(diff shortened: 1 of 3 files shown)


DeionSi:
It seems something is wrong with your decoders.h file.

PR #988, opened 2023-01-21
Add lambdaTable and memory map to afrTable

DeionSi:
This fixes the below warning when changing the "Lambda display" option. It also removes the hardcoded and now unneeded 14,7:1 conversion.

> Warning: MSQ Units Mismatch for afrTable! Lambda found in current configuration, AFR found in MSQ, values were converted to Lambda based on Gasoline 14.7:1.

AFR Target and Lambda Target tables can now be viewed simultaneously hence both options are available in the menus. Any changes in either are reflected in the other.

The lastOffset option causes the constant to be mapped to the same raw values as the previous row.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -5706,7 +5706,6 @@ cmdVSSratio6 =      "E\x99\x06"
      wueAnalyzeMap = warmup_analyzer_curve, warmup_afr_curve, afrTable1Tbl, afr, coolant, warmupEnrich, egoCorrection
      lambdaTargetTables = afrTable1Tbl, afrTSCustom
 #endif
-     lambdaTargetTables = afrTable1Tbl, afrTSCustom
      filter = std_DeadLambda ; Auto build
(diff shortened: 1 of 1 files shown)


DeionSi:
If "Lambda Display" is set to Lambda, on first run a conversion to Lambda will happen.

> Warning: MSQ Units Mismatch for afrTable! AFR found in current configuration, Lambda found in MSQ, values were converted to Afr based on Gasoline

With a stoichiometric ratio of 14,7:1 this makes to change to the table, so all is good.
With a different stoichiometric ratio than 14,7:1, TunerStudio's conversion will be incorrect. TunerStudio will flag that TunerStudio's settings are different than the controller's settings. The controller's settings are correct and should/must be used.

noisymime:
Nice way of getting both views! 

Just to confirm, the AFR table generator tool still works fine? Just not with the lambda table by the looks.

DeionSi:
Yes, the AFR table generator seems to be working fine.

PR #991, opened 2023-01-25
Piggyback decoder

giannaros900:
A piggyback decoder would be useful for users who want to use fuel only


digmorepaka:
This PR looks a little messed up. Do you have an example waveform or description of what you have implemented?

TBAMax:
Yeah something ir not right here. It shows 100 files changed.
@giannaros900  You would be probably better off close that and make new fresh branch, copy your changes over and reapply with the new PR.

PR #993, opened 2023-01-28
Add Lambda options for EGO correction and Nitrous leanest AFR

DeionSi:
These options could only be set with AFR. Now they can be set with Lambda as well. The fields point to the same data, so updating one updates the other.


VitorBoss:
Nice to have all lambda conversion at the same place, also the current mode speeduino does lambda conversion is a known problem, TS always send the calibration with 14.7 stoich, if you have the stoich value defined to any other value the lambda value will be wrong.
This is long known speeduino limitation, the stoich must be changer to 14.7 after calculation of req_fuel, very annoying and prone to user error.

DeionSi:
I'm halfway (?) done with a PR to change speeduino to use Lambda instead of AFR. It will make changing the stoichiometric ratio change the AFR instead of Lambda. https://github.com/DeionSi/speeduino/tree/AFRtoLambda-2. There are still roadblocks ahead though, like EGO correction calculations.

PR #994, opened 2023-01-28
Optimize PW function

DeionSi:
A few optimizations for the PW function. Details are in commit messages.

The calculations are unchanged, they are just structured differently.

All tests pass. Bench-tested and getting the same PW values before and after. 208 bytes flash reduction and 1-2 loops/sec increase.

### DIFF
--- a/speeduino/speeduino.ino
+++ b/speeduino/speeduino.ino
@@ -1312,6 +1312,9 @@ void loop(void)
  * @param injOpen Injector opening time. The time the injector take to open minus the time it takes to close (Both in uS)
  * @return uint16_t The injector pulse width in uS
  */
+#ifndef UNIT_TEST
+inline
+#endif
 uint16_t PW(const int& REQ_FUEL, const byte& VE, const long& MAP, const uint16_t& corrections, const int& injOpen)
 {
   //Standard float version of the calculation
--- a/speeduino/speeduino.ino
+++ b/speeduino/speeduino.ino
@@ -1317,46 +1317,44 @@ uint16_t PW(int REQ_FUEL, byte VE, long MAP, uint16_t corrections, int injOpen)
   //Standard float version of the calculation
   //return (REQ_FUEL * (float)(VE/100.0) * (float)(MAP/100.0) * (float)(TPS/100.0) * (float)(corrections/100.0) + injOpen);
   //Note: The MAP and TPS portions are currently disabled, we use VE and corrections only
-  uint16_t iVE, iCorrections;
-  uint16_t iMAP = 100;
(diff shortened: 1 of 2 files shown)


speeduino/speeduino.ino
```diff
@@ -1312,51 +1312,49 @@ void loop(void)
  * @param injOpen Injector opening time. The time the injector take to open minus the time it takes to close (Both in uS)
  * @return uint16_t The injector pulse width in uS
  */
-uint16_t PW(int REQ_FUEL, byte VE, long MAP, uint16_t corrections, int injOpen)
+uint16_t PW(const int& REQ_FUEL, const byte& VE, const long& MAP, const uint16_t& corrections, const int& injOpen)
```
TBAMax, line 1315: `uint16_t PW(const int& REQ_FUEL, const byte& VE, const long& MAP, const uint16_t& corrections, const int& injOpen)`
Just out of curiosity, a question. What is the intention behind using this style of parameters here? I mean did you find that it reduces flash or ram usage or something?

TBAMax: Ok, that is a interesting way to make code faster. And it works indeed! I even checked the dissassembly listing for this part, and it is considerably smaller/faster. Strange that compiler do not optimize this by itself... maybe some optimizations are configured other way.
It was not easy to find the correct place in the dissassembly listing because this function is used only once and so it is actually directly inlined to the loop() and no function call is generated, so you can not find it by the function name there.

noisymime: > Strange that compiler do not optimize this by itself... maybe some optimizations are configured other way.

You have to be very careful with passing by references like this as you can end up changing the original variables values unintentionally. A compiler would never enable this by default as it's something that can have very unintended behaviour. 

MISRA C 2012 forbids modifying the value of any function parameters, so in theory it would get picked up there, but it is something that needs to be carefully monitored.

adbancroft: > Ok, that is a interesting way to make code faster. And it works indeed!

I've never managed to get a performance improvement using this method. And Compiler Explorer doesn't show much difference in the assembler: [https://godbolt.org/z/944vEza8K](https://godbolt.org/z/944vEza8K)

Could you give more detail on what you've found?

DeionSi: Enabling reference parameters for the PW function saves 100 bytes of flash. When adding noinline to PW function the code grows by 82 bytes.

My assumption is that when the code is inlined the compiler is able to optimize more since references effectively make the passed and local variable the same.

adbancroft:
Closing: out of date & PW calcs have been optimized by other work.

PR #1000, opened 2023-02-04
Fix extra unaccounted fuel squirts at low RPM on Teensy 4.1

rafolg:
Channels 3 and 4 (in a 4x cyl application) have extra unaccounted squirts from cranking to under 1100 RPM
This PR fixes that and has been checked with logic analyzer and in real running car - my daily driver.

### DIFF
--- a/speeduino/init.ino
+++ b/speeduino/init.ino
@@ -459,6 +459,15 @@ void initialiseAll(void)
     ignition7EndAngle = 0;
     ignition8EndAngle = 0;
 
+    fuelSchedule1.scheduleFlags = 0;
+    fuelSchedule2.scheduleFlags = 0;
+    fuelSchedule3.scheduleFlags = 0;
+    fuelSchedule4.scheduleFlags = 0;
+
+    fuelSchedule5.scheduleFlags = 0;
+    fuelSchedule6.scheduleFlags = 0;
+    fuelSchedule7.scheduleFlags = 0;
+    fuelSchedule8.scheduleFlags = 0;
     if(configPage2.strokes == FOUR_STROKE) { CRANK_ANGLE_MAX_INJ = 720 / currentStatus.nSquirts; }
     else { CRANK_ANGLE_MAX_INJ = 360 / currentStatus.nSquirts; }
 
--- a/speeduino/scheduler.h
+++ b/speeduino/scheduler.h
@@ -159,6 +159,10 @@ inline void refreshIgnitionSchedule1(unsigned long timeToEnd) __attribute__((alw
  */
 enum ScheduleStatus {OFF, PENDING, STAGED, RUNNING}; //The statuses that a schedule can have
 
+//Define bit positions within engine variable
+#define BIT_SCHEDULE_NEXT        0    // Enable flag for planned next schedule (when current schedule is RUNNING)
+#define BIT_SCHEDULE_DECODER     1    // Decoder had set the scheduler timer
+#define BIT_SCHEDULE_REPEATED    2    // Ignition had already repeated
 /** Ignition schedule.
  */
 struct Schedule {
@@ -190,6 +194,7 @@ struct FuelSchedule {
   COMPARE_TYPE nextStartCompare;
   COMPARE_TYPE nextEndCompare;
   volatile bool hasNextSchedule = false;
+  volatile byte scheduleFlags;
 };
 
 //volatile Schedule *timer3Aqueue[4];
--- a/speeduino/scheduler.ino
+++ b/speeduino/scheduler.ino
@@ -154,343 +154,241 @@ duration: The number of uS after startCallback is called before endCallback is c
 endCallback: This function is called once the duration time has been reached
 */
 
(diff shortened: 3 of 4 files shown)


noisymime:
Can you provide some more details of your setup and under what conditions you’re seeing this (Log file and traces preferably)? It’s certainly not occurring on any of the 4 cylinder setups I’ve tested with (And there are many)

The PR in its current form will break a number of different configurations.

rafolg:
Below are screen captures of before and after modifications to the code to fix extra squirts. This is on Teensy 4.1

700RPM before changes
image

700RPM after changes
image

1050RPM before changes
image

1050RPM after changes
image

noisymime:
> This is on Teensy 4.1

Ahhhh OK. The T4.1 has a shorter max timer period and hence a shorter total period available for future schedules. Suspect that’s probably the cause here as I’ve never seen this behaviour on 2560 or T3.5 set ups. 

I’ll review in more detail over the next few days as I’ll need to confirm there’s no impact to other platforms.

rafolg:
Sure, I've got this running on Teensy 3.5 in my daily. By all means due diligence to be done on other platforms

noisymime:
The max timeouts for a given MCU are fixed and known, so the easier way of working around this might be to simply have an RPM limit on when the pre-defined next pulse is used. They are only there for use at very high RPM, so having such a check doesn't change their utility at all. 

Eg:
With the T4.1 timers the maximum is timeout is 55,923uS, which equates to the rotation time at around 1073rpm. This obviously aligns with what you're seeing with the 1100 RPM. 

An easy workaround would be to define a threshold (Eg 2000 as a safe value) and only set the `hasNextSchedule` flag if RPM is above that. This would be a lot less intrusive than the proposed change, which potentially has other implications.

rafolg:
Wouldn't the safe threshold value be satisfied if timeout+duration is less than max timer value?

TBAMax:
Hehe, looking at the code, the setFuelSchedule functions for different channels are also different. Some have check for `< MAX_TIMER_PERIOD` some have not, also some other things such as some variables set or not to set. 
So it may work to just copy the code from the channel that do not show the extra pulses!

@rafolg 

> Wouldn't the safe threshold value be satisfied if timeout+duration is less than max timer value?

@rafolg feel free to look #804  there this issue should be also corrected.

brunob45:
> Hehe, looking at the code, the setFuelSchedule functions for different channels are also different. Some have check for `< MAX_TIMER_PERIOD` some have not, also some other things such as some variables set or not to set. So it may work to just copy the code from the channel that do not show the extra pulses!

After my experimentation with #1003, I saw that INJ1 is the only one different from the others, but the issue seems to affect every INJ channel.

The issue doesn't seem to affect the ignition schedules though...

rafolg:
This PR also works, quite well. If 804 and 1003 also do the job, then it's up to Josh to take his pick. Or do something else different.

noisymime:
I had a quick test of this today and confirmed both the issue and the PR. It’s a fairly extensive PR though and not sure what base it was made from as there’s some odd diffs showing. Which part of the PR specifically was the fix? 

Just for reference, #804 is currently non-functional for Teensy 4.1, so I’ll have to go through that a bit more to correct it and verify whether it also fixes the issue.

TBAMax:
> Just for reference, #804 is currently non-functional for Teensy 4.1, so I’ll have to go through that a bit more to correct it and verify whether it also fixes the issue.

@noisymime can you comment the exact symptoms you are seeing when trying #804 on Teensy 4.1 ? Maybe I can do something to help correcting things there for Teensy 4.1.

noisymime:
> > Just for reference, #804 is currently non-functional for Teensy 4.1, so I’ll have to go through that a bit more to correct it and verify whether it also fixes the issue.
> 
> @noisymime can you comment the exact symptoms you are seeing when trying #804 on Teensy 4.1 ? Maybe I can do something to help correcting things there for Teensy 4.1.

It's the same as what was fixed with 1c5ae26d7b4de7910c17db3bfd46269cf8404344 but simply applying the same defaults didn't correct it in this case. Basically the board hangs on startup. 

I imaging the fix will be similar, I just haven't had a chance to play with it against that PR yet

rafolg:
> I had a quick test of this today and confirmed both the issue and the PR. It’s a fairly extensive PR though and not sure what base it was made from as there’s some odd diffs showing. Which part of the PR specifically was the fix?
> 
> Just for reference, #804 is currently non-functional for Teensy 4.1, so I’ll have to go through that a bit more to correct it and verify whether it also fixes the issue.

Hi @noisymime, I have updated the PR to reflect the new scheduler code but have not uploaded it to my Github, hence why you see the odd differences. Over the weekend I can verify my updated PR with signal analyzer before uploading it here.

The two main lines of code that fix the issue are:
if ((timeout+duration) > MAX_TIMER_PERIOD) { timeout_timer_compare = uS_TO_TIMER_COMPARE( (MAX_TIMER_PERIOD - 1) ); }

and 

if ((timeout+duration) < MAX_TIMER_PERIOD)

PR #1003, opened 2023-02-06
Generic scheduler function

brunob45:
Create generic static functions for INJ and IGN schedulers.

Instead of copy-pasting the same function for every scheduler, each scheduler function calls the same `inline static` function.
By providing the specific information as function parameters, the compiler is able to optimize the generic function for each scheduler. The output binary is thus exactly the same as before.

No change in function size (checked with `avr-nm -S --size-sort firmware.elf`).


DeionSi:
There is similar/same work going on in #804. Especially https://github.com/TBAMax/speeduino/pull/12/files towards that PR.

brunob45:
Wow, this is very close indeed. The goal of my PR was to refactor the code while generating the exact same binary. I'll let @noisymime close it.

adbancroft:
Closing: out of date and duplicated by other work

PR #1022, opened 2023-03-17
stm32 tweaks for pills and boards with 128k flash

dantob:


speeduino/board_stm32_official.h
```diff
@@ -126,14 +147,9 @@ extern "C" char* sbrk(int incr);
   #define EEPROM_LIB_H "src/SPIAsEEPROM/SPIAsEEPROM.h"
   typedef uint16_t eeprom_address_t;
   #include EEPROM_LIB_H
-    extern InternalSTM32F4_EEPROM_Class EEPROM;
```
Tjeerdie, line 129: This must stay the way was, so not remove it.
When using a STM32F401 with 256k of flash with internal flash as eeprom, only 128kb of flash is still available. The other 128kb is used for eeprom usage.

PR #1023, opened 2023-03-17
make it possible to disable legacy comms

dantob:
not enabled anywhere yet, but can be tested by adding -DDISABLE_LEGACY_COMMS to compiler flags


noisymime:
I think the intent of this PR is probably covered by 8c85b0c99860b02ae2881a6afdab64ed10b911d2
This will disable the legacy comms option after the first successful new serial request has been made, which will prevent any noise on the serial line from potentially causing erroneous changes to the tune.

noisymime:
Closing as this is covered by 8c85b0c99860b02ae2881a6afdab64ed10b911d2

PR #1024, opened 2023-03-17
platformio config updates for stm32

dantob:
* optimisation was set as -Os before by environment, should actually use -O2 now
* new board BlackPill_F401CE
* deps updated
* general consistency between boards


TBAMax:
@dantob Can you be more specific why the -O2 is now needed instead of -Os? I may have missed the discussion about that somwhere. Asking as a user of 407VE.

dantob:
It's definitely not required, all of these boards have >=256k flash space, so code size is not really a problem at this time. Some boards were trying to use -O3 before, but it was overridden by environment defaults (-Os), I felt like -O2 was a safer default (it's already used by the teensy default environment).

Whether any/much performance is gained from -Os to -O2 is probably worth some testing, as far as I understand -Os already includes most of the optimisations from -O2, except those that are known to greatly increase code size.

I'll try to do some stim testing over the next few days.

dantob:
Here are some results, from stm32 and mega2560. In both cases -O2 seems to have to be the best size/performance compromise. -O3 seems to have very little effect on stm32, but definitely seems worth it for mega2560 (I couldn't find any benefit to using -ffast-math -fshort-enums -funroll-loops with only the stim). TBH, any option on stm32 is okay, but -O2 matches what teensy is already using.

```
Mega2560 stim 36-1 crank with 1 cam

-Og (debug)
RAM:   [========= ]  93.7% (used 7676 bytes from 8192 bytes)
Flash: [========  ]  84.2% (used 213744 bytes from 253952 bytes)
 2500 RPM LPS  872
15000 RPM LPS  552

-O1
RAM:   [========= ]  94.0% (used 7697 bytes from 8192 bytes)
Flash: [========  ]  81.2% (used 206318 bytes from 253952 bytes)
 2500 RPM LPS  950
15000 RPM LPS  600

-Os (size)
RAM:   [========= ]  94.3% (used 7727 bytes from 8192 bytes)
Flash: [======    ]  62.8% (used 159574 bytes from 253952 bytes)
 2500 RPM LPS  988
15000 RPM LPS  610

-O2
RAM:   [========= ]  94.3% (used 7729 bytes from 8192 bytes)
Flash: [========  ]  76.3% (used 193746 bytes from 253952 bytes)
 2500 RPM LPS 1050
15000 RPM LPS  665
VS -Os
20  % flash size increase
 6  % speed increase at  2500 RPM
 8.5% speed increase at 15000 RPM

-O3 (aggressive)
RAM:   [========= ]  94.4% (used 7735 bytes from 8192 bytes)
Flash: [========  ]  79.5% (used 201796 bytes from 253952 bytes)
 2500 RPM LPS 1074 
15000 RPM LPS  688
VS -Os
23.5% flash size increase
 8.5% speed increase at  2500 RPM
 12 % speed increase at 15000 RPM

-O3 -ffast-math -fshort-enums -funroll-loops
RAM:   [========= ]  94.2% (used 7716 bytes from 8192 bytes)
Flash: [========  ]  83.4% (used 211918 bytes from 253952 bytes)
 2500 RPM LPS 1074
15000 RPM LPS  686 (10k more flash, but no real change, needs more testing on hardware)
VS -Os
28  % flash size increase
 8.5% speed increase at  2500 RPM
 12 % speed increase at 15000 RPM


-----------------------------------------------------------------------------
BlackPill F411CE 36-1 crank with 1 cam

-Os (size)
RAM:   [=         ]  10.8% (used 14200 bytes from 131072 bytes)
Flash: [===       ]  30.3% (used 158648 bytes from 524288 bytes)
 2500 RPM LPS 5674
15000 RPM LPS 5378

-Og (debug)
RAM:   [=         ]  10.8% (used 14212 bytes from 131072 bytes)
Flash: [=====     ]  47.6% (used 249432 bytes from 524288 bytes)
 2500 RPM LPS 5750
15000 RPM LPS 5422

-O1
RAM:   [=         ]  10.8% (used 14212 bytes from 131072 bytes)
Flash: [=====     ]  45.7% (used 239820 bytes from 524288 bytes)
 2500 RPM LPS 5792
15000 RPM LPS 5464

-O2
RAM:   [=         ]  10.8% (used 14212 bytes from 131072 bytes)
Flash: [====      ]  43.1% (used 226160 bytes from 524288 bytes)
 2500 RPM LPS 6468
15000 RPM LPS 6166
VS -Os
35  % flash size increase
13  % speed increase at  2500 RPM
13.5% speed increase at 15000 RPM

-O3
RAM:   [=         ]  10.8% (used 14212 bytes from 131072 bytes)
Flash: [====      ]  43.8% (used 229872 bytes from 524288 bytes)
 2500 RPM LPS 6534
15000 RPM LPS 6240
VS -Os
36.5% flash size increase
14  % speed increase at  2500 RPM
14.5% speed increase at 15000 RPM
```

TBAMax:
@dantob On blackpill (or stm32 in general) the loops per sec is so low because of blocking ADC is used (Most of the time it simply waits for the ADC to complete conversion). This is what instantly screams out to me looking at those numbers. Good review though.

noisymime:
If I'm reading right, this PR is going from the current `-O3` to `-O2`, is that right? I understand that `-Os` is the default, however `-O3` is explicitly set at the moment, but removed in this PR. 

Was that `-O3` being overridden?

dantob:
The intention is to use the default optimisation level -Os for all STM boards. Originally, yes, I wanted to use -O2, but the issue is that most of these boards reserve a very large (128K of flash, 50% on 256K parts, which is the size of the last flash erase sector, when using EEPROM emulation).

The F411CE boards originally used -O3, which idk how it can ever work, speeduino cannot fit into the first 128K, I think even with -Os it's still too big, and can only be used with external EEPROM.

Maybe, I am understanding this incorrectly, but this is what I came to after reading the docs and Tjeerdie's comment over here thttps://github.com/noisymime/speeduino/pull/1022#discussion_r1161105856

image

PR #1037, opened 2023-04-24
Renix 44 tooth decoder (includes 66 tooth code)

mike501:
Decodes renix 44 & 66 tooth trigger wheels. Proven for 9 months on two 44 tooth trigger wheels. Unable to successfully test on 66 tooth so far.
Works by using the fact every 11th tooth is either 90 degress (44 tooth wheel) or 60 degrees (66 tooth wheels). Speeduino believes it only has 4 or 6 teeth depending on the decoder and counts 11 teeth before telling speeduino its seen a "tooth".

Original code included cam detection but testers didn't have cam so couldn't be proven and was removed. Needs distributor and single channel ignition due to lack of a single point on the trigger wheel that is unique or cam.


noisymime:
Would it be possible to combine these into a single Renix pattern in TunerStudio rather than separate 44/66 ones, then use the cylinder count to determine which should get used? Just trying to avoid duplicate options in that menu if possible is all.

mike501:
Updated the code to use cylinder count to identify if 44 or 66 tooth trigger patterns as discussed.

noisymime:
I made a couple of small changes that looked like they came in accidentally. One thing remaining, could you please revert the changes in `speeduino.ino`? Looks like they were added incorrectly.

mike501:
updated with the latest version of master includes fixing accidental changes to speeduino.ini

PR #1048, opened 2023-05-12
Added User Development Reserved Area of Page15

HWright9:
Added User Development Reserved Area in Page 15. To protect a space for user customisation once all the EEPROM is used on the Mega.

These last 80 bytes of memory are proposed to be reserved to allow users space to develop their own code and functions in the spirit of always allowing user customisation in speeduino. This I think is core to the appeal of the speeduino environment for many users.

It should never be used by the main branch release as this will break any user customised features already developed outside the main branch. 

When all the EEPROM is used, more difficult conversations will happen about what features to change/remove/refactor etc. But this should always preseve an area for individual customisation of speeduino.

80 bytes was chosen as the size of one 8bit 8x8 table with axes.


giovanni97bs:
Would also add additional "free" channel output for data transfer not interfering with the AUX which is already shared between CAN and Analog.

skundlony (approved):

Issue #1051, opened 2023-05-24
HalfsyncToFullSync issues found on testing.


TBAMax:
When switching from half to full sync at engine startup, the closeInjector and openInjector assignments are changed. But this creates the possibility that when this is done during the fuel or ignition pulse, the started pulse on some channel is not turned off. This shows up occassionally on logic analyser when testing(image below). 
Indicated by very long injection pulse. This happens only occassionally, but will probably flood the corresponding cylinder.
Probably this might be  already a known issue.

<img width="960" alt="FUELpulse" src="https://github.com/noisymime/speeduino/assets/36799067/e3b98afe-f0c8-4857-8fd6-6309e8a17adf">

VitorBoss:
Yes, that could happen if the CAM sensor came after a pulse started, this is easily avoided by checking for stopped timers before changing.

TBAMax:
> 



> Yes, that could happen if the CAM sensor came after a pulse started, this is easily avoided by checking for stopped timers before changing.

Ok, I am on it. Hoping to get the PR submitted soon.

noisymime:
Fixed with #1052

PR #1052, opened 2023-05-25
HalfsyncToFullSync update

TBAMax:
Fix against possible condition that when fuel closeInjector assignments are changed, some injectors could remain open. To prevent this, check was added to ensure that no fuel injection is ongoing when switching is done from halfsync to fullsync.
Fixes issue #1051 (explanatory logic analyser images there).

### DIFF
--- a/speeduino/init.ino
+++ b/speeduino/init.ino
@@ -3587,11 +3587,13 @@ void initialiseTriggers(void)
 void changeHalfToFullSync(void)
 {
   //Need to do another check for injLayout as this function can be called from ignition
-  if( (configPage2.injLayout == INJ_SEQUENTIAL) && (CRANK_ANGLE_MAX_INJ != 720) )
+  noInterrupts();
+  if( (configPage2.injLayout == INJ_SEQUENTIAL) && (CRANK_ANGLE_MAX_INJ != 720) && fuelSchedule1.Status!=RUNNING && fuelSchedule2.Status!=RUNNING && fuelSchedule3.Status!=RUNNING && fuelSchedule4.Status!=RUNNING &&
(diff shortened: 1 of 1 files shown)


noisymime:
Is there the possibility with this, particularly on a 6 or 8 cylinder engine, that there's not a time when there's no injection pulse running? 

The additional conditions on the if statement mean that the half to full transition will only take place if there are no fuel channels currently running, but it's possible, particularly with high cylinder counts, that this might never happen and hence the full switch isn't made. 

This might be a theoretical problem only and never actually happen, but I thought I'd play devil's advocate and ask the question

noisymime:
I'm going to merge this as I think the pro's outweigh the potential con's, but I'll have a think about whether there's any other way to do this.

noisymime:
Merged into `202207-fixes` also

TBAMax:
> 



> Is there the possibility with this, particularly on a 6 or 8 cylinder engine, that there's not a time when there's no injection pulse running?
> 
> The additional conditions on the if statement mean that the half to full transition will only take place if there are no fuel channels currently running, but it's possible, particularly with high cylinder counts, that this might never happen and hence the full switch isn't made.
> 
> This might be a theoretical problem only and never actually happen, but I thought I'd play devil's advocate and ask the question

Good question actually, I never thought about it from this perspective. I think this particular thing is not problem, because during starting at very low RPM there is no such high duty cycle on injectors. But this brings into attention that we really do not have to check all 8 scheduler channels, because only half of the schedules are used in the halfSync mode... so things can be little further optimized already...  but lets give this a little more thought, maybe there is some more other way to do it, for now the burning issue is fixed.

PR #1059, opened 2023-06-09
Fix analog pin selection

pazi88:
The  `pinTranslate` -function doesn't really work for anything other than Arduino Mega. The digital pins part is ok, but for STM32 and Teensy, it gives out wrong analog pin. So for example if you try to use analog input as secondary spark table switch, it wont work on STM32/Teensy. Other thing is that this function is only called at the initialization, so all the inputs that use this function need to be on the `requiresPowerCycle` -list on speeduino.ini, for the analog input pin selections to work properly. This PR fixes these problems.


pazi88:
This is already included in another PR

PR #1060, opened 2023-06-16
Small fix so that vvt2PID tunings get updated in a correct frequency.

vaizq:
Now vvt2PID tunings get updated approximately once per second. As is the case with vvtPID.


jonbills:
looks like @noisymime made the same fix a couple of years ago, https://github.com/noisymime/speeduino/commit/b2878d66f8ddda3b6d478d9ad57ade24083f9335 
but somehow it didn't stick.

noisymime:
From what I can see b2878d66f8ddda3b6d478d9ad57ade24083f9335 updated this for vvt1 but not vvt2. Ideally both of these should switch to using the timer flags, but this is a good solution for now.

PR #1066, opened 2023-07-06
Amend Speeduino ini hint for VR sensors to make it clearer which edge to pick

mike501:
Despite having details in the wiki we still get frequent questions on trigger edge settings relating to VR sensors. Amended the hint against the primary and secondary edge selection to inform which edge to pick for Max or DSC VR conditioners. 

Hopefully this will remove some confusion, reduce frustration and ease peoples start with speeduino as well as get rid of some comments about timing drift.


PSIG-dev (commented):
Ah, I didn't see this coming, or would have suggested the difference of the modules in output inversion, e.g.:
VR conditioners need to have Trigger Edge in TunerStudio set to the correct option based on the conditioner you are using, based on either signal inverted output or non-inverted output. Rising for MAX or LM conditioners (inverted output), and Falling for DSC and most others (non-inverted output).

Issue #1061, opened 2023-06-19
Possibility of extended ignition pulse and Dwell limiter bypassed when switching from half to full sync.


TBAMax:
This is related to the #1051 issue that is fixed by now, and was for the fuel injection. The same is possible for the ignition channels. That when switching from half sync to full sync, the ignStartFunction() ignEndFunction() assignments are changed, and this may result in the already started ignition pulse to stick ON, resulting in very long dwell pulse. May happen occassionally when switching moment happens to align with the ignition pulse. 
 Dwell limiter also do not prevent of catch this, because dwell limiter also uses every channels individual ignEndFunction() to end the pulse, but when this function do not affect the real particular output any more this has no effect.

TBAMax:
Managed to capture the ignition pulse sticking during test, just a quick image here while the solution is in the proggress:
![image](https://github.com/noisymime/speeduino/assets/36799067/afa9913e-8f05-4d5c-a259-e619c6c8fd1f)

TBAMax:
Just another image showing more clear that this is happening during starting up:
![image](https://github.com/noisymime/speeduino/assets/36799067/bff34725-3a47-40e0-8ad8-62997a0e5e39)

TBAMax:
I close the issue here, just to get the issues folder a little cleaner. Also since there is already a pull request(s) up for fixing the issue.

PR #1068, opened 2023-07-07
Fixes to the HalfToFullSync

TBAMax:
Fixes against the possibility that some ignition outputs can at rare cases be left fully ON for the whole cycle.
Fixes the #1061.
Related to the #1051 and #1052 , but this time the different approach taken to resolve the issue.


TBAMax:
Replaced by #1069.

PR #1069, opened 2023-07-08
Fix against rare possibility of overdwell ignition pulse.

TBAMax:
New better version of #1068.


TBAMax:
The issue it was correcting still stands, but this branch was outdated.

PR #1073, opened 2023-07-21
Performance: optimize 32-bit division

adbancroft:
This PR applies some optimizations around 32-bit division that result in **90% to 100%+** performance improvement (loop/s, dependent on tune) and a small reduction in reported RAM usage.

1. Simplify crankMaths: base angle to time and time to angle conversions on decoder computed variables. The conversions use fixed point integer math. I.e no division. The necessary division is done by the decoder code, which is called much less frequently than the conversion functions (~20 to 1 ratio)
  1.1 This allows us to eliminate doCrankSpeedCalcs().
2. Decoders now flag when revolutionTime changes: when it doesn't change, we can avoid running many scheduler calculations and re-use previously computed values. **THIS NEEDS CAREFUL REVIEW**.
3. Decoders that use a fixed triggerToothAngle now use libdivide to optimise triggerSetEndTeeth()
4. Optimize div100([u]int32_t): test if 16-bit calls are possible.
5. Add `udiv_32_16()` (and unit tests). 
  This is optimised for U32/U16 => U16 and is ~60% faster than using the standard U32/U32=>U32 division operator (i.e. __udivmodsi4).
  Apply `udiv_32_16()` where feasible. I.e. everywhere that is already doing U32/U16 => U16 
6. Misc. small changes to avoid 32-bit division where possible.


adbancroft:
Replaced by #1081 & #1082

PR #1076, opened 2023-07-26
Fix STM32duino RTC library version to 1.2.0

pazi88:
Seems that the RTC on STM32 fails to keep up time when powered on VBAT if the RTC library version is higher than 1.2.0. So for now, I did fix the RTC library version to 1.2.0 which seems to be the newest one that works.

I did also report this bug on the library itself, so lets see if this gets fixed at some point: https://github.com/stm32duino/STM32RTC/issues/96

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -71,7 +71,8 @@ platform = ststm32
 framework = arduino
 ;board = genericSTM32F407VET6
 board = black_f407ve
-lib_deps = stm32duino/STM32duino RTC, greiman/SdFat
+; RTC library fixed to 1.2.0, because in newer than that the RTC fails to keep up time. At least up to 1.3.7 version
+lib_deps = stm32duino/STM32duino RTC @ 1.2.0, greiman/SdFat
 board_build.core = stm32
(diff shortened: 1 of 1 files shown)


noisymime:
What was the outcome of this one? 

Based on the stm32duino issue, this is a code issue within the Speedy firmware somewhere?

noisymime:
Is this still needed?

pazi88:
> Is this still needed?

Yes. It doesn't work without. Though I haven't had time to test if the newer library works with code changes.

noisymime:
Have merged this one for now, but let's keep an eye on the situation with the new versions and see if this can be removed at any point in the future.

pazi88:
> Have merged this one for now, but let's keep an eye on the situation with the new versions and see if this can be removed at any point in the future.

Good. I have been trying to make this work with the newest library version based on the authors suggestions, but I haven't been able to make it work, as well as it works with 1.2.0 version. I can make it "work" with the latest version, but that will cause 2-second delay when powering on without RTC battery, which seems like forever when trying to start the car. Which is unacceptable in my opinion.

But I try to work on this, when I have time.

PR #1077, opened 2023-07-27
TS Engine constants menu

adamg-86:
Moved the fueling calculation options from VE Table to the engine constants menu.
The option takes a lot of space in the VE table so I moved them to another menu. 
TS-menu


noisymime:
I understand trying to get more space for the VE table, but the engine constants dialog really isn't the right place for these settings. They are closely tied to the VE table and need to be alongside it somehow to show that association. 

Perhaps a column on the right of the table rather than another row underneath? It's the vertical real estate that is tighter than the horizontal, so that might make for a better fit .

noisymime:
Closing as this is not the correct location for these settings. I am considering removing some of these for future versions, which would make this redundant.

PR #1079, opened 2023-07-28
T4.1 Flex timers for knock monitoring

rafolg:
Place holder code to implement flex timers for knock monitoring on T4.1


rafolg:
This code is still incomplete outside of my private branch. Just wanted to show my logic to communicating to the TPIC via advanced SPI. The TPIC will always lag one response behind the current quey, the datasheet sets the query order. So in order to get a specific response, two queries need to be made - That is the one you want plus the next one on the list.

PR #1081, opened 2023-08-06
Performance: use cached values instead of recalculating

adbancroft:
In `loop()`, only recalculate pulse widths, injection angles and ignition angles when the inputs to those calculations have changed. Otherwise use the values from the previous iteration of `loop()`.

Exact speed up varies dependent on tune & rpm. Testing between 3900 & 4100 RPM shows a 15% to 40%+ speed up in loop/s on Mega2560 (tested under various combinations of spark mode, fuel trims etc.).

Log comparison below is 4 cylinder, 60-2 missing tooth, full sequential fuel + fuel trims, sequential spark and RPM 1000 to 12000:
avoid_calcs

### DIFF
--- a/speeduino/auxiliaries.ino
+++ b/speeduino/auxiliaries.ino
@@ -887,10 +887,10 @@ void nitrousControl(void)
     if( (isArmed == true) && (currentStatus.coolant > (configPage10.n2o_minCLT - CALIBRATION_TEMPERATURE_OFFSET)) && (currentStatus.TPS > configPage10.n2o_minTPS) && (currentStatus.O2 < configPage10.n2o_maxAFR) && (currentStatus.MAP < (uint16_t)(configPage10.n2o_maxMAP * 2)) )
     {
       //Config page values are divided by 100 to fit within a byte. Multiply them back out to real values. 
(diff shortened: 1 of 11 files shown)


TBAMax:
Quite a lot of changes for implementation of idea that seems quite easy to do by description...
Though interested to hear about the tests on that. In some sence seems somewhat dangerous optimization, because of possibility of providing false sence of good loop speed, but it actually falls during transitional conditions when loop speed is actually needed. Talking about the general idea in the headline. No idea about the specific implementation.

adbancroft:
> Quite a lot of changes for implementation of idea that seems quite easy to do by description...

Totally agree: there was quote a bit of prep work needed make the final implementation "simple". There are only a few commits that really make the PR: ae979dfef3c7ba965789e3920a05e7088a58e09f, 727085679d190183c466196974a9e7d17a666de8, 1129c2ae3c3c4c56e5883744be1b380fe2480f83, ca3c4dc488ee85187081222e3803a18cc87a53aa

adbancroft:
> In some sence seems somewhat dangerous optimization

Yes - it's a little fragile (easy for future PRs to break it) and and the data flow/dependency flow is (possibly?) complex for maintainers who aren't familiar with the code base.

But the performance gain is big enough to justify it IMO

PR #1082, opened 2023-08-07
Performance: optimize division

adbancroft:
This PR applies some optimizations around 32-bit division that result in 60%+ performance improvement (loop/s, dependent on tune) and a small reduction in reported RAM usage:

1. Simplify crankMaths: base angle to time and time to angle conversions on decoder computed variables. The conversions use fixed point integer math. I.e no division. The necessary division is done by the decoder code, which is called much less frequently than the conversion functions (~20 to 1 ratio)
   This allows us to eliminate doCrankSpeedCalcs().
3. Decoders that use a fixed triggerToothAngle now use libdivide to optimise triggerSetEndTeeth()
4. Optimize div100([u]int32_t): test if 16-bit calls are possible.
5. Add udiv_32_16() (and unit tests). 
    This is optimised for U32/U16 => U16 and is ~60% faster than using the standard U32/U32=>U32 division operator (i.e. __udivmodsi4).
7. Apply udiv_32_16() where feasible. I.e. everywhere that is already doing U32/U16 => U16
8. Misc. small changes to avoid 32-bit division where possible.

Log comparison below is 4 cylinder, 60-2 missing tooth, full sequential fuel + fuel trims, sequential spark and RPM 1000 to 12000:
optimize division


adbancroft:
@noisymime FYI, the ASM implementation of `udiv_32_16()` is ~40% faster than the C version. Overall that is worth 25-35 loop/s (or ~2-3% speedup).

adbancroft:
Improved accuracy of fixed point math by rounding divisions & right shifts, instead of truncating. See `SetRevolutionTime, angleToTimeMicroSecPerDegree, timeToAngleDegPerMicroSec`

PR #1093, opened 2023-09-09
Add filter count for when a filter rejects a primary trigger pulse

mike501:
Enables users to see if a filter is having lots of fails to help diagnose issues and confirm if they've got the right filter setting.

A number of decoders had the formatting corrected when pasting in the 2 lines to count the failures.


mike501:
updated code to latest release, did a push and its not appeared here. Closing this PR and raising a new one.

PR #1097, opened 2023-09-15
Map Threshold for DFCO

155ac:
*Adds a MAP threshold option for DFCO, This will make the DFCO code to work like the ms2, It will only trigger the DFCO when both MAP and TPS are below the threshold

*Adds minimum value of 0.5% for TPS this fix a problem when the values is set to 0% the DFCO is not triggering because the TPS never gets below 0%

*Adds default values for DFCO in TS

*Change DCFO settings locations and change some names for better understanding

new dfco

### DIFF
--- a/speeduino/board_stm32_official.h
+++ b/speeduino/board_stm32_official.h
@@ -22,8 +22,7 @@
   #include "stm32f3xx_ll_tim.h"
 #elif defined(STM32F4)
   #include "stm32f4xx_ll_tim.h"
-  #include "stm32f4xx_hal_can.h"
-  #ifndef HAL_CAN_MODULE_ENABLED
+  #if defined(STM32F407xx) && !defined(HAL_CAN_MODULE_ENABLED)
     #warning "CAN module is not enabled. Internal CAN will NOT be available"
   #endif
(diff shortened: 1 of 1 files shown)


Nocnafuria22:
Great feature but don't forget to add hysteresis because the map value can oscillate during DFCO.

155ac:
> Great feature but don't forget to add hysteresis because the map value can oscillate during DFCO.

there is no need for hysteresis because the speeduino it self have a option to remove oscillations from the map.

Nocnafuria22:
> > Great feature but don't forget to add hysteresis because the map value can oscillate during DFCO.
> 
> there is no need for hysteresis because the speeduino it self have a option to remove oscillations from the map.

Not with instantaneous map sample method. Even with event average it can oscilate 1-2 kpa per 0.5 sec.

noisymime:
1 concern I have with this is what happens if someone doesn't have a MAP sensor connected? The same can probably be said for TPS, but that is more unusual than no MAP sensor. 

Is it worth adding a separate option to enable this?

Nocnafuria22:
``
> 
> Is it worth adding a separate option to enable this?

In my case it would be helpfully.

PR #1103, opened 2023-09-26
Add missing braces around expressions in INI

adbancroft:
The INI file is missing braces around some expressions.


noisymime:
Just for future reference `lastOffset` is a TunerStudio keyword rather than a variable. Putting it in brackets apparently causes TS to produce an error.

Issue #1100, opened 2023-09-16
DFCO Hysteresis limited to 255rpm


mbbrinkman:
It is suggested in the notes that 200-300rpm of DFCO hysteresis is typical and that some users might want higher.  However, because dfcoHyster is a byte, it can take values between 0-255.  dfcoRPM is also a byte and allows us to go higher, but this is because it is multiplied by 10 in corrections.ino when actually used thus giving us range at the cost of precision.

The obvious fix is to just multiply dfcoHyster by 10 also, but I'm not sure if there is an easy way to make this change without breaking systems, so I'll just point this out rather than sending a pull request.  Perhaps change it from byte to word?

VitorBoss:
The easy change is multiply by 2 or 4 as 10 would be too much, if you need more than 500 you are doing it wrong

PR #1105, opened 2023-09-29
Fix wrong info in DFCO and sets default values for TS

155ac:
This pull request should fix de issue #1100 because the DFCO hysteresis rpm is not in the range in the info ( 300 rpm when the maximum setting range is 255 ) causing user confusion about the setting, and adds default values for all DFCO settings in Tuner Studio.


155ac:
I'm going to close this pull because I'm going to combine this with the pull MAP in DFCO

PR #1109, opened 2023-10-04
Suzuki K6A Decoder

mike501:
K6A decoder used in 3 cylinder Suzuki engines from 1994 to 2018. May work with other Suzuki K engines. 

Has fixed trigger angles from a decoder on the cam. Pattern is 6+1. The key trigger angle is 70 degrees before a cylinder TDC. Code works out the trigger angle each time crank angle is called to ensure tooth based ignition can be done to improve accuracy when the engine spins up to 9k rpm.


noisymime:
Is there a true need to change the tooth edge on the default case in init.cpp or has that been done by mistake?

mike501:
No idea why that has changed - it's not part of the decoder as far as I can
see or remember. I think it can revert back to what is on the current
master.

PR #1063, opened 2023-06-26
Save 600+ bytes RAM (step 4 of 9) - deduplicate scheduler ISR functions

adbancroft:
This is the 4th PR in a series of scheduler focused PRs that end up saving >600 bytes RAM (compared to master as of 2023-06-11)

This PR removes the duplicate code in the per-channel scheduler ISR functions, routing them through an equivalent function that handles all channels:
```
ignitionScheduleISR
fuelScheduleISR
```
These are completely inlined, so no loss in performance.

The end game is [this branch](https://github.com/adbancroft/speeduino/tree/SCHD_8_scheduler_arrays). Since there are so many changes, I split them into as a series of PRs (each with a specific focus), to make the changes easier to understand, test and get feedback on.

1. https://github.com/noisymime/speeduino/pull/1010 (merged)
2. https://github.com/noisymime/speeduino/pull/1039 (merged)
3. #1018 
4. This PR
5. De-duplicate setSchedule functions
6. De-duplicate schedule structs
7. Optimize ignition schedule struct
8. Optimize fuel schedule struct
9. Use arrays to store schedulers.

PR #1018, opened 2023-03-10
Save 600+ bytes RAM (step 3 of 9) - remove per-channel schedule calculation functions, use generic equivalents instead

adbancroft:
This is the 3rd PR in a series of scheduler focused PRs that end up saving >600 bytes RAM (compared to master as of 2023-06-11) 

This PR removes the per-channel schedule calculation functions, use generic equivalents instead:

```
- calculateInjectorStartAngle
- calculateInjectorTimeout
- calculateIgnitionAngle
- calculateIgnitionTrailingRotary
- calculateIgnitionTimeout
```
No loss in loop/sec, unit tests all pass.

The end game is [this branch](https://github.com/adbancroft/speeduino/tree/SCHD_8_0_scheduler_arrays). Since there are so many changes, I split them into as a series of PRs (each with a specific focus), to make the changes easier to understand, test and get feedback on.

1. #1010 (merged)
2. #1039 (merged)
3. This PR
4. De-duplicate scheduler ISR functions
5. De-duplicate setSchedule functions
6. De-duplicate schedule structs
7. Optimize ignition schedule struct
8. Optimize fuel schedule struct
9. Use arrays to store schedulers.

This will also remove 400+ MISRA violations :-)

PR #1010, opened 2023-02-21
Unit test the schedule calculations

adbancroft:
Add unit tests for the scheduling calculations:

- calculateInjectorStartAngle
- calculateInjector1Timeout
- calculateInjectorNTimeout
- calculateIgnitionAngle1 - calculateIgnitionAngle8
- calculateIgnition1Timeout
- calculateIgnitionNTimeout

This required moving these functions out of speeduino.ino, but that's all that has been changed.

PR #1111, opened 2023-10-07
Save 600+ bytes RAM (step 5 of 9) - De-duplicate setSchedule functions

adbancroft:
This is the 5th PR in a series of scheduler focused PRs that end up saving >600 bytes RAM (compared to master as of 2023-06-11)

This PR removes the duplicate code in the per-channel setSchedule functions (setFuelSchedule\d|setIgnitionSchedule\d), routing them through a generic function that handles all channels:
```
setIgnitionSchedule
setFuelSchedule
```

The end game is [this branch](https://github.com/adbancroft/speeduino/tree/SCHD_8_scheduler_arrays). Since there are so many changes, I split them into as a series of PRs (each with a specific focus), to make the changes easier to understand, test and get feedback on.

1. #1010 
2. #1039 
3. #1018 
4. #1063 
5. This PR
6. De-duplicate schedule structs
7. Optimize ignition schedule struct
8. Optimize fuel schedule struct
9. Use arrays to store schedulers.

### DIFF
--- a/speeduino/decoders.cpp
+++ b/speeduino/decoders.cpp
@@ -385,17 +385,6 @@ static __attribute__((noinline)) int crankingGetRPM(byte totalTeeth, bool isCamT
 
   return currentStatus.RPM;
 }
-#define MIN_CYCLES_FOR_ENDCOMPARE 6
-
-inline void adjustCrankAngle(IgnitionSchedule &schedule, int endAngle, int crankAngle) {
-  if( (schedule.Status == RUNNING) ) { 
-    SET_COMPARE(schedule.compare, schedule.counter + uS_TO_TIMER_COMPARE( angleToTimeMicroSecPerDegree( ignitionLimits( (endAngle - crankAngle) ) ) ) ); 
-  }
-  else if(currentStatus.startRevolutions > MIN_CYCLES_FOR_ENDCOMPARE) { 
-    schedule.endCompare = schedule.counter + uS_TO_TIMER_COMPARE( angleToTimeMicroSecPerDegree( ignitionLimits( (endAngle - crankAngle) ) ) ); 
-    schedule.endScheduleSetByDecoder = true; 
-  }
-}
 
 /**
 On decoders that are enabled for per tooth based timing adjustments, this function performs the timer compare changes on the schedules themselves
--- a/speeduino/decoders.cpp
+++ b/speeduino/decoders.cpp
@@ -385,6 +385,17 @@ static __attribute__((noinline)) int crankingGetRPM(byte totalTeeth, bool isCamT
 
   return currentStatus.RPM;
 }
+#define MIN_CYCLES_FOR_ENDCOMPARE 6
+
+inline void adjustCrankAngle(IgnitionSchedule &schedule, int endAngle, int crankAngle) {
+  if( (schedule.Status == RUNNING) ) { 
+    SET_COMPARE(schedule.compare, schedule.counter + uS_TO_TIMER_COMPARE( angleToTimeMicroSecPerDegree( ignitionLimits( (endAngle - crankAngle) ) ) ) ); 
+  }
+  else if(currentStatus.startRevolutions > MIN_CYCLES_FOR_ENDCOMPARE) { 
+    schedule.endCompare = schedule.counter + uS_TO_TIMER_COMPARE( angleToTimeMicroSecPerDegree( ignitionLimits( (endAngle - crankAngle) ) ) ); 
+    schedule.endScheduleSetByDecoder = true; 
+  }
+}
 
 /**
 On decoders that are enabled for per tooth based timing adjustments, this function performs the timer compare changes on the schedules themselves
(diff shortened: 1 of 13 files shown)


noisymime:
The introduction of `adjustCrankAngle()` in `schedule_calcs.cpp` adds a new context switch out of the primary trigger interrupt that would be good to avoid if possible. In-lining likely won't be possible because the function is in a different compilation unit (Though I'd need to confirm this in the asm). 

That context switch could get expensive is all, particularly given the timing critical nature of that interrupt.

adbancroft:
By context switch you mean function call? (it occurs within the ISR scope, so there is no traditional context switch of saving registers, the stack, PC etc.)

`adjustCrankAngle` is defined inline & is in-lined during compilation - confirmed by using the PIO memory inspect tool & disassembly of `firmware.elf`.

I did notice that `checkPerToothTiming` is not _always_ in-lined - _even in current master_. Probably because it's called in so many places. And forcing it inline causes `adjustCrankAngle` to not be in-lined.

So I don't think this PR will slow down the decoder interrupts. 

I can try limiting calls to `checkPerToothTiming` to see if it can be coaxed inline without negative side effects.

adbancroft:
I applied the "flatten" attribute to the decoder I test with (dual wheel 60-2). Maybe.....5-10  loop/sec speed increase. IMO, not worth it in over using additional 1.3kb flash _per decoder_.

TBAMax:
@adbancroft  Probably while on the schedulers it would be nice to facilitate the support for the new envs for only one and three ignition channels. This means not declaring more ignition schedule structs than needed. Previously it was minimum of 5 ignition channels. Now it is that some environments have only 1. So `#if IGN_CHANNELS >= x` needs to be updated/added at all places. As I understand @noisymime added the new envs in without actually implementing those yet. There may be other places in the code that need to be altered to facilitate the less ignition channels but really can not tell exactly because it do not give any compilation errors (yet) . When omitting declarations for the unneeded ignition channels some places may come out from the compilation errors.

adbancroft:
> Probably while on the schedulers it would be nice to facilitate the support for the new envs for only one and three ignition channels.

@TBAMax 

The end game I hope to get to is arrays of schedulers, which takes care of this issue automatically (for the most part). See [here](https://github.com/adbancroft/speeduino/tree/SCHD_8_scheduler_arrays) for a preview.

TBAMax:
> The end game I hope to get to is arrays of schedulers, which takes care of this issue automatically (for the most part). See [here](https://github.com/adbancroft/speeduino/tree/SCHD_8_scheduler_arrays) for a preview.

Ok, yeah. This can be implemented after that, of course. Otherwise there would be double work, implementing it now and then re-doing it later again.
Looked at the mentioned preview... still not sure about the sheduler arrays. There were some inconsistency with the naming, some ignition related macros(*that are actually functions now whitch is nice) seemed to be 1 based counting, for example IGN1_TIMER_ENABLE() and shedulers are obviously then 0 based counting, so that probably the before mentioned macro actually affects channel that is schedule[0] in the shedule array. So i do not see much benefit there, unless all without exeptions is converted to 0 based counting logic. Which in turn is quite strange to force to the end user because there the cylinder numbers still are counted starting from 1... Yeah rant here.
Anyway @noisymime, do not be discouraged because us chatting here already about some other topics. This particular PR itself seems quite solid, and definetly again improvement.

@adbancroft one thing I noticed to pick on again: the name of the function reset() for resetting the schedules. Would'nt it be better to use something more like resetSchedule() or resetChannel() because it seems higly likely that "reset" alone can be some special built in keyword already or will be some day...

adbancroft:
> There were some inconsistency with the naming, some ignition related macros(*that are actually functions now whitch is nice) seemed to be 1 based counting, for example IGN1_TIMER_ENABLE() and shedulers are obviously then 0 based counting, so that probably the before mentioned macro actually affects channel that is schedule[0] in the shedule array.

I don't think this is bad. One is a domain concept (cylinder #<span/>1 is first in firing order); the other is the implementation (it's stored in an array & in C/C++ arrays are 0-indexed). Using domain concepts within a code base is normal and good practice (there's a whole [software design paradigm](https://en.wikipedia.org/wiki/Domain-driven_design) based on it).

However, we should be consistent in that usage/terminology across the code base - not so sure about that. I'll review for any inconsistencies

adbancroft:
> the name of the function reset() for resetting the schedules. Would'nt it be better to use something more like resetSchedule() or resetChannel() because it seems higly likely that "reset" alone can be some special built in keyword already or will be some day

[YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). Those functions are local to one cpp file - we can rename them in the future with no side effects.

adbancroft:
> This particular PR itself seems quite solid, and definetly again improvement.

Thank you!

noisymime:
> Anyway @noisymime, do not be discouraged because us chatting here already about some other topics. This particular PR itself seems quite solid, and definetly again improvement.

All good, I've very much across this PR and support it fully. @adbancroft has shown INCREDIBLE patience and commitment by breaking these changes up and keeping at them over such a long period and these will all be brought in, but as always I just need to find the time to sit down an thoroughly test everything like this. Anything that touches the schedulers in particular has the ability to have subtle and difficult to detect impacts, so I like to be paranoid with them. 

I'm expecting to pull this one in sometime this week though.

PR #1115, opened 2023-10-09
Add a gauge to tuner studio to report trigger filter events

mike501:
Enables users to see if a decoder trigger filter is having lots of fails to help diagnose issues and confirm if they've got the right filter setting.

A number of decoders had the formatting corrected when pasting in the 2 lines to count the failures.

NB this is a repeat submission as after updating to the latest release a push from GIT wasn't updating the previous PR.


noisymime:
Line 5172 of speeduino.ini (https://github.com/noisymime/speeduino/pull/1115/files#diff-3a79112606ac8fec5bc323aab7f9332785973542dc4ef9f3675e3ef8afd5e71bR5172) you've added a field called `triggerFilterCounter`. 

Is that there by mistake? If not, it's using the same location as the existing `syncLossCounter` field.

PR #1118, opened 2023-10-11
Added taper entering DFCO

VitorBoss:
Current DFCO is `ON/OFF` and this can be very hard on engine mounts and the driving experience(_fail wife check_).

This add a taper to remove ignition gradually to reduce torque before the fuel cut, a fuel taper is also added for those with knowledge how/when to use it.

Added unit tests for check functionality


noisymime:
Bench testing looks good and unit tests all checkout. Great addition!

Only suggestion would be to add some defaultValue entries in the ini, but I can look at this after merging.

VitorBoss:
Awesome, yeah, I did added on the update function but forgot to add to ini.

Nocnafuria22:
Is there also tapper when exit DFCO? What if Taper advance remove is set to 40 and DFCO occurs when actually ignition advance is for example 20? It will taper to -20? Maybe better option will be "tapper down to x advance (deg)"? What do You think?
BTW Thaks for this feature, with "Map threshold for DFCO" it should work perfect for my car.

VitorBoss:
Yes, it will remove how much you type in, choose wisely, otherwise will give you pops and bangs.
Isn't hard to add a taper exiting DFCO, but that would need a copy of all parameters to be precise.
Without the map threshold it payed very nice on my engine 2 years ago.

PR #1122, opened 2023-10-21
Changes in the idle code new idle feature

155ac:
*Now the threshold in the PID idle will keep the idle valve open for preventing rpm dips when coming back to idle. The threshold function work for Closed loop and Closed+Open loop, when in Closed loop only the TPS threshold will be used but when using Closed+Open loop the RPM threshold and TPS threshold will be used.

*The PWM Closed loop and PWM Closed+Open loop now run in one switch case for saving code space.

*Implements a setting for increasing the idle when the fan is turned on. The idle increase is control by the user using a setting group in the fan menu call fan idle up in there the user can turn on this feature and configure how much the idle will be increase.
fan idle up

*Better arrangement in the idle menu settings.
new idle menu

### DIFF
--- a/.github/workflows/misra.yml
+++ b/.github/workflows/misra.yml
@@ -44,7 +44,7 @@ jobs:
       env:
         MISRA_GIST: ${{ secrets.MISRA_GIST }}
         VIOLATIONS: ${{ env.VIOLATIONS }}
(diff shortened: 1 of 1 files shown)


155ac:
i will convert the code for the cpp file version

Issue #1086, opened 2023-09-02
Idle integral reset rpm hysteresis not working causing stall 


155ac:
The integral reset rpm hysteresis option is not working, making the engine stall in deceleration.

The iac value should be held until rpm drops below the rpm hysteresis setting then the pid should start.

![Capturar](https://github.com/noisymime/speeduino/assets/143891251/de3926fc-7f46-40fa-a918-bbec1fc645c1)
Here the iac value should come from the iac stepper motor until the rpms drop below 1110 rpm, then the pid should start.

Here are my idle configs.
![idle](https://github.com/noisymime/speeduino/assets/143891251/6b337e58-d7b9-4d98-bd0a-d5270cc16d22)

VitorBoss:
That is very odd, the check is in place and is right AFAIK
https://github.com/noisymime/speeduino/blob/master/speeduino/idle.ino#L719
Also, don't use 0 as minimum steps as it will definitely stall the engine, try something like 10.

155ac:
the minimum iac value is near zero because i adjust the throttle stop screw in a way that wen the engine is warm and with no load the value stay close to zero.

the problem is when is load on the engine i stalls because the pid is closing the valve very fast, because ther is no rpm hysteresis. When it should use the table and then it slowly going back to pid control, like how he does when the tps is above the reset. as you can see in the log as i close the throttle it go slowly from table to pid control, but is not doing based on rpm as it should.

Only the tps value is working, i try to increase and decrease the rpm hysteresis in settings, but it's still not working.
![idle error](https://github.com/noisymime/speeduino/assets/143891251/e83f1f6f-6d43-48d4-a6e2-0da7080ed963)

155ac:
I'm also running the latest build 2023.05 and the idle work great in open loop, only in closed loop the problem appears so no hardware problems in the stepper motor and driver.

paoloponzano:
> the minimum iac value is near zero because i adjust the throttle stop screw in a way that wen the engine is warm and with no load the value stay close to zero.
> 
> 
> 
> the problem is when is load on the engine i stalls because the pid is closing the valve very fast, because ther is no rpm hysteresis. When it should use the table and then it slowly going back to pid control, like how he does when the tps is above the reset. as you can see in the log as i close the throttle it go slowly from table to pid control, but is not doing based on rpm as it should.
> 
> 
> 
> Only the tps value is working, i try to increase and decrease the rpm hysteresis in settings, but it's still not working.
> 
[...]

PR #1123, opened 2023-10-21
Idle code changes

155ac:
New features

*The Idle thresholds now will hold the valve open when in using the closed loop and open+closed loop idle for the stepper and PWM versions, when using the open+closed loop the valve can be held open by the TPS threshold and the RPM threshold, but when using only the closed loop only the TPS threshold will old the valve open. This will fix the issue #1086

*Idle taper for all idle types.

*Adds a option for increasing the idle by a set value when the fan is on this works for all idle mode except the fast idle mode.

*Change the check that tunerstudio uses for detecting what type of idle mode is being used for some settings.

*Change some items locations and names in the idle menu settings.

Bug fixes

*Fix a bug that the idle load gauge will not display steps or duty cycle when using Stepper closed+open loop or PWM closed+open loop idle type.

*Fix a bug in the Stepper motor that when the engine stalls makes the valve overshot when the valve closes to the crank position and makes the valve lost the home positioning. This was fix by removing stepper motor code from the function disableIdle, this function is call when the engine stalls but this is not needed for the stepper motor because the idle code for the Stepper motor itself contains the code for closings the valve when the engines stalls. Running the code 2 times makes the function dostep overshot also the disableIdle was setting the valve to the crank value end then setting to 0 back and forth .

*Fix a bug that the duty cycle is not going to 0  when using PWM open+closed loop when the engine stalls and the setting run before start is set to no, this was fix by adding the PWM open+closed loop the check along with all the others PWM idle types for the disableIdle. This code must only execute when using any of the PWM idles because it sets the valve to 0 and disables the valve output. 

This is what the idle menu  and the fan menu with the fan idle up looks like 
new menus


speeduino/idle.cpp
```diff
@@ -485,6 +484,11 @@ void idleControl(void)
       {
         //Currently cranking. Use the cranking table
         currentStatus.idleLoad = table2D_getValue(&iacCrankDutyTable, currentStatus.coolant + CALIBRATION_TEMPERATURE_OFFSET); //All temps are offset by 40 degrees
+        //Add Idle Up amount if active
```
adbancroft, line 487: This block of code is repeated many times. Could you hoist it into a `static inline ` function? E.g. `getIdelUpLoad(void)`

adbancroft (changes requested):

drozdowsky:
@155ac imho these are good changes, any reason this was closed?

Issue #1104, opened 2023-09-27
O2 feedback based on load 


Admanirv123:
Allow o2 feedback based on load and look up of afr table as opposed to just current tps. Will aid spool by running leaner mixture, but suggest 6 or 7 psi max for throughput/cpu limitations of the 2560 in altering mixture in timely manor.

VitorBoss:
Are you using alpha-N fuelling?

Admanirv123:
No I am not.

Admanirv123:
Hi, for example if I had requested 12.6 afr all the way to 150 kpa in afr target I have to be x throttle defined in afr/o2 settings typically 100%. But I don't want to run 12.6 afr all the way in max boost ie target 250kpa in full throttle as its too lean, hence why a switch off would be good ie max o2 closed loop feedback (active below x kpa) so effectively another option 'active below kpa' in afr/o2 option. I wouldn't like to use closed loop fueling all the way to redline at high boost unless the mega can alter and maintain in timely fast manor?

Admanirv123:
Sure I will change, I thought its was clear in the first post esp with max psi in o2/afr, I've had to justify it by more explanation.

PR #1124, opened 2023-10-24
MAP Limits for EGO

155ac:
This will add two new settings for the EGO. The Active Below MAP that will only allow for the ego to run when the current map is below that option. And the Active Above MAP that will only allow for the ego to run when the current map is above that option. These new settings will allow to use the EGO under boost more safely. This will close the issue  #1104

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -1005,8 +1005,8 @@ page = 9
       
       unused10_110         = scalar, U08,     110,        "",       1, 0, 0, 255, 0 
       unused10_111         = scalar, U08,     111,        "",       1, 0, 0, 255, 0
-      unused10_112         = scalar, U08,     112,        "",       1, 0, 0, 255, 0
-      unused10_113         = scalar, U08,     113,        "",       1, 0, 0, 255, 0
+      egoMAPMax = scalar, U08, 112, "kPa", 2.0, 0.0, 2.0, 511.0, 0
+      egoMAPMin = scalar, U08, 113, "kPa", 2.0, 0.0, 2.0, 511.0, 0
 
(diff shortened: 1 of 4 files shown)


noisymime:
Looks good. I would suggest changing the values in `updates.ino` to be:

```
    configPage9.egoMAPMax = 255,
    configPage9.egoMAPMin = 0,
```

The values you've got are good as defaults (So leave the defaults you have in the ini file), but they would modify the existing behaviour for anyone upgrading, which is best to avoid. There may be cases where people are relying on this operating above 100kPa, which would get disabled with those values.

155ac:
> Looks good. I would suggest changing the values in `updates.ino` to be:
> 
> ```
>     configPage9.egoMAPMax = 255,
>     configPage9.egoMAPMin = 0,
> ```
> 
> The values you've got are good as defaults (So leave the defaults you have in the ini file), but they would modify the existing behaviour for anyone upgrading, which is best to avoid. There may be cases where people are relying on this operating above 100kPa, which would get disabled with those values.

The configPage9.egoMAPMax = 255 i think is a bit to big because it get multiply by 2 so it will be 510 kpa I don't think anyone would be crazy enough to use the ego under so much boost

155ac:
> Looks good. I would suggest changing the values in `updates.ino` to be:
> 
> ```
>     configPage9.egoMAPMax = 255,
>     configPage9.egoMAPMin = 0,
> ```
> 
> The values you've got are good as defaults (So leave the defaults you have in the ini file), but they would modify the existing behaviour for anyone upgrading, which is best to avoid. There may be cases where people are relying on this operating above 100kPa, which would get disabled with those values. 

You are right. I fix my mistake,I don't know what I was thinking when I wrote that part of the code.

PR #1134, opened 2023-11-07
Prevent possibility of some ignition output sticking ON after switching from half to full sync.

TBAMax:
Add check for running schedules. So that switching from half to full sync only happens when there is no schedules currently running. This prevents possibility of some outputs sticking open for whole engine cycle immediately after switching.

This is replacement for now closed #1069
With the new style copied from the new implementation of the same thing for the injection schedules.

Please review before pulling because it is not tested, but is purely theoretical, and implemented quickly.

If this is not considered essential, please close with the comment, so that I do not have to maintain it further.
Best regards

### DIFF
--- a/speeduino/init.cpp
+++ b/speeduino/init.cpp
@@ -3653,6 +3653,32 @@ static inline bool isAnyFuelScheduleRunning(void) {
       ;
 }
 
+static inline bool isAnyIgnScheduleRunning(void) {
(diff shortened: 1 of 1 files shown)


adbancroft:
I realize you are following the previous commits style, but could/should all these checks be enclosed in a nointerupts/interupts pair?

PR #1146, opened 2023-12-11
Add support for RusEFI CAN Wideband based on STM32F042

dynfer:
This PR allows the use of the CAN controller created by the RusEFI project based on the stm32f042 and stm32f103.
The module works on ID 400 and can be further expanded for example to enable the heater once the engine is running.


speeduino/comms_CAN.cpp
```diff
@@ -36,7 +36,17 @@ void sendVAGCluster()
   DashMessage(CAN_VAG_VSS);
   Can0.write(outMsg);
 }
-
+void reciveCANwbo()
+  {
+    if (inMsg.id == 400 && configPage15.canREWBOCAN == true) // RusEFI CAN Wideband support
```
mck1117, line 41: I'd love if you could throw a link to my repo in here (I'm the creator): https://github.com/mck1117/wideband

dynfer (commented):
@mck1117 here You go and thanks for Your work!

mck1117:
I would also recommend at least checking the valid bit. I'm not sure if Speeduino has support for "this sensor is present but currently not working" or not, but blindly reading the lambda field without first checking validity (at least to set lambda=0 or something) could have unexpected behavior.

dynfer:
@mck1117 what is the expected range of the controller? 0.5 - 1.4?

speeduino/comms_CAN.cpp
```diff
@@ -36,7 +36,26 @@ void sendVAGCluster()
...
 }
-
+void reciveCANwbo() // RusEFI CAN Wideband support https://github.com/mck1117/wideband
+  {
+    if(configPage15.canREWBOCAN == true && BIT_CHECK(currentStatus.engine, BIT_ENGINE_RUN)) // Enable heater once engine starts to avoid heater running on cold engine with ignition on
+    {
+      outMsg.id = 0x190;
+      outMsg.len = 2;
+      outMsg.buf[0] = currentStatus.battery10; // We don't do any conversion since factor is 0.1 and speeduino value is x10
+      outMsg.buf[1] = 0x1; // Heater enable bit 0 - off 1 - on
```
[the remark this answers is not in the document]
dynfer: Thank You for the insight changed it :)

reference/speeduino.ini
```diff
@@ -2426,6 +2424,7 @@ menuDialog = main
   tachoMode       = "The output mode for the tacho pulse. Fixed timing will produce a pulse that is always of the same duration, which works better with mode modern digital tachos. Dwell based output creates a pulse that is matched to the coil/s dwell time. If enabled the tacho pulse duration and timing is same as coil dwell and the number of pulses is same as number of ignition events. This can work better on some styles of tacho but note that the pulse duration might become problem on higher cylinder number engines."
   canBMWCluster   = "Enables CAN broadcasting for BMW E46, E39 and E38 instrument clusters with message ID's 0x316, 0x329 and 0x545"
   canVAGCluster   = "Enables CAN broadcasting for VAG instrument clusters with message ID's 0x280 and 0x5A0"
+  canREWBOCAN     = "Enables to recive AFR via CAN from the Rusefi Wideband module"
```
rusefillc, line 2427: s/Rusefi/rusEFI/

reference/speeduino.ini
```diff
@@ -3563,6 +3562,7 @@ menuDialog = main
     dialog = CanBcast, "CAN Broadcasting menu"
         field = "BMW Instrument Cluster Broadcast",    canBMWCluster
         field = "VAG Instrument Cluster Broadcast",    canVAGCluster
+        field = "RusEFI WBO CAN",    canREWBOCAN
```
rusefillc, line 3565: s/RusEFI/rusEFI/

speeduino/comms_CAN.h
```diff
@@ -21,6 +21,7 @@ int CAN_read();
 void CAN_write();
 void sendBMWCluster();
 void sendVAGCluster();
+void reciveCANwbo();
```
rusefillc, line 24: receive?

rusefillc:
This is somewhat compatible with AEM CANbus protocol it should probably be mentioned somewhere?

dynfer:
> This is somewhat compatible with AEM CANbus protocol it should probably be mentioned somewhere?

Yes we discussed this already with @noisymime to make it more general to allow for easier use of other CAN wbo solutions like AEM or Ecumaster.

Also thanks for pointing out the typo and case mismatches.

speeduino/comms_CAN.h
dynfer: done

reference/speeduino.ini
dynfer: done

reference/speeduino.ini
dynfer: done

mck1117:
> This is somewhat compatible with AEM CANbus protocol it should probably be mentioned somewhere?

No, it isn't. That protocol was abandoned in favor of the current one.

noisymime:
I made some small changes to this so that it can be expanded to other module in the future (IE the rusEFI module is just the first and only in the list, but others can be added now). 

One thing I did come across when looking at this though is that the check for the CAN msg ID is separate from the other CAN reads, which makes me wonder what happens if this is on a bus that has many other messages on it. 

If there is already another msg waiting in the CAN buffer, the check for id's 0x190 or 0x192 will always fail. At best it will mean that the O2 signal is not read until the next 30Hz cycle, at worst they may never get read. 

Has this been tested with other CAN devices broadcasting or on an isolated bus only?

reference/speeduino.ini
```diff
@@ -476,8 +476,7 @@ page = 1
       canBMWCluster             = bits,   U08,  126, [0:0], "Off", "On"
       canVAGCluster             = bits,   U08,  126, [1:1], "Off", "On"
       ;These are reserved for future use, in case of more CAN broadcasting features are added
-      canREWBOCAN               = bits,   U08,  126, [2:2], "Off", "On"
-      enable2Cluster            = bits,   U08,  126, [3:3], "Off", "On"
+      canWBO                    = bits,   U08,  126, [2:3], "Off", "RusEFI WBO", "INVALID", "INVALID"
```
rusefillc, line 479: please spell "rusEFI"

dynfer:
@mck1117 do You recall the update rate of the controller? 50hz?

PR #1148, opened 2023-12-13
Add support for BlitzBox hardware in speeduino firmware

chaoschris:
As our hardware development for BlitzBox comes to an end we want to add support for BlitzBox hardware in speeduino firmware.

https://github.com/oelprinz-org/BlitzboxBL49sp


noisymime:
Looks mostly fine. What I'd suggest doing though is adding some default pins for a few of the aux functions as well. 

Eg: Things like VVT allow the use to select 'Board Default' in Tuner Studio, which will cause problems with this current mapping because you haven't defined a default `pinVVT_1`

chaoschris:
I am not sure i have understood you right. You want me to add/activate the unused functions of the board and set them in the pin layout to a "fictive" MCU pin which isn't in use? Like in this proposal:

     //Pin mappings for all BlitzboxBL49sp variants
      pinInjector1 = 6; //Output pin injector 1
      pinInjector2 = 7; //Output pin injector 2
      pinInjector3 = 8; //Output pin injector 3
      pinInjector4 = 9; //Output pin injector 4
      pinCoil1 = 24; //Pin for coil 1
      pinCoil2 = 25; //Pin for coil 2
      pinCoil3 = 23; //Pin for coil 3
      pinCoil4 = 22; //Pin for coil 4
      pinTrigger = 19; //The CRANK Sensor pin
      pinTrigger2 = 18; //The Cam Sensor pin
      pinFlex = 20; // Flex sensor PLACEHOLDER value for now
      pinTPS = A0; //TPS input pin
      pinSpareTemp1 = A1; //LMM sensor pin
      pinO2 = A2; //O2 Sensor pin
      pinIAT = A3; //IAT sensor pin
      pinCLT = A4; //CLT sensor pin
      pinMAP = A7; //internal MAP sensor
      pinBat = A6; //Battery reference voltage pin
      pinBaro = A5; //external MAP/Baro sensor pin
      pinO2_2 = A9; //O2 sensor pin (second sensor) PLACEHOLDER value for now
      pinLaunch = 2; //Can be overwritten below
      pinTachOut = 10; //Tacho output pin
      pinIdle1 = 11; //Single wire idle control
      pinIdle2 = 14; //2 wire idle control PLACEHOLDER value for now
      pinFuelPump = 3; //Fuel pump output
      pinVVT_1 = 15; //Default VVT output PLACEHOLDER value for now
      pinBoost = 13; //Boost control
      pinSpareLOut1 = 49; //enable Wideband Lambda Heater
      pinSpareLOut2 = 16; //low current output spare2 PLACEHOLDER value for now
      pinSpareLOut3 = 17; //low current output spare3 PLACEHOLDER value for now
      pinSpareLOut4 = 21; //low current output spare4 PLACEHOLDER value for now
      pinFan = 12; //Pin for the fan output
      pinResetControl = 46; //Reset control output PLACEHOLDER value for now

noisymime:
Yep, that should be fairly complete. 

Think of the scenario where someone enables VVT, which will go to Board Default as standard. If you don't have a pin set for that then it will default to pin 0, which is one of the comms pins. So potentially the users TunerStudio connection dies and then they are unable to reconnect because the board is trying to use VVT on the serial RX pin. You need to have safe board defaults for this reason

chaoschris:
Thanks for the detailed explication. I have changed it, is now a new pull request necessary? Sorry for the dumb question, this is my first try doing such a thing in GitHub...  ;-)

PR #1157, opened 2023-12-28
Sensors.cpp: MISRA fixes & reduce conditional compilation blocks

### LINKED ISSUE #1156: Bug: analog sensor ISR conditional compilation uses wrong preprocessor symbol(s)
In sensor.cpp, `AnChannel[]` is conditionally defined & declared using the `ANALOG_ISR` symbol. However, the code that uses `AnChannel[]` is conditionally compiled using _one of _2__ _symbols_:

- `ANALOG_ISR` (E.g. `readTPS()`); or
- `ANALOG_ISR_MAP` (E.g. `readBaro()`).

Shouldn't they all use the same symbol? (`ANALOG_ISR`)

### LINKED ISSUE #1251: MAP - non-Instantaneous sampling methods skip readings
Cycle Average, Cycle Minimum & Event Average do not read the MAP sensor(s) when the cycle/event ends. E.g.  using cycle average a call to readMap at the end of the cycles does not read the sensors.

This is probably really minor.....

adbancroft:
- Limit scope of conditional compilation by hoisting duplicate sensor read code blocks into functions: `readAnalogSensor`, `readMAPSensor`, `readAnalogPin`
- Fix all MISRA issues in sensor.cpp.
  - Necessitated replacing `fastMap10Bit` & `ADC_FILTER` macros with inline functions to enforce integer type consistency. **Worth double checking this (even though they are unit tested).**
- Remove a few unused globals (saves a few bytes RAM)
- Made MAP reading modes consistent (mostly around error handling). This required de-duplicating the MAP reading code, which in turn split apart readMAP() into mode/algorithm specific functions. 
  - Added unit tests for the majority of the MAP read/sampling code.
- forced loop() inline - saves RAM
- Pulled in the SimplyAtomic library so we can use a consistent, reliable and cross-platform ATOMIC idiom to prevent tearing of some volatile globals during read.
- pulled duplicate tooth based engine running check into engineIsRunning() in decoders.*
- separated baro init out from readMap & readBaro -simplifies both & slightly faster

Fixes #1251


noisymime:
`static uint8_t a=0U, x=0U, y=0U, z=0U;`

Moving the state variables of the random function inside the function itself means that the random result is essentially resetting every call and just returning a variation of the value in micros. 

As `micros()` only has a resolution of 4uS on AVR boards and the `random1to100()` function is called within a fairly fast running loop, this is likely to return the same value multiple times in row. Despite the extra memory required, these state variables need to keep their values outside the function

adbancroft:
They are declared static within the function, so they are only intitialised the first time the function is called & retain their value between successive calls: [https://stackoverflow.com/a/12186903](https://stackoverflow.com/a/12186903).

Happy to undo this though - it doesn't save any memory, just limits the scope of the variables to where they are used.

noisymime:
@adbancroft Sorry to be a pain, but are you able to double check the multiplies that have been added in `updates.cpp`? They don't seem to align with the behaviour that is there current. 

**Eg** Quite a number of multiplies appear to have been added to the 202402 updates that were previously in different releases. 

`vvt2Table` Y axis is now multiplied by 2 in both the 202202 and 202402 updates. 202402 also multiplies its X axis and values by 2, which wasn't being done previously. 

I may have missed something about this change, but this didn't appear correct at first glance?

adbancroft:
> vvt2Table Y axis is now multiplied by 2 in both the 202202 and 202402 updates. 202402 also multiplies its X axis and values by 2, which wasn't being done previously.

In the 202402 update code the old function, `multiplyTableValue()`, was misnamed: it actually multiplied *all* bytes on a page (`multiplyPage` might have been a better name). For the `wmiMapPage`, this means the *x axis, y axis & values* of `wmiTable`, `vvt2Table` & `dwellTable` were all multiplied.

I don't know if that was the intent, but I preserved that behavior.

adbancroft:
I have reduced the scope of this PR - the storage related changes really shouldn't have been included. Also restricted scope of MISRA fixes

adbancroft:
Added unit tests for the MAP sampling algorithms and some of the new functions. E.g. LOW_PASS_FILTER, fastMap10Bit

speeduino/decoders.cpp
```diff
@@ -335,6 +337,23 @@ static inline bool IsCranking(const statuses &status) {
   return (status.RPM < status.crankRPM) && (status.startRevolutions == 0U);
 }
 
+bool engineIsRunning(uint32_t curTime) {
+  // Check how long ago the last tooth was seen compared to now. 
+  // If it was more than MAX_STALL_TIME then the engine is probably stopped. 
+  // toothLastToothTime can be greater than curTime if a pulse occurs between getting the latest time and doing the comparison
+  ATOMIC() {
+    return (toothLastToothTime > curTime) || ((curTime - toothLastToothTime) < MAX_STALL_TIME);
```
TBAMax: There is a risk that interrupt state is not restored correctly.
image

PR #1167, opened 2024-01-22
FIX A/C IDLE UP

nuosu:


noisymime:
Have you confirmed that this works? At a quick glance, it looks like this is just moving the AC idle adder outside the engine running check (IE it will run even when the engine is off). 

Is that correct? 
If so, that seems like a strange solution as this should only be getting used when the engine is on

nuosu:
> Have you confirmed that this works? At a quick glance, it looks like this is just moving the AC idle adder outside the engine running check (IE it will run even when the engine is off).
> 
> Is that correct? If so, that seems like a strange solution as this should only be getting used when the engine is on


In the void airConControl(void) function, A/C Idle UP is only activated when the engine is running. Therefore, in the idle.cpp file, there is no need to redundantly check whether the engine is running.
In fact, my code modifications have not been tested yet. I simply referred to the OLCL PWM idle section of the code and made changes. I will test it as soon as possible. My coding skills are limited, so I believe there must be better solutions to address this issue.

PR #1170, opened 2024-02-02
Fix eeprom initialization and div by 0 on new board

VitorBoss:
Currently the supported flash is locked to W25Q16 and this PR enables it to use any windbond flash from W25Q80, W25Q16, W25Q32, W25Q64 and W25Q128.

When start the board from an empty eeprom all pages are set to 0 and that causes a divide by 0 on missing tooth decoder initialization, now a valid value is set to avoid infinite loop due to exception.

@noisymime This is a critical problem right now, please pull it


VitorBoss:
@adbancroft Fixed what you pointed out last month

PR #1171, opened 2024-02-03
Stepper enable invert option

fdasbfg:
I built a custom speeduino board with MP6500 ic for the stepper driver, which uses a high level to enable the driver instead of a low signal like the drv8825. These changes add an option in tunerstudio to fully integrate this driver.


adbancroft:
```
    if(configPage9.iacStepperEnInv) {
      digitalWrite(pinStepperEnable, HIGH);
    } else {
      digitalWrite(pinStepperEnable, LOW); //Enable the DRV8825
    }
```

Could you pull out this repeated code block into a separate function?

PR #1185, opened 2024-02-21
Conditional heater enable CANwbo

dynfer:
Change the current behaviour to prevent the heater being enabled all the time, instead it will activate once the engine goes above into running state.


noisymime:
You can just use BIT_CHECK() here without the ternary operator. Bit_CHECK returns a boolean true or false value, so no need for the additional check

PR #1187, opened 2024-02-27
Performance: optimized 32-bit shifts

adbancroft:
We use 32-bit shifts heavily for time (microseconds) calculations. 

As of AVR-GCC 13.2.0, the assembly code produced for most 32-bit shifts is very poor: either too many instructions or falls back to a loop. However, clang produces great quality AVR shift code. See comparison [here](https://godbolt.org/z/71cPnMYqs). 

This PR takes the clang assembly code & packages it into a set of templates (`lshift, rshift`) covering shifts from 1 to 15 bits.

5% speed improvement.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -1,4 +1,4 @@
-# Doxyfile 1.9.4
+# Doxyfile 1.10.0
 
 # This file describes the settings to be used by the documentation system
 # doxygen (www.doxygen.org) for a project.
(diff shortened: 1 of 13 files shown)


noisymime:
I had been holding off on this as there had been some hope that these AVR improvements might make it into gcc 14 similar to clang, but alas not.

Great work on this PR, a solid improvement in performance.

PR #1189, opened 2024-03-01
Comment typos correction

dantob:


dantob:
This should also apply to 202402-fixes, if you want it there

noisymime:
These ones won't go into -fixes as that is strictly for things that are directly correcting a bug. As harmless as any other PR may seam (And this is about as harmless as it gets), this is still categorised more as an enhancement rather than a bug fix. 

Nothing against the PR, just the policy for what goes into -fixes and what doesn't

PR #1190, opened 2024-03-01
add codespell CI action

dantob:
Adds CI action to run codespell against the master branch.

Pulls in latest codespell dictionaries.
Ignore list at `.github/workflows/codespell-ignored-words.txt`

### DIFF
--- a/.github/workflows/codespell-ignored-words.txt
+++ b/.github/workflows/codespell-ignored-words.txt
@@ -1,0 +1,6 @@
+fram
+iterm
+ntegral
+numer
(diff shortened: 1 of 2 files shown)


dantob:
This should actually pass now, it's up to you though which branches you want the CI to run on. It's currently set to master and all pull requests against master or manual dispatch.

noisymime:
Can you please move the ignored list into the `.github/workflows` directory? I think that makes more sense than creating a new one just for it. 

Otherwise looks good.

PR #1193, opened 2024-03-07
Faster switching between accel and decel

155ac:
Comparing TPS DOT and MAP DOT to 0 to make the switching between accel and decel more faster and predictable. This will also will improve throttle response when changing gears.


noisymime:
Can you explain a little more exactly what you're trying to achieve here as it looks redundant at first glance. The code you're modifying is there to check for increases in the AE rate when it's already active and this change will break that functionality.

You also have a problem in your TPS section where you're comparing it to the `MAE` value not the `TAE` one.

155ac:
> Can you explain a little more exactly what you're trying to achieve here as it looks redundant at first glance. The code you're modifying is there to check for increases in the AE rate when it's already active and this change will break that functionality.
> 
> You also have a problem in your TPS section where you're comparing it to the `MAE` value not the `TAE` one.

Thanks for pointing the problem in the TPS part. 

The way the code works is basically removes a problem. Let's say that you are driving and closes the throttle creating a DOT value of -350 and the decel is flag, then you accelerate creating a dot of 200 but the original speeduino code will not flag a accel until the decel time runs out because it will only flag a accel if the DOT is bigger then the 350, creating a flat spot in the acceleration.

Whit the 0 crossing check it will change bethinking accel or a decel immediately if the dot is higher then the threshold, this will not create a flat spot when accelerating or changing gears, because it will end the decel and immediately flags a accel and if you are accelerating with a flag accel and you closes the throttle it will change to the decel much more quick.

PR #1194, opened 2024-03-14
Small bug fix, to correct TPS value over OBD2

pazi88:
OBD2 TPS PID, still assumed that Speeduino uses 0-100 range for TPS, when it's now 0-200. So everything above 50% TPS was shown as 100% over OBD2. This fixes the bug.


noisymime:
Good pickup! I'm guessing this also need changing on the broadcast side of things in the same file? Eg:

`temp_TPS = map(currentStatus.TPS, 0, 100, 0, 254);//TPS value conversion (from 0x00 to 0xFE)`

Should have 200 there rather than 100?

pazi88:
You are right. Also tje low end is also slightly wrong. I'll fix that on this PR.

PR #1197, opened 2024-03-15
Fix Stepper idle cranking outputting half steps

VitorBoss:
Just hooked up my counter to stepper control pins and found an error while on cranking mode, it was counting twice per call due to `doStep()` been called twice on that condition.
Changed `doStep()`  function checks to be easier to read and debug.

Removed unnecessary RAM use to store stepper direction with inline conditions, this free 2 bytes, not a huge deal.

### DIFF
--- a/speeduino/idle.cpp
+++ b/speeduino/idle.cpp
@@ -196,16 +196,6 @@ void initialiseIdle(bool forcehoming)
         idleStepper.stepperStatus = SOFF;
       }
 
-      if (! configPage9.iacStepperInv)
-      {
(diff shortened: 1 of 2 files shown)


noisymime:
In `doStep()` I think you need to keep the check on the else clause so that it only does a step if error > 1. If I'm reading this right, if error is 0 (ie we're at the target step) it would add another step on?

VitorBoss:
I don't think the check is needed as it will fall there only when is greater than 0 because previous checks is below `-iacMinSteps` or above `iacMinSteps`, the effect will be only when the `error > iacMinSteps`

PR #1203, opened 2024-04-05
fix stepper driver always enabled causing motor to run hot

john1786:
Currently when using "Stepper Open Loop", the DRV8825 often remains enabled even though Stepper Power is set "When Active" which causes the stepper motor to run unnecessarily hot.  Specifically,  the idle step hysteresis [when Minimum Steps > 1] is not accounted for when determining if the final step of the cycle has been reached so the DRV8825 ENABLE pin often never transitions from LOW to HIGH.


noisymime:
Tested and works great. Thanks for the submission!

PR #1209, opened 2024-04-25
modified poll level mode

adamg-86:
Added the option to poll at any teeth of the trigger wheel. This way you can use the stock cas in conjunction with a missing tooth trigger wheel to operate in sequential mode.  I know the config page 10 is not the best for the trigger settings but the page 4 is full. Please chime in to make the code better I am not a programmer, I just wanted to use my stock cas.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -1288,7 +1288,9 @@ page = 10
 
       oilPressureProtTime   = scalar, U08,    190, "seconds", 0.1, 0.0, 0.0, 25, 1
 
-      unused11_190_191      = array,  U08,    191,  [1], "RPM",  100.0,      0.0,  100,     25500,    0
+      PollLevelTeeth  = scalar,  U08,      191,  "",        1,      0,   1,      255,    1
+
+      ;unused11_190_191      = array,  U08,    191,  [1], "RPM",  100.0,      0.0,  100,     25500,    0
 
 ;Page 11 is the fuel map and axis bins only
 page = 11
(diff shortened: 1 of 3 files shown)


pazi88:
Have you checked how much this affects loop speeds on ie. 60-2 trigger pattern at high RPM? I remember that I left out option to select the primary trigger tooth for poll level, because it affected loop speeds too much for my liking.

adamg-86:
Good point. I tested it and it seems to have little to no effect on loop speed. 
First image is last firmware at 10 000RPM missing tooth 

image

Second image is with polling at 10 000RPM 
image

At 2500RPM same story about 900

I am using the polling only to achieve sync, and then it flip-flops between revolutionOne = 1 and revolutionOne = 0

Maybe I am missing something

speeduino/decoders.cpp
```diff
@@ -500,6 +500,15 @@ void triggerPri_missingTooth(void)
       {
         bool isMissingTooth = false;
 
+        if(( (currentStatus.hasSync == false) || (currentStatus.RPM < 2000)))
+        {
+          if ((configPage4.trigPatternSec == SEC_TRIGGER_POLL) && (configPage10.PollLevelTeeth > 1) && (toothCurrentCount == configPage10.PollLevelTeeth)) // at selected tooth check if the cam sensor is high or low in poll level mode
+          {          
+            if (configPage4.PollLevelPolarity == READ_SEC_TRIGGER()) { revolutionOne = 1; }
```
adbancroft, line 507: Isn't this the same as `revolutionOne = configPage4.PollLevelPolarity == READ_SEC_TRIGGER();`

speeduino/decoders.cpp
```diff
@@ -500,6 +500,15 @@ void triggerPri_missingTooth(void)
...
+            else { revolutionOne = 0; }
```
adbancroft, line 508: Isn't this the same as `revolutionOne = configPage4.PollLevelPolarity == READ_SEC_TRIGGER();` ?

reference/speeduino.ini
```diff
@@ -2060,7 +2062,8 @@ menuDialog = main
   TrigPattern       = "The type of input trigger decoder to be used."
   useResync         = "If enabled, sync will be rechecked once every full cycle from the cam input. This is good for accuracy, however if your cam input is noisy then this can cause issues."
   trigPatternSec    = "Cam mode/type also known as Secondary Trigger Pattern."
-  PollLevelPol      = "The level of the cam trigger input will be checked at tooth #1 and this defines if the level is supposed to be High or Low at 1st phase of the engine."
+  PollLevelPol      = "The level of the cam trigger input will be checked at tooth # and this defines if the level is supposed to be High or Low at 1st phase of the engine."
+  PollLevelTeeth    = "The crank teeth at witch the level of the cam trigger will be checked"
```
adbancroft, line 2066: "witch" -> "which"?

adamg-86: this makes it that you can choose at witch teeth you look a the cam signal. Not only at the first tooth. For example on a miata you can use a 36-1 crank wheel and the stock cas for sync if you look at a tooth that is low one revolution and high the other
Screenshot 2024-04-25 093810

PR #1212, opened 2024-05-09
Provide hook for developer local PIO configuration

adbancroft:
Occasionally I need to change the PIO configuration for testing or analysis purposes. It'd be useful to keep those changes, but they are only used by me so shouldn't be checked in. This leads to 2 possible problems:

1. Accidentally committing `platform.ini` as part of other changes
2. Changing `platform.ini` triggers my dev tooling (VS Code Git plugin) into showing `platform.ini` as changed (annoying)

This PR allows an extra PIO config file (`local.ini`) that is ignored by Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ speeduino.ino.cpp
 test/output_export.cpp
 misra/.results
 ~$*
+local.ini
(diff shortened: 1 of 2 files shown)


noisymime:
Great idea, I've nearly done the same thing myself once or twice.

PR #1215, opened 2024-05-22
Fix coil 6 declaration for STM32F407 SPECTRE board

VitorBoss:
Currently the coil 6 pin definition on STM32F407 SPECTRE board(case 60).

This is critical, user reported this issue kill his coil.

### DIFF
--- a/speeduino/init.cpp
+++ b/speeduino/init.cpp
@@ -2568,7 +2568,7 @@ void setPinMapping(byte boardID)
         // = PB4;  //(DO NOT USE FOR SPEEDUINO) SPI1_MISO FLASH CHIP
         // = PB5;  //(DO NOT USE FOR SPEEDUINO) SPI1_MOSI FLASH CHIP
         // = PB6;  //NRF_CE
-        // = PB7;  //NRF_CS
+        pinCoil6 = PB7;  //NRF_CS
         // = PB8;  //NRF_IRQ
         pinCoil2 = PB9; //
         // = PB9;  //


noisymime:
Just to check, is these needed for the other stm32 variants also? 

AS a 2nd question, are those variants even needed for this Spectre board?

VitorBoss:
No, this is a fix for the SPECTRE board only.
Which variants?

noisymime:
I just noticed under the Spectre board ID (60) in init that there were variants for `STM32F411`, `STM32F401` and `CORE_STM. This fix looks to be for `STM32F407` only?

VitorBoss:
Yes, the fix is a very simple missing definition, the F4x1 is there too because on TS there was just one STM32 board listed when I added F4x1 boards.
I'm adding another one, if you want I can separate them and add case 61 for F4x1 boards.

VitorBoss:
@noisymime Just to be clear, those definitions was added when I was adding support for those boards with @Tjeerdie and they stayed pretty much the same for the last 5 years.

noisymime:
> @noisymime Just to be clear, those definitions was added when I was adding support for those boards with @Tjeerdie and they stayed pretty much the same for the last 5 years.

That's fine, I just wanted to check whether you needed to define a `pinCoil6` for those use cases on this board as well. If not I'll merge this in.

VitorBoss:
The change is only for the F407 and it is ready to be merged, is a very simple change to be exactly

PR #1221, opened 2024-06-20
Added Trigger for the Yamaha XTZ750

giovannni97:
This is a trigger I`ve been riding around with for about a year based on the Vmax one from @RempageR1 . The XTZ flywheel has 4 unevenly sized lobes, where one is the wide lobe used for triggering. This trigger therefore has a routine to note the times when an upward and downward tooth edge is seen and can therefore calculate the tooth width. The actual trigger needs to happen on the upward tooth edge so this is included in the routine.

IMPORTANT: The VR conditioner must have the zero-crossing disabled in order to have reliable triggering

This is my first commit to the Speeduino project, please let me know if there is anything I need to change or of you need any additional information. Thank you!

### DIFF
--- a/speeduino/decoders.cpp
+++ b/speeduino/decoders.cpp
@@ -5995,5 +5995,164 @@ void triggerSetEndTeeth_SuzukiK6A(void)
   { tempIgnitionEndTooth = 7; } // didn't find a match, use tooth 7 as it must be greater than 7 but less than 1.  
   ignition3EndTooth = tempIgnitionEndTooth;
 }
+
+
+
+
+/** Yamaha XTZ750 1989+ with 4 uneven teeth, triggering on the wide lobe.
+Within the decoder code, the sync tooth is referred to as tooth #1. Derived from Harley and made to work on the Yamaha XTZ.
+Trigger is based on 'CHANGE' so we get a signal on the up and downward edges of the lobe. This is required to identify the wide lobe.
+* @defgroup dec_xtzV1 Yamaha XTZ V1
+* @{
+*/
+void triggerSetup_XTZ(void)
+{
+  triggerToothAngle = 0; // The number of degrees that passes from tooth to tooth, ev. 0. It alternates uneven
+  BIT_CLEAR(decoderState, BIT_DECODER_2ND_DERIV);
+  BIT_CLEAR(decoderState, BIT_DECODER_IS_SEQUENTIAL);
+  BIT_CLEAR(decoderState, BIT_DECODER_HAS_SECONDARY);
+  //MAX_STALL_TIME = (3333UL * 60); //Minimum 50rpm. (3333uS is the time per degree at 50rpm)
+  MAX_STALL_TIME = ((MICROS_PER_DEG_1_RPM/50U) * 60U); //Minimum 50rpm. (3333uS is the time per degree at 50rpm)
+  if(currentStatus.initialisationComplete == false) { toothLastToothTime = micros(); } //Set a startup value here to avoid filter errors when starting. This MUST have the initi check to prevent the fuel pump just staying on all the time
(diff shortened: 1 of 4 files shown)


speeduino/decoders.cpp
```diff
@@ -5995,5 +5995,164 @@ void triggerSetEndTeeth_SuzukiK6A(void)
...
+* @defgroup dec_xtzV1 Yamaha XTZ V1
+* @{
+*/
+void triggerSetup_XTZ(void)
+{
+  triggerToothAngle = 0; // The number of degrees that passes from tooth to tooth, ev. 0. It alternates uneven
+  BIT_CLEAR(decoderState, BIT_DECODER_2ND_DERIV);
+  BIT_CLEAR(decoderState, BIT_DECODER_IS_SEQUENTIAL);
+  BIT_CLEAR(decoderState, BIT_DECODER_HAS_SECONDARY);
+  //MAX_STALL_TIME = (3333UL * 60); //Minimum 50rpm. (3333uS is the time per degree at 50rpm)
```
adbancroft, line 6014: Could you remove the commented out code?

giovannni97: Hi,
I hope I have done it right

DFlak:
Hey,
Im trying to use your decoder with a TDM850 3VD, from the scope output I see that it has two small lobes, one medium (?) and one long. Your decoder corretly calculates RPM however I get increments in sync loss. I will try to increase the 3x small lobe multiplier to something more like 5 or 6 times more, hopefully it correctly assigns the "medium" lobe as a short one. Could you explain a bit about the disabling of zero crossing in the VR conditioner, what is it supposed to accomplish? In the datasheet of MAX9926 there is no easy option to disable zero crossing (mode a2), you would also have to provide an external bias voltage and disable adaptive peak threshold.

giovannni97:
Hi,
the issue I had revolved around the zero crossing. Basically when a tooth the VR sensor generates a rapid positive spike that gets rapidly back to zero due to the parallel resistor mounted on the VR decoder. The MAX9926 interpret the positive spike and turns it into a positive output UNTIL it detects a "negative" one, the problem is that the negative one, due to the zero crossing feature enabled, is almost always due to noise, making it effectively uncapable of detecting the true negative spike that comes a lot later due to the long tooth my (and yours) decoder have. From what I understand looking online at pictures of the flywheel on the TDM850 it should have the same tooth pattern as mine, it is the same engine after all.
In order to disable the zero crossing I had to put the MAX9926 into C mode by lifting the pins and connecting some flying wires (the potentiometer was to rapidly adjust the signal threshold since mode C disable the adaptive peak threshold, I basically turned it into a glorified differential window comparator with hysteresis).
On my new ECU I implemented the C mode directly onto the PCB itself.
The photo is from my old and pretty messed up first ECU attempt.
PXL_20250718_175129639

DFlak:
Thanks for the reply! It clarified a lot. Are you running fuel only or do you have ignition working with this setup? If you have ignition how stable is it? 
I'm asking because that is a lot of work and would require me to redesign a pcb and order new ones, I'm thinking of picking up signal from negative of the ignition coil but then I only have the option to run fuel only. I already have some optoisolator cobbled together but it requires me to desolder the VR conditioner and I'm wondering which option is better.

giovannni97:
Hi,
before the modification I only managed to have fuel working but with startup and idle issues. After the modification it seemed a new bike and I managed to remove the stock ecu and do both ignition and injection with Speeduino.
As a note for the ignition i used the ISL9V3040 directly connected to the microcontroller and the coils and had no issues since it takes care of the voltage spikes of the coils.
If you want to modify an existing MAX9926 it is not so difficult if you have access to a good soldering station, you need to lift up the pins you want to disconnect and re-wire them.

PR #1222, opened 2024-06-22
Introduced maximum MAP threshold for decel enleanment

russellE46:
I had been experiencing some rich stumbles when letting off of the throttle, so I decided to give decel enleanment a try. A value of 80% worked very well for city driving. Completely took care of the stumble. However, I noticed that when I lifted just a little bit under boost, decel enleanment kicked in and I was left with 80% of my usual fueling. 13.6:1 AFR near 300 kPa. I have added a MAP limit for decel enleanment to prevent this from happening. Without this feature, decel enleanment is unusable to me.

image

image


joeclone:
@russellE46 if you're able to integrate [https://speeduino.com/forum/viewtopic.php?p=66295#p66295](url) as well I think it'd work better overall

russellE46:
@joeclone that link is dead. Is it another PR?

russellE46:
Oh I see now. I had to manually copy the link for some reason. I can add that too

Issue #1218, opened 2024-06-07
Incorrect BMW Can broadcast Coolant value conversation.


Zazins:
Incorrect Can message conversation leads to incorrect coolant temperature displayed.
Incorrect value is more then half wrong from correct vales. 

Correct conversation with very low incorrect value keeping old conversation style. 
temp_CLT = (currentStatus.coolant + CALIBRATION_TEMPERATURE_OFFSET)*4/3+11; //CLT conversion

noisymime:
@pazi88 Are you able to confirm here at all? 

It does appear that the `CALIBRATION_TEMPERATURE_OFFSET` value in `comms_CAN.cpp` is currently being subtracted when that may not be needed. 

From what I can find online the conversion should be:
`Temp in C = .75 * byte - 48.373`
or to rework it for the byte value:
`byte = (C + 48.373)/.75` 
aka:
`temp_CLT = (currentStatus.coolant + 48)*4/3; //CLT conversion`

I don't have a BMW setup to test this on. @Zazins Are you able to test this with the above line?

Zazins:
Hi! 

This is correct calculation that Im using on other projects Calculation = (HEX * 0.75) - 48°C, but I dont know why this calculation leads to incorrect values by Speeduino. If i sent coolant temperature over Serial and then calculate temperature then calculation is correct. 
However yes I can test it.

noisymime:
> (HEX * 0.75) - 48°C, but I dont know why this calculation leads to incorrect values by Speeduino.

That calculation would be to turn whatever is sent on CAN back into degrees C. What Speeduino needs to do is the opposite of this, convert degrees C into the value that is expected on the bus. 

I believe this line should be correct:

`temp_CLT = (currentStatus.coolant + 48)*4/3; //CLT conversion`

pazi88:
> @pazi88 Are you able to confirm here at all?
> 
> It does appear that the `CALIBRATION_TEMPERATURE_OFFSET` value in `comms_CAN.cpp` is currently being subtracted when that may not be needed.
> 
> From what I can find online the conversion should be: `Temp in C = .75 * byte - 48.373` or to rework it for the byte value: `byte = (C + 48.373)/.75` aka: `temp_CLT = (currentStatus.coolant + 48)*4/3; //CLT conversion`
> 
> I don't have a BMW setup to test this on. @Zazins Are you able to test this with the above line?

[...]

PR #1223, opened 2024-06-23
Fix for #1218 and rusEFI code simplification

VitorBoss:
On issue #1218 the user reported the standard messages stopped to been broadcast when the rusEFI wideband is enabled, that boils down to the extended flag been left set on the global `outMsg` variable, this PR adds lines to make sure it is always reset to 0 were is needed.

I also took the liberty to simplify rusEFI code and do a single division instead of 2 on each ARBID on the code.

### DIFF
--- a/speeduino/comms_CAN.cpp
+++ b/speeduino/comms_CAN.cpp
@@ -59,6 +59,7 @@ void CAN_write()
 
 void sendBMWCluster()
 {
+  outMsg.flags.extended = 0; //Make sure to set this to standard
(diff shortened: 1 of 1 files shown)


noisymime:
Not sure if I'm missing something, but this appears to remove the conversion from lambda to AFR? Is that right?

VitorBoss:
The conversion still happens, but instead of 4 conversions it does only one at line 97.

The conversion is done once and them it checks the ARBID and byte boundary.

PR #1225, opened 2024-07-07
New dfco menu

155ac:
*Creates a DFCO menu for removing clutter from acceleration menu.
*Changes the Priming PW menu, WUE menu and ASE menu for better fitting in smaller screens.
image


adbancroft:
Thumbs up from me: the analog gauges currently in these dialogs don't add anything and just suck up screen real estate.

PR #1227, opened 2024-07-19
Bugfix for serial I/O hanging when the payload size is only partially sent

adbancroft:
Serial I/O will hang the system if a 2nd payload byte is never sent. This PR adds a timeout to that operation.

Plus MISRA fixes & DOxygen clean up/corrections for comms.cpp

### DIFF
--- a/misra/misra_2012_text.txt
+++ b/misra/misra_2012_text.txt
@@ -273,6 +273,8 @@ Rule 21.11
 No text specified
 Rule 21.12
 No text specified
+Rule 21.15
+The pointer arguments to the Standard Library functions memcpy, memmove and memcmp shall be pointers to qualified or unqualified versions of compatible types
 Rule 22.1
 No text specified
 Rule 22.2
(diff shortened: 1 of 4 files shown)


noisymime:
Do you think it's worth pulling the value of `SERIAL_TIMEOUT` down to somewhere around 500-700 as part of this? I've been testing lower values and 700 seems to be fine for all comm functions and should be more than enough for any single transfer from TunerStudio.

adbancroft:
I’d have to defer to you on this - but lowering a timeout would improve responsiveness, which would make for a better user experience.
Sent from my iPhone

On Aug 9, 2024, at 3:10 AM, Josh Stewart ***@***.***> wrote:

﻿

Do you think it's worth pulling the value of SERIAL_TIMEOUT down to somewhere around 500-700 as part of this? I've been testing lower values and 700 seems to be fine for all comm functions and should be more than enough for any single transfer from TunerStudio.

noisymime:
> I’d have to defer to you on this - but lowering a timeout would improve responsiveness, which would make for a better user experience.

For most users it's never going to be something they run into as timeouts should be fairly rare, but playing around with it I can't seem to see any issues with dropping it either. Probably mostly impacts people using 3rd party interfaces as TunerStudio itself is pretty good at handling these things. 

I'll make that a separate commit anyway to isolate it from this bugfix

PR #1228, opened 2024-07-20
fix: Inconsistent type on revolutionTime between extern header declaration and code definition

FL0WL0W:


noisymime:
Happy to pull this in for consistency, but keep in mind that the Arduino standard is for a `long` to be a 4 byte value, so `unsigned long` should already be an alias of `uint32_t`. 

If you're working on another architecture, you might want to check that is defined as it could potentially cause issues elsewhere if a long isn't 4 bytes

PR #1229, opened 2024-07-20
fix: CORE_AVR is the only architecture that supports ATOMIC_BLOCK.

FL0WL0W:
Since CORE_AVR is the only archiecture that supports ATOMIC_BLOCK, change the preprocessor definitions to be exclusive to that architecture. rather than having to define all other architecture as exclusive from a default ATOMIC_BLOCK

### DIFF
--- a/speeduino/auxiliaries.h
+++ b/speeduino/auxiliaries.h
@@ -24,34 +24,7 @@ void wmiControl(void);
 #define SIMPLE_BOOST_I  1
 #define SIMPLE_BOOST_D  1
 
-#if(defined(CORE_TEENSY) || defined(CORE_STM32))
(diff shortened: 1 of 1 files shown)


adbancroft:
Teensy supports it as well.

How about we use `#if defined(ATOMIC_BLOCK)`?

adbancroft:
#1157 pulls in the SimplyAtomic library - does this fix your problem? Will that library work on your platform?

FL0WL0W:
It might. Put a pause on this for now. I will come back to it later

noisymime:
Closing this for now as I think it is largely resolved within #1157. Feel free to reopen or raise a new issue if this remains a problem

PR #1230, opened 2024-07-20
Only create preprocessor error when BOARD_H is not defined and not when architecture not in big long list

FL0WL0W:
Really only want to cause this error when BOARD_H is not defined. 

Reason for this change is i am pulling in speeduino and compiling from CMAKE using my own board_h. I don't want to have to edit and maintain my own globals.h when i can just define BOARD_H from the CMAKE.

### DIFF
--- a/speeduino/globals.h
+++ b/speeduino/globals.h
@@ -100,7 +100,9 @@
   #define CORE_SAM
   #define INJ_CHANNELS 8
   #define IGN_CHANNELS 8
-#else
+#endif
+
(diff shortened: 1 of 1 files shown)


adbancroft:
This feature is available in PR #1338. You can define BOARD_H as part of the compiler command line. E.g. `-DEXTERNAL_BOARD_H=\"board_native.h\" `

Can we close this PR in favor of #1338?

noisymime:
Closing as this is incorporated in #1338

PR #1231, opened 2024-07-30
fix pin definition for Blitzbox in init.cpp

chaoschris:
fixed boost pin definition for Blitzbox pin mapping


chaoschris:
not 100% sure where this error came from, but i think this was a pin mapping from a very early hardware revision and nobody has used this output until now....

chaoschris:
Thanks a lot!


Merged #1231 into master.

Issue #1232, opened 2024-07-30
Issue with stm32f103c8t6 (Bluepill 128k) compiling with TIMERS error


ArtCodeZen:
In file included from speeduino/globals.h:108,
                 from speeduino/acc_mc33810.h:5,
                 from speeduino/acc_mc33810.cpp:1:
speeduino/board_stm32_official.h: In function 'void FUEL5_TIMER_ENABLE()':
speeduino/board_stm32_official.h:269:47: error: 'TIM5' was not declared in this scope; did you mean 'TIM1'?
  269 | static inline void FUEL5_TIMER_ENABLE(void) {(TIM5)->CR1 |= TIM_CR1_CEN; (TIM5)->CR1 |= TIM_CR1_CEN; (TIM5)->SR = ~TIM_FLAG_CC1; (TIM5)->DIER |= TIM_DIER_CC1IE;}
      |                                               ^~~~
      |                                               TIM1
speeduino/board_stm32_official.h: In function 'void FUEL6_TIMER_ENABLE()':
speeduino/board_stm32_official.h:270:47: error: 'TIM5' was not declared in this scope; did you mean 'TIM1'?
  270 | static inline void FUEL6_TIMER_ENABLE(void) {(TIM5)->CR1 |= TIM_CR1_CEN; (TIM5)->CR1 |= TIM_CR1_CEN; (TIM5)->SR = ~TIM_FLAG_CC2; (TIM5)->DIER |= TIM_DIER_CC2IE;}
      |                                               ^~~~
      |                                               TIM1
speeduino/board_stm32_official.h: In function 'void FUEL7_TIMER_ENABLE()':
speeduino/board_stm32_official.h:271:47: error: 'TIM5' was not declared in this scope; did you mean 'TIM1'?
  271 | static inline void FUEL7_TIMER_ENABLE(void) {(TIM5)->CR1 |= TIM_CR1_CEN; (TIM5)->CR1 |= TIM_CR1_CEN; (TIM5)->SR = ~TIM_FLAG_CC3; (TIM5)->DIER |= TIM_DIER_CC3IE;}
      |                                               ^~~~
      |                                               TIM1
speeduino/board_stm32_official.h: In function 'void FUEL8_TIMER_ENABLE()':
speeduino/board_stm32_official.h:272:47: error: 'TIM5' was not declared in this scope; did you mean 'TIM1'?
[...]

PR #1234, opened 2024-08-02
removed unsupported/unused STM32F103 references

FDSoftware:
Considering that support for STM32F103 was removed due to the size of available flash, I removed all references to it to avoid creating more issues like:  #1232


adbancroft:
Happy to merge this if you can rebase onto latest master

PR #1238, opened 2024-08-14
feat: adds AEM 30-0300 X-Series UEGO wideband over CAN support 

wilmardo:
Currently the code is untested and based on the comment of @NickZ1969  

https://discord.com/channels/879495735912071269/937475043976441917/1229441506230796399

I am currently moving so all my car stuff is in a box somewhere and I can't test this for half a year or so. So please, if someone has a AEM wideband, feel free to test!

CAN data from the manual:
image
[source](https://static.summitracing.com/global/images/instructions/avm-30-0300.pdf)

Currently using the default 0x180 message ID, this can be changed on the gauge itself to 16 different addresses.
We can maybe say that ID1 is the primary O2 and ID2 is the secondary O2?
For now I just kept it simple with the default CAN ID :)

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -477,7 +477,7 @@ page = 1
 
       unused0_126               = bits,   U08,  126, [0:2]
       ;These are reserved for future use, in case of more CAN broadcasting features are added
-      canWBO                    = bits,   U08,  126, [2:3], "Off", "rusEFI WBO", "INVALID", "INVALID"
+      canWBO                    = bits,   U08,  126, [2:3], "Off", "rusEFI WBO", "AEM", "INVALID"
       vssAuxCh                  = bits,   U08,  126, [4:7], "Aux0", "Aux1", "Aux2", "Aux3", "Aux4", "Aux5", "Aux6", "Aux7", "Aux8", "Aux9", "Aux10", "Aux11", "Aux12", "Aux13", "Aux14", "Aux15"
       decelAmount               = scalar, U08,  127,        "%",  1.0,       0.0,   0.0,     150.0,    0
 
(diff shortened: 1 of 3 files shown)


noisymime:
Two comments around this one after a quick look:

```
if ((inLambda * configPage2.stoich / 10000) > 250) { //Check if we dont overflow the 8bit O2 variable
  currentStatus.O2 = 250;
}
currentStatus.O2 = (unsigned int)(inLambda * configPage2.stoich / 10000); 

1. I think the final line should be within an `else` clause? Otherwise the 250 will simply get overwritten 
2. `currentStatus.O2` is multiplied by 10 to give 1 point of AFR precision. Dividing by 10,000 will result in this being off by a factor of 10. I believe (If I'm reading the screenshot right) that these should be divided by 1000 instead. 

Otherwise looks fine, but really needs to be tested on an actual device if possible. Does this unit just broadcast at a given frequency or do it require a request to be sent before transmitting?

wilmardo:
> I think the final line should be within an else clause? Otherwise the 250 will simply get overwritten

Yeah thanks, good catch! MIssed it in the refactor from the rusefi case statement to an simpler if/else

> currentStatus.O2 is multiplied by 10 to give 1 point of AFR precision. Dividing by 10,000 will result in this being off by a factor of 10. I believe (If I'm reading the screenshot right) that these should be divided by 1000 instead.

Yeah not sure, I see where you are coming from but @NickZ1969 reported this working (I see you tagged him on Discord as well), otherwise this will then just have to wait for a test. Then at least the code is up to scratch only then testing remains, that was kinda my goal with this PR.

As said, I can probably test somewhere at the end of the year in the earliest.

noisymime:
> currentStatus.O2 is multiplied by 10 to give 1 point of AFR precision. Dividing by 10,000 will result in this being off by a factor of 10. I believe (If I'm reading the screenshot right) that these should be divided by 1000 instead.

OK, ignore this, the stoich config value is already 10x so the 10,000 will work.

PR #1242, opened 2024-09-02
Make O2 IAT and CLT sensor curves visible

HWright9:
This PR makes the O2, IAT and CLT tables visible and adjustable in Tuner Studio. I did this for myself as I wanted to make a custom curve, but since a lot of people have trouble with making custom .inc files etc I thought others may benifit from this.. 

This is accomplished by moving the SRAM memory of the O2 IAT and CLT calibration tables to a new "page16." structure.
The actual EEPROM memory address and values are completely unchanged. No updates are required to existing calibrations. Only the structure in the SRAM is updated so that TunerStudio can read and write these values. The existing methods of calibrating using the Tools - > Calibrate XX Sensors menu is also maintained. There are no changes to the total RAM or EEPROM used.

The calibrations are visible as tables in a new submenu. The X axis is also scaled from ADC counts into Volts to make it a bit easier when developing a custom calibration. This could easily be reveresed and expressed as straight ADC counts if others feel that is more accurate.

So if you have always wondered what the 3 point therm generator was doing, now you can see. 

Sensor cal overview
IAT sensor cal

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -244,22 +244,21 @@
    ;----------------------------------------------------------------------------
 
     endianness          = little
-    nPages              = 15
-    pageSize            = 128,   288,     288,    128,     288,    128,    240,     384,    192,    192,    288,    192,    128,    288,    256
+    nPages              = 16
+    pageSize            = 128,   288,     288,    128,     288,    128,    240,     384,    192,    192,    288,    192,    128,    288,    256,  352
 
     ; New commands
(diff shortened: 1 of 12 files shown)


HWright9:
Right now it will view and live edit the tables, but won't burn them. I think I need to update the CRC somehow for O2, IAT and CLT when page 16 burns. But not sure how to do this. If anyone wants to take this on and fix that then go for it!

PR #1252, opened 2024-10-01
Run unit tests on SimAVR

adbancroft:
This PR allows the unit tests to be run under SimAVR.

1. Anyone can run the unit test suite - no need for hardware :grin:
2. No need for remote board access in the CI build (PR includes modified GitHub action)

Some minor caveats:

- Because of [this bug](https://github.com/platformio/platformio-core/issues/4985) we need a separate environment for Ubuntu
    - `pio test -v -e megaatmega2560_sim_unittest_windows`
    - `pio test -v -e megaatmega2560_sim_unittest_ubuntu` 
        - And developer needs to install SimAVR manually: `sudo apt-get install -y simavr`
- Because of [this bug](https://github.com/buserror/simavr/issues/489), output is slow to appear on Windows (but is acceptable IMO).

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -38,12 +38,14 @@ jobs:
         platformio upgrade
         platformio pkg update
         
+    - name: Install SimAVR
+      run: |
+        sudo apt update
+        sudo apt-get install -y simavr
+        
     - name: Run Unit Tests
-      env:
-        PLATFORMIO_AUTH_TOKEN: ${{ secrets.PLATFORMIO_AUTH_TOKEN }}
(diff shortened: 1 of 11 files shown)


noisymime:
The addition of an option for using SimAVR is great and happy to bring that in, but whenever possible it's still FAR better to test on hardware IMHO. I'd strongly prefer to keep the workflow testing on hardware rather than using SimAVR, but otherwise this looks good.

adbancroft:
> I'd strongly prefer to keep the workflow testing on hardware rather than using SimAVR, but otherwise this looks good.

Sure - I reverted the CI unit test action changes

PR #1267, opened 2024-12-01
Add comprehensive unit tests for Pulse Width calculations

### LINKED ISSUE #1266: Nitrous RPM range calculation can overflow
Speeduino.ino:
`int16_t adderPercent = ((currentStatus.RPM - (configPage10.n2o_stage1_minRPM * 100)) * 100) / adderRange; //The percentage of the way through the RPM range`

Since `int` is 16-bits on AVR, if current RPM is more than 330 RPM greater than min stage RPM, the above line will overflow: 330*100=33000.

adbancroft:
Add 42 new unit tests for PW calculations, covering PW limits, secondary fueling calculation, injector staging, nitrous adjustments, calculation of required fuel. 

To support this, the PW calcs were extracted from speeduino.ino, unit tested & then broken into smaller functions and made independent of global state (easier unit testing). 

As a side effect, testing shows a 6 to 7% performance increase, a few bytes of RAM saved and 84 MISRA violations fixed.

Fixes #1266


adbancroft:
@noisymime - didn't think your were reviewing this PR. I've rebased it onto latest master.

PR #1272, opened 2024-12-16
Fuel schedule improvements

### LINKED ISSUE #1181: Dual injections on low RPM with Teensy 3.5 and 4.1
Hi,

I am opening this issue as a continuation of #1163. There I thought there was an issue with oddsquirt but the issue is much bigger than that. 

In any mode (paired, semi-sequential, sequential) there are multiple injections below a certain RPM. With Teensy3.5, the issue occurs sporadically at certain RPM`s and goes away around 450 RPM. 
I also tested it with a Dropbear-like unit with Teensy 4.1 and MC33810s which also works fine up to around 450 RPM. Then, channel 4 starts to behave strangely (dual injections) sequential mode followed by channel 3. A bit higher in the RPM, channel 2 also shows dual injections and around 1000RPM all four channels have dual injections. At 1080RPM, suddenly all injector channels start to behave as they should. This only affects the injectors, NOT the ignition.

On the mega2560 the issue isn’t there.

Attached are five pictures. Blue is the injector, yellow is ignition of cylinder 1 for reference.
1 shows channel 4 @ 543 RPM
2 shows channel 4 @ 940 RPM
3 shows channel 3 @ 940 RPM
4 shows channel 1 @ 940 RPM
5 shows channel 1 @ 1110 RPM.

Pict1
Pict2
Pict3
Pict4
Pict5

And here is the current tune:
[CurrentTune.zip](https://github.com/noisymime/speeduino/files/14271528/CurrentTune.zip)

I`m running the latest dev version. I hope someone can have a look at this. If I need to test anything with the Teensy 3.5 or 4.1 please let me know.

noisymime:
This addresses concerns raised in #1181 as well as fixing a number of additional edge case issues with fuel schedulers. 

Fixes #1181


ByMatzs:
Sorry for this return to a real situation with STM32.
For some reason that I don't understand it doesn't work correctly with misfires and an unstable engine under 1500 rpm.
Perhaps the previous version of fuel schedulers smooths out a false detection of an intermediate missing tooth.

noisymime:
> Sorry for this return to a real situation with STM32. For some reason that I don't understand it doesn't work correctly with misfires and an unstable engine under 1500 rpm. Perhaps the previous version of fuel schedulers smooths out a false detection of an intermediate missing tooth.

Are you able to give this another try now? There were a few stm32 commits that had to be reversed and may have been impacting this. No one else has reported any issues with this is all.

ByMatzs:
I did lots of tests on my simulation bench and in real conditions.
On the bench it's perfect; It seems to me that it is during rapid rev changes that the problem appears on an engine with a large camshaft. This significantly modifies the TPSdot and MAPdot enrichments. 
I will give you the feedback with other tests.

noisymime:
@ByMatzs Can you please test with the latest master code now and see if the issue is resolved?

ByMatzs:
I did the tests with the latest code.
Reassigning the STM32 timers to resolve the problem.
Many thanks for this improvement; This fixes the problems of the old code, particularly on the 6 full sequential cylinders and improves the overall operation on all configurations.

Commit 23cc14e

PR #1275, opened 2025-01-08
Scheduler RAM saving (step 6 of ?) - De-duplicate schedule structs

adbancroft:
This is the 6th PR in a series of scheduler focused PRs that end up saving >600 bytes RAM (compared to master as of 2023-06-11).

This PR:
- Removes duplication from the `IgnitionSchedule` and `FuelSchedule` structs by deriving them from a common base struct, `Schedule`.
- Optimizes the memory and code footprint of `Schedule` (use smaller types, remove unused struct members etc.)
- Adds 11 new unit tests
- Re-enables 64 existing unit tests

See #1111 for more the context on this PR.


TBAMax:
I looked into this briefly. The constexpr constructor for Schedule looks very nice.

TBAMax:
Forward from this PR:
About the ScheduleStatus, using the powers of 2 this way probably do not give any benefit. As the compiler have no way of knowing that we only want to check for one bit.
As discovered in #1280, using bit fields gives quite lot of benefit.
Maybe go Union with bitfields here too? Bits within byte, in the union with the whole byte. 
This should make possible to:
Read from the individual bits (when checking status).
Write the whole byte (when status changes).
The drawbacks would be increased code complexity and reduced readability. Because of writing to one variable and reading from another. Maybe it is possible to make it somewhat readable by choosing the names of the bits reflecting the STATUS enum names.

adbancroft:
> Forward from this PR: About the ScheduleStatus, using the powers of 2 this way probably do not give any benefit. 

It definitely does. See [here](https://github.com/adbancroft/speeduino/blob/044ac43080c7b48a448182fdb9423a34b03055b1/speeduino/scheduler.h#L191). The straightforward solution of using a logical OR (||) to check for `RUNNING || RUNNING_WITHNEXT` resulted in -5% performance hit on Mega.

adbancroft:
> Maybe go Union with bitfields here too? Bits within byte, in the union with the whole byte.

I didn't try a bit field - not sure it's as applicable here since the schedule can only be in one state at a time? Unlike #1280, where multiple flags can be on or off at the same time.

TBAMax:
Maybe a stupid question, but why use static member function `bool isRunning(const Schedule &schedule)`
instead of just member function `bool isRunning()` ?
Flash savings?

adbancroft:
> Maybe a stupid question

Not at all!

> why use static member function `bool isRunning(const Schedule &schedule)` instead of just member function `bool isRunning()` ? Flash savings?

Ever since @noisymime made [this comment](https://github.com/speeduino/speeduino/pull/520#issuecomment-778859196), I've stuck more to "C" style "struct + free function" code style rather than C++ (I'd much rather use the full expressiveness of C++, but whatever...)

TBAMax:
> Ever since @noisymime made https://github.com/speeduino/speeduino/pull/520#issuecomment-778859196, I've stuck more to "C" style "struct + free function" code style rather than C++ (I'd much rather use the full expressiveness of C++, but whatever...)

Ok, that is understandable. It has just been the usual style indeed. So no worries at the moment. We can always convert to member functions later. As of my understanding reaches, anything that builds on Platformio and/or Arduino IDE goes.

After spending 5 hours today on topic of schedule structs this I would say:
The main points are all good:
> - Removes duplication from the IgnitionSchedule and FuelSchedule structs by deriving them from a common base struct, Schedule.
> - Optimizes the memory and code footprint of Schedule (use smaller types, remove unused struct members etc.)
and SheduleStatus powers of 2

However the macro [here](https://github.com/adbancroft/speeduino/blob/044ac43080c7b48a448182fdb9423a34b03055b1/speeduino/speeduino.ino#L147C1-L147C41 ) seems complete mess. The `channelIdx` is not very readable and given the time constraints I was unable to determine the potential gains from this, and what this does.

PR #1280, opened 2025-01-15
Statuses struct: use bit fields for status bytes

adbancroft:
Make the compiler do the heavy lifting: use a combination of type punning and anonymous structs to replace manual bit twiddling of `statuses` status byte fields.

This makes for much simpler code, as well as setting us up for optimization of the struct in a future PR.


TBAMax:
From where the flash savings on atmega are coming from in this case?

All efforts to introduce bitfields in the past have failed with the argument of possibly slower code on atmega and with the clause for the applicant to prove otherwise. And prove hard -- I mean full blown thesis of masters degree grade work...

adbancroft:
> From where the flash savings on atmega are coming from in this case?

Since we've told the compiler these are 1-bit bools, it can do a whole bunch of optimisations when there are multiple set/clear/check operations within a function. When it's an opaque byte, it is forced to emit assembler for each macro invocation.

> All efforts to introduce bitfields in the past have failed with the argument of possibly slower code on atmega and with the clause for the applicant to prove otherwise. 

For individual set/clear/check, the compiler produces pretty much identical assembly. 

See here for examples showing both situations: (https://godbolt.org/z/a5dhjP4Y1)[https://godbolt.org/z/a5dhjP4Y1]

PR #1289, opened 2025-03-20
Fix Stepper Motor Issue in idle.cpp by Reverting Hysteresis Check Logic

bahram-azimbeyk:
This pull request addresses an issue with the stepper motor control in the latest firmware (202501), where the stepper motor skips steps or behaves erratically. The problem was traced to the hysteresis check condition in `idle.cpp` (line 337), introduced in the recent update.

**Issue:**
The current condition:
```cpp
if ((idleStepper.curIdleStep >= (idleStepper.targetIdleStep - configPage6.iacStepHyster)) && 
    (idleStepper.curIdleStep <= (idleStepper.targetIdleStep + configPage6.iacStepHyster)))
```
disables the stepper motor (pinStepperEnable set to HIGH) too early, before it fully reaches the target position. This causes the stepper to skip steps or become unstable, especially noticeable on certain vehicles.

Fix:
Reverting to the previous hysteresis check logic:

```cpp
if ((idleStepper.targetIdleStep > (idleStepper.curIdleStep - configPage6.iacStepHyster)) && 
    (idleStepper.targetIdleStep < (idleStepper.curIdleStep + configPage6.iacStepHyster)))
```
resolves the issue. This condition ensures the stepper only stops when the target position is sufficiently close to the current position, preventing premature disabling and stabilizing motor behavior.

Testing:
I tested this change on my vehicle, and it completely fixed the stepper motor issue—no skipping or erratic behavior observed. The old logic appears better aligned with the stepper control loop and hardware timing.

Why This Change?
While the new condition might have been intended for readability or optimization, it seems to conflict with the stepper control timing or hardware specifics in some setups. Reverting to the original logic is a safer fix until the root cause of the new condition's failure is fully investigated.

Please review and let me know if further testing or adjustments are needed


noisymime:
What do you have for your step hysteresis and cool time values?

bahram-azimbeyk:
image

adbancroft:
Could you move out the check to a separate function? This can't be the only place where we are doing approximate comparisons. E.g 
```
template <typename T>
bool isWithin(const T& compare, const T& lower, const T& upper)
```
Perhaps in `math.h`?

And then add unit tests for the new function..

bahram-azimbeyk:
@adbancroft I'll update it

Zeiberstein:
I think the reshuffling of the `if` condition is distracting.

Let:

* `cur = idleStepper.curIdleStep`
* `target = idleStepper.targetIdleStep`
* `hys = configPage6.iacStepHyster`

Then the current form

```cpp
(cur >= target - hys) && (cur <= target + hys)
```

can be rewritten directly as:

* `cur >= target - hys` is equivalent to `target <= cur + hys`
* `cur <= target + hys` is equivalent to `target >= cur - hys`

so together that becomes:

```cpp
(target >= cur - hys) && (target <= cur + hys)
```

**So the reshuffling itself does not change the logic.** It is still the same inclusive range check, just written in a different way.
<BR><BR>
Bahram-azimbeyk's PR description suggests that the intended behavioral change is to use **exclusive** bounds instead:

```cpp
(target > cur - hys) && (target < cur + hys)
```

However, adbancroft then suggested introducing a generic helper:

```cpp
static inline bool isWithin(const T& compare, const T& lower, const T& upper) {
  return (compare >= lower) && (compare <= upper);
}
```

**This helper uses inclusive bounds.**

So the important behavioral difference is not whether the condition is written around `cur` or around `target`, but only whether the bounds are **inclusive** (`>=` / `<=`) or **exclusive** (`>` / `<`).

**As the PR currently stands, it does not appear to introduce a functional change to the hysteresis check.** The current force-pushed version uses `isWithin(target, cur - hys, cur + hys)`, and since `isWithin()` is inclusive, that is functionally equivalent to the current inclusive comparison already in place.
<BR><BR>
If the goal is to make this more generic and still keep the behavioral intent explicit, it might be clearer to provide two helpers, for example `isWithinInclusive()` and `isWithinExclusive()`, so the intended behavior remains explicit in the call site.

PR #1292, opened 2025-04-02
fix stepper driver overheating - unaccessible pinStepperEnable set if cool time is set to 0

okenagod:
If user set iacCoolTime to 0, program run will newer get to 'stepper motor enable pin setting' part. 
Explanation:
Stepper status have to be STEPPING or COOLING to get into condition on line 300, then there is next condition for STEPPING status and power disable check is in else statement, so only status COOLING can get there. But if iacCoolTime=0, status is set to SOFF and stepper motor remains enabled.


noisymime:
Tested and confirmed. Thanks for the PR!

PR #1295, opened 2025-04-18
Add baud option for primary canbus on Teensy 4.1

rushtonmd:


Autohome2:
I would suggest a warning message (either always present or being made visible) in TS when setting to anything other than 500k warning users that for OBD2 it needs to be set at 500k for most(all) code readers to be able to connect

RempageR1:
Small typo I think in the default speed:

`        default: // Fallback to 250k if value is invalid
          can_speed = 500000;
          break;`

pazi88:
This is much needed PR. But has couple of problems:

- As @Autohome2 mentioned, this needs a warning that the OBD2 (and BMW/VAG broadcasts) wont work with other than 500k baud rate.
- Also the PR is missing the most important baud rate which is the 1M (1000k) needed by the Haltech protocol. Currently the CAN bus is fixed to 500k and because the Haltech protocol specs out 1M speed, it wont work on devices where you can't change the baud rate.
- And the PR description is wrong. This affects other CAN capable processors too, like STM32, Teensy 3.5, 3.6 etc. Not just Teensy 4.1. Which is good, but just false description.

PR #1296, opened 2025-04-26
Refactor dfco for better readability and minor performance improvement

GatisPastars:
Decided to open a can of worms here :)

Refactored two following methods. MISRA is not violated since single return is only recommendation and these methods do not allocate resources that needs to be released before return.

1. DFCO fuel correction calculation method for cleaner code: 
a) gave it a name with verb in it - getDfcoFuelCorrection
b) moved early exit conditions to top
c) replaced comparing condition with min()
d) replaced deep nesting and else statements with simple one level conditions

2. Refactored DFCO Activation and status reading method for cleaner code:
a) gave it a name with verb in it - isDfcoOngoing
b) moved early exit conditions to top
c) got rid of condition that checked if dfco is off when it was on to reset dfcoDelay. Now dfcoDelay is reset when state changes. Should give minor performance improvement while dfco is running.
d) swapped dfco conditions so that the fastest one would be executed first
e) replaced deep nesting and else statements with simple one level conditions
f) added more tests and gave existing ones longer names

Code was compiled, built and unit tested for Teensy 4.1

### DIFF
--- a/speeduino/corrections.cpp
+++ b/speeduino/corrections.cpp
@@ -138,7 +138,7 @@ uint16_t correctionsFuel(void)
   currentStatus.launchCorrection = correctionLaunch();
   if (currentStatus.launchCorrection != 100) { sumCorrections = div100(sumCorrections * currentStatus.launchCorrection); }
 
-  bitWrite(currentStatus.status1, BIT_STATUS1_DFCO, correctionDFCO());
+  bitWrite(currentStatus.status1, BIT_STATUS1_DFCO, isDfcoOngoing());
   byte dfcoTaperCorrection = getDfcoFuelCorrectionPercentage();
   if (dfcoTaperCorrection == 0) { sumCorrections = 0; }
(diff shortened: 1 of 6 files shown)


speeduino/corrections.cpp
```diff
@@ -517,53 +520,69 @@ byte correctionLaunch(void)
...
+  }
+  return map(dfcoTaper, configPage9.dfcoTaperTime, 0, 100, configPage9.dfcoTaperFuel);
 }
 
 /*
  * Returns true if deceleration fuel cutoff should be on, false if its off
+ * Mutates dfcoDelay
  */
-bool correctionDFCO(void)
+bool isDfcoOngoing(void)
```
GatisPastars, line 553: @noisymime Would You prefer to isolate these functions from configs and statuses for clarity and easier testing? Just like calculateAfrTarget does for example? As I understand - cost for passing few pointers as parameters is insignificant.

PR #1297, opened 2025-04-28
Ford TFI Decoder

PascalGodin:
Decoder for Ford TFI Distributor mounted wheel with narrower signature tooth for cylinder no 1


noisymime:
Thanks for the PR. Has this been tested on a car yet?

PascalGodin:
Yes, tested on a 8 cylinder (my truck). 4 and 6 cylinders are not tested but assume the same width ratio of regular vs signature teeth for all cylinder count.

PR #1306, opened 2025-06-08
Optimise arduino library management in speeduino repo

adbancroft:
Both Platform IO and Arduino support the library concept as a means of sharing code across projects. This PR replaces 4 chunks of code with existing libraries:

1. src/libdivide - replaced by libdivide library
2. Optimised division (udiv_32_16) - replaced by avr-fast-div library
3. Optimised bit shifts (lshift, rshift) - replaced by avr-fast-shift library
4. src/fastcrc - replaced by FastCRC library

(note: all library versions are pinned)

This:
- improves maintainability: conceptually less code to grok, unit test, scan etc. E.g. no need to rename libdivide folder before & after MISRA scan.
- Works with our tooling as intended
- Supports the OSS community by dogfooding our own code (avr-fast-div, avr-fast-shift) & reusing others (libdivide, FastCRC).

### DIFF
--- a/speeduino/decoders.cpp
+++ b/speeduino/decoders.cpp
@@ -6065,8 +6065,8 @@ void triggerSec_FordTFI(void)
   {     
     if ((curGap > 0) && (curGap < 20000000)) // Limit to prevent overflow
     {  
-      targetGap2 = curGap * 110UL / 100UL; // Wide last teeth gap min
-      targetGap3 = curGap * 90UL / 100UL; // Narrow last teeth minus one gap max
+      targetGap2 = percentage(110, curGap); // Wide last teeth gap min
+      targetGap3 = percentage(90, curGap); // Narrow last teeth minus one gap max
     }
     else 
     {
--- a/speeduino/fuel_calcs.cpp
+++ b/speeduino/fuel_calcs.cpp
@@ -73,7 +73,7 @@ TESTABLE_INLINE_STATIC uint16_t calculatePWLimit(const config2 &page2, const sta
       break;
     default:
       //Non-PoT squirts value. Perform (slow) uint32_t division
-      tempLimit = tempLimit / current.nSquirts;
+      tempLimit = fast_div (tempLimit, current.nSquirts);
       break;
   }
   return (uint16_t)min(tempLimit, (uint32_t)UINT16_MAX);
@@ -82,11 +82,11 @@ TESTABLE_INLINE_STATIC uint16_t calculatePWLimit(const config2 &page2, const sta
 
 static inline uint32_t applyMapMode(uint32_t intermediate, const config2 &page2, const statuses &current) {
   if ( page2.multiplyMAP == MULTIPLY_MAP_MODE_100) { 
-    uint16_t multiplier = div100((uint16_t)((uint16_t)current.MAP << 7U));
+    uint16_t multiplier = div100(lshift<7U>((uint32_t)current.MAP));
     return rshift<7U>(intermediate * (uint32_t)multiplier); 
   }
   if( page2.multiplyMAP == MULTIPLY_MAP_MODE_BARO) { 
-     uint16_t multiplier = ((uint16_t)current.MAP << 7U) / current.baro; 
+     uint16_t multiplier = fast_div32_16(lshift<7U>((uint32_t)current.MAP), current.baro); 
     return rshift<7U>(intermediate * (uint32_t)multiplier); 
   }
   return intermediate;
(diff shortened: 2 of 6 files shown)


ric355:
Is this what is actually desired here? I have a feeling the sources were originally included so that the versions of those libraries were effectively fixed, and therefore anyone building from source would be building exactly the same thing as everyone else.

noisymime:
There wer 2 main reasons for bundling the libraries within the main source:

1. It allowed versions to be locked or customised as needed. That can be achieved through having versions of the libraries hosted on github under the speeduino organisation though, so this can be worked around, but it defeats the benefit of removing them from the main source though. 
2. Probably more importantly, it allows the code to be compiled out of the box on the Arduino IDE. No additional steps are needed, just open the project and compile.

The second point is probably the one that concerns me the most here. Adding extra barriers to people compiling the code on the Arduino IDE is something I'd prefer to avoid if possible.

adbancroft:
> Is this what is actually desired here? I have a feeling the sources were originally included so that the versions of those libraries were effectively fixed, and therefore anyone building from source would be building exactly the same thing as everyone else.
>
>It allowed versions to be locked or customised as needed. That can be achieved through having versions of the libraries hosted on github under the speeduino organisation though, so this can be worked around, but it defeats the benefit of removing them from the main source though.

I've updated the configurations to pull specific versions of the libraries, which achieves the same thing ("versions of those libraries were effectively fixed", "versions to be locked") in a much cleaner manner.

For library customisation: I'd argue this is very rare, but that should be done as a PR on the original library so it can benefit the wider OSS community. **This is exactly what I did with libdivide, by adding 16-bit support, before including it into speeduino in PR #1082**

adbancroft:
>Probably more importantly, it allows the code to be compiled out of the box on the Arduino IDE. No additional steps are needed, just open the project and compile.

That's not the case in `master` today! 
1. You need to install the SimplyAtomic library (and EEPROM & SPI?)
2. You *should* set various compiler flags. E.g. USE_LIBDIVIDE, lto etc.

adbancroft:
Huge differences in the binary generated by Platform IO:
```
RAM:   [========= ]  91.2% (used 7468 bytes from 8192 bytes)
Flash: [========= ]  **89.2%** (used 226426 bytes from 253952 bytes)
```

vs Arduino IDE:
```
Sketch uses 171076 bytes (**67%**) of program storage space. Maximum is 253952 bytes.
Global variables use 7562 bytes (92%) of dynamic memory, leaving 630 bytes for local variables. Maximum is 8192 bytes.
```

adbancroft:
> The second point is probably the one that concerns me the most here. Adding extra barriers to people compiling the code on the Arduino IDE is something I'd prefer to avoid if possible.

To what end? What tooling is your primary development community using?

I'd argue the project has out-grown the Arduino IDE (that's a positive thing!) and we should move on. Probably a whole other conversation though!

adbancroft:
To add to my point: when we include source instead of a library, we aren't including unit tests. In some cases they can't be included (ever) due to tooling incompatibilities. 

Several of the libraries included in this PR have large & comprehensive unit tests.

Huge weakness :-(

coolhandle01:
> There were 2 main reasons for bundling the libraries within the main source:
> 
> 1. It allowed versions to be locked or customised as needed. That can be achieved through having versions of the libraries hosted on github under the speeduino organisation though, so this can be worked around, but it defeats the benefit of removing them from the main source though.
> 2. Probably more importantly, it allows the code to be compiled out of the box on the Arduino IDE. No additional steps are needed, just open the project and compile.
> 
> The second point is probably the one that concerns me the most here. Adding extra barriers to people compiling the code on the Arduino IDE is something I'd prefer to avoid if possible.

My 2c, because I was wondering if this PR existed:

Maybe this PRs title "using libraries instead of source code" is a bit too colloquial. From the capital L, I infer that on balance this probably means "optimise arduino library management in speeduino repo", and I don't think the intent is to add friction at all. It looks like @adbancroft is effectively saying that there are proper ways to tweak a third party library (fork it! Then push your fork to github, and use your fork in your platform io config (that is, add your fork of the library to your arduino libraries) to build the tweaks into your firmware). The can of worms here is that it's not really Speeduino's job to grease those wheels: platform io, the arduino library format, and github were build for these purposes.

.. I think if @adbancroft had resolved to edit the original PR description with the binary size differences, then his case may have ended up being a bit more persuasive. I do agree though.

> the project has out-grown the Arduino IDE (that's a positive thing!) 

I would say so. Honestly since I figured out how to compile-for and debug a nano with VS Code, I haven't looked back. It's a real IDE! I'm hoping I can have the same experience with speeduino

adbancroft:
> It looks like @adbancroft is effectively saying that there are proper ways to tweak a third party library 

Yes!

> (fork it! Then push your fork to github, and use your fork in your platform io config (that is, add your fork of the library to your arduino libraries) to build the tweaks into your firmware). 

You could do that, but I was really suggesting the change should be incorporated using a PR **against the original library**. Once that PR is accepted, a new version of the library would be published. Look at [libdivide](https://github.com/ridiculousfish/libdivide) for example - I've authored multiple pull requests to support Arduino.

> The can of worms here is that it's not really Speeduino's job to grease those wheels: platform io, the arduino library format, and github were build for these purposes.

It sounds like this is in support of the PR, otherwise I'm not sure what you meant here? 

> @adbancroft had resolved to edit the original PR description

Done

coolhandle01:
I am in support of this PR, aye :)

And in terms of not just forking but requesting prs upstream, yes! I was basically talking about a compromise, talking about forks and platform io but you're right - if the things you guys fix in speeduino can a) come back via mainline/stable releases after landing upstream PRs and thus b) benefit a wider audience overall, then that's good for everyone surely for all the reasons outlined

PR #1309, opened 2025-06-18
Use 8-bit 3D table axes (save 390+ bytes RAM)

adbancroft:
### Summary
Save 390+ byte of RAM by using `uint8_t` for 3D table axes instead of `int16_t`.

### Explanation
Tunerstudio sends the 3D table axes as 8 bit values which we re-hydrated into 16-bits in SRAM and de-hydrated back to 8 bits for reading/writing to EEPROM. I.e. byte<-->word<-->byte.

This PR just cuts out the middle step. It also removes the concept of an "axis domain" (no longer needed), which greatly simplifies the 3d table code/macros.

No loss in loops/sec.

### DIFF
--- a/speeduino/corrections.cpp
+++ b/speeduino/corrections.cpp
@@ -357,7 +357,7 @@ uint16_t correctionAccel(void)
           else
           {
             BIT_SET(currentStatus.engine, BIT_ENGINE_ACC); //Mark acceleration enrichment as active.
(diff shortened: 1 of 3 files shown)


adbancroft:
Closing this - as implemented there is a loss of fidelity: when interpolating the axis bin position, we're almost certain to end up on a bin edge.

PR #1310, opened 2025-06-19
[Feature] Add RusEFI/Fome CAN Broadcast protocol

JakubPrzystasz:


noisymime:
Has this one been tested and confirmed on any hardware?

In its current state it doesn't compile as there's a missing definition for `CAN_RUSEFI_EGT`, which just make me wonder how much testing this has had.

PR #1326, opened 2025-10-14
Add SRAM, SPI & FRAM based EEPROM STM32 environments to CI build

adbancroft:
IMO, the CI build is missing some build environments:
* STM + SRAM based EEPROM
* STM + SPI based EEPROM
* STM + FRAM based EEPROM


adbancroft:
Alternatively, if these aren't required they should be removed from the code base

PR #1327, opened 2025-10-14
Remove volatile from PIN_MASK variable declarations

adbancroft:
The volatile modifier isn't needed as they are never written to after initialization.


TBAMax:
Easy way to save some bytes of flash again. Nice!

PR #1336, opened 2025-11-04
Free Timer2 by switching to millis()-based timing

TBAMax:
This frees the Timer2 resource on Atmega. And corresponding timers on other chips.
Replaces the old #807, but without the tacho implementation. Only frees timer2 and replaces the timing with millis() timer.

### DIFF
--- a/speeduino/board_avr2560.cpp
+++ b/speeduino/board_avr2560.cpp
@@ -46,19 +46,6 @@ void initBoard(uint32_t baudRate)
       idle_pwm_max_count = (uint16_t)(MICROS_PER_SEC / (16U * configPage6.idleFreq * 2U)); //Converts the frequency in Hz to the number of ticks (at 16uS) it takes to complete 1 cycle. Note that the frequency is divided by 2 coming from TS to allow for up to 512hz
     }
 
-    /*
-    ***********************************************************************************************************
-    * Timers
-    */
-    //Configure Timer2 for our low-freq interrupt code.
-    TCCR2B = TIMER_PRESCALER_OFF;   //Disable Timer2 while we set it up
-    TCNT2  = 131;                   //Preload timer2 with 131 cycles, leaving 125 till overflow. As the timer runs at 125Khz, this causes overflow to occur at 1Khz = 1ms
-    TIMSK2 = (1<<TOIE2);            //Timer2 Set Overflow Interrupt enabled.
-    TCCR2A = TIMER_MODE_NORMAL;     //Timer2 Control Reg A: Wave Gen Mode normal
-    /* Now configure the prescaler to CPU clock divided by 128 = 125Khz */
-    TCCR2B = (1<<CS22)  | (1<<CS20); // Set bits. This timer uses different prescaler values, thus we cannot use the defines above.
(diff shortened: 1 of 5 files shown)


noisymime:
Unfortunately I don' think the accuracy of this is going to be even close to good enough. This method relies upon the mainloop running at at least 1000 loops/s, which is not always going to be the case for mega 2560 units. 

Things like the tacho output and particularly the flex fuel calcs rely on this being run as close to 1000ms as possible, which isn't going to be the case here.

TBAMax:
> Unfortunately I don' think the accuracy of this is going to be even close to good enough. This method relies upon the mainloop running at at least 1000 loops/s, which is not always going to be the case for mega 2560 units.
> 
> Things like the tacho output and particularly the flex fuel calcs rely on this being run as close to 1000ms as possible, which isn't going to be the case here.

This is correct. 

- The flex sensor reading should also be reworked to use only the pin change interrupts instead of polling with timer.
- The tacho is a different issue. When Timer2 is free (after the proposed change) it could be used for tacho only with better dedicated implementation. Or some other implementation for tacho is nessesary, possibly even digital only, because current implementation seems poor anway, and there are no proof that anybody is using it actually at all at the moment... Because I personally would like to see Timer2 used in better decoder filtering.

noisymime:
> * The flex sensor reading should also be reworked to use only the pin change interrupts instead of polling with timer.

The flex sensor does use the pin change interrupt to count the pulses, it then uses the periodic low-res timer to reset that count every 1s and store the value. 

> Or some other implementation for tacho is nessesary, possibly even digital only, because current implementation seems poor anway, and there are no proof that anybody is using it actually at all at the moment

Not sure where you're getting that idea, there are hundreds of setups using the existing tacho setup. It covers the vast majority of use cases, the only thing I'm aware that it can't do easily is freely drive a tacho that was designed for a different number of cylinders.

TBAMax:
> Not sure where you're getting that idea, there are hundreds of setups using the existing tacho setup. It covers the vast majority of use cases, the only thing I'm aware that it can't do easily is freely drive a tacho that was designed for a different number of cylinders.

Sorry, I was probably being too arrogant about that. I somehow was under the impression, that most installations would use some kind of separate tacho controller or digital dash.
 Anyway it seems impossible to solve the tacho without some kind of timer interrupts, when the loop speeds are below 1000loops/s (unless tacho that can work with the dwell fixed to the coil dwell). I understand the tacho dwell needs to be different from the coil dwell however... So it is currently unresolved.

-  when leaving tacho with the current implementation on timer2 interrupt, it will delete the main purpose of the current PR - freeing Timer2. So I am not gonna suggest it.

>  The flex sensor does use the pin change interrupt to count the pulses, it then uses the periodic low-res timer to reset that count every 1s and store the value.

Seems that the flex sensor can be easily converted to millis() then. Just would need to check the millis() at every pulse, and when second is passed save the pulse count. If I am getting it correctly.

- I am not going to add this change into this PR however, to keep things simpler. Also I am not flex fuel expert.
- The mentioned change can be done separately before this PR, then this PR rebased.

PR #1338, opened 2025-11-09
Separate board headers from globals

adbancroft:
This PR separates inclusion of the board headers from `globals.h`. This:
1. Improves cohesion, unit testability and granularity (no need to include `globals.h` just for a board item such as `INJ_CHANNELS`)
2. Fixes a circular header dependency (globals.h -> board_avr2560.h -> globals.h)
See `board_definition.h`

The changes in this PR are 99% moving code around. The only material changes are to Teensy board initialization (see `initBoard`)

This supports future PRs that will:
1. Generate unit test code coverage
2. Move all board specific code into the various `board*.cpp` files (which improves cohesion).

### DIFF
--- a/speeduino/board_avr2560.cpp
+++ b/speeduino/board_avr2560.cpp
@@ -130,4 +130,9 @@ void boardInitRTC(void)
   // Do nothing
 }
 
+void boardInitPins(void)
+{
+  // Do nothing
+}
+
 #endif //CORE_AVR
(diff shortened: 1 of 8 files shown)


adbancroft:
These same code changes were required in multiple other PRs I'm working on, so seems sensible to submit as a separate, small, PR

noisymime:
Looks fine to me.

PR #1339, opened 2025-11-09
Fix "atomic" CI build issue 

### LINKED ISSUE #1320: "Calculate memory deltas" action is broken for STM32
It's failing for [master & all PRs](https://github.com/speeduino/speeduino/actions/workflows/pr-memory-deltas-generate.yml):
```
  /home/runner/work/speeduino/speeduino/speeduino/statuses.h:17:12: fatal error: util/atomic.h: No such file or directory
     17 |   #include <util/atomic.h>
        |            ^~~~~~~~~~~~~~~
```

adbancroft:
Fixes #1320

### DIFF
--- a/.github/workflows/pr-memory-deltas-generate.yml
+++ b/.github/workflows/pr-memory-deltas-generate.yml
@@ -36,7 +36,7 @@ jobs:
                 name: teensy:avr
             compile-flags: |
               - --build-property
-              - build..extra_flags="-Wall"
+              - build.extra_flags="-Wall"
             lib_deps: |
               - name: SDfat
               - name: SimplyAtomic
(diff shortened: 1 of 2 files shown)


noisymime:
Worth keeping in mind that with this change, the ArduinoIDE will always require the addition of the Simply Atomic library. There's no longer any fallback to the AVR built-in ATOMIC().

adbancroft:
> Worth keeping in mind that with this change, the ArduinoIDE will always require the addition of the Simply Atomic library. There's no longer any fallback to the AVR built-in ATOMIC().

I know - but IMO that is outweighed by the broken CI (MISRA & the STM32 memory compare). E.g. as part of this PR I had to fix an issue that was hidden by the failing CI.

PR #1344, opened 2025-11-22
Rename speeduino.ino files to cpp

smclt30p:
It's time to switch to the cpp file format, i think :) 

Tested on the bench with platformIO...compiles, flashes, communicates...

Reasoning is tooling; most IDEs don't understand ino files and switching to cpp provides proper support (completio, go to definition etc.)


noisymime:
Alas we can't, or it would've been done a while back. This completely breaks compiling on the Arduino IDE, which requires that the main project file have a .ino extension.

smclt30p:
Ah makes sense, for people flashing the firmware with the Arduino IDE. I thought VSCode/PlatformIO was the main tooling.

noisymime:
>  I thought VSCode/PlatformIO was the main tooling.

It is and it's what all the main devs are using (To the best of my knowledge), but a large number of users still make their own customisations and then compile with the Arduino IDE, so I'm committed to retaining support for it.

noisymime:
FWIW there are some signs recently of some nice updates coming to the Arduino IDE that *may* allow for this, but they're not there currently.

adbancroft:
> > I thought VSCode/PlatformIO was the main tooling.
> 
> It is and it's what all the main devs are using (To the best of my knowledge), but a large number of users still make their own customisations and then compile with the Arduino IDE, so I'm committed to retaining support for it.

IMO, the project has out grown the Arduino IDE. It's OK for a single file sketch, but we are way beyond that.
1. Can't use sub-folders (other than `src` ) for organizing the code base.
2. Can't set compiler flags.

Do we know what ratio of the developer user base use it?

Issue #1345, opened 2025-11-23
Starting the tooth or composite logger will interfere with the VMax decoder


adbancroft:
The VMax decoder sets the primary trigger edge to true or false, but then attaches the primary interrupt to a `CHANGE` trigger:
```
      if(configPage4.TrigEdge == 0) { primaryTriggerEdge = true; } // set as boolean so we can directly use it in decoder.
      else { primaryTriggerEdge = false; }
      
      attachInterrupt(triggerInterrupt, triggerHandler, CHANGE); //Hardcoded change, the primaryTriggerEdge will be used in the decoder to select if it`s an inverted or non-inverted signal.
```
Since these don't match the logger will not call the primary trigger appropriately and when the logger is stopped, the primary trigger will not be reattached to the correct interrupt trigger.

PR #1356, opened 2025-12-04
Refactor and unit test decoder initialization

adbancroft:
This PR separates out decoder initialization (from `init.cpp`) and unit tests it (see `test_decoder_init.cpp`).

To support this, it also:
1. Encapsulates individual decoder setup into the existing `triggerSetup_` functions.
2. Limits the public surface area of the decoders. I.e. `decoders.h` only declares what needs to be accessed by other code (limit coupling, increase cohesion)
3. Unit tests attaching & detaching tooth logger for all decoders (as part of fixing #1345)
4. Encapsulate trigger ports & pin masks into `decoders.cpp` (which is the only place they are used)

**All of this is covered by new unit tests in `test_decoder_init.cpp`, `test_decoderbuilder.cpp` & `test_toothloggers.cpp`**


adbancroft:
Fixes #1345

PR #1369, opened 2026-01-01
Use an enum for decoder sync status

adbancroft:
Decoder sync state can be one of **3** values: None, Partial, Full. However, it's stored in 2 [^1] `bool` variables in the `statuses` struct, which have **4** possible states:

`hasSync` | `halfSync` | SyncState
-----------|------------|------
F|F|None
T|F|Full
F|T|Partial
T|T|Invalid?

This PR:
1. Replaces the 2 bools with an enum (see `SyncStatus` et. al.)
  So no 4th, invalid, state 
3. The sync state is stored as part of the decoder state
  It's natural home

[^1]: It's actually 3 `bool` vars, but `hasFullSync` shadows `hasSync`

### DIFF
--- a/speeduino/decoders.cpp
+++ b/speeduino/decoders.cpp
@@ -185,7 +185,7 @@ static inline void addToothLogEntry(unsigned long toothTime, byte whichTooth)
         }
         if(whichTooth > TOOTH_CRANK) { BIT_SET(compositeLogHistory[toothHistoryIndex], COMPOSITE_LOG_TRIG); }
       }  
-      if(currentStatus.hasSync == true) { BIT_SET(compositeLogHistory[toothHistoryIndex], COMPOSITE_LOG_SYNC); }
+      if(getSyncStatus()==SyncStatus::Full) { BIT_SET(compositeLogHistory[toothHistoryIndex], COMPOSITE_LOG_SYNC); }
 
       if(revolutionOne == 1)
       { BIT_SET(compositeLogHistory[toothHistoryIndex], COMPOSITE_ENGINE_CYCLE);}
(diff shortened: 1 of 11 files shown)


TBAMax:
I like the idea of the updated sync states.

Also I like the use of getter function `getSyncStatus()`. This allows actually allows us to check for fresh info from decoder about the sync status, in cases when the trigger pulses stop suddenly (for example hardware breaks), just checking the sync variable would actually not be enough. When there are no more trigger pulses the sync variable is not updated, so it stays active longer than nessesary. This PR however do not update the sync checking in this function, and it still just returns the variable statuses.  I expect this to be adressed in the follow up PR, as this would be the nice addition. This PR seems to make good way for such further improvements. So I highly approve this.

I am not happy with the incorporated change of the structure of `changeHalfToFullSync` and `changeFullToHalfSync` parameters, this is difficult to test or think through together with the main theme of the PR. So I consider this a side hustle trying to sneak in with the main theme. An better would be in separate PR. But it seem to be not very big issue.

I have not actually tested all aspects. But this is just review from looking at the code changes briefly overall.

adbancroft:
> I am not happy with the incorporated change of the structure of changeHalfToFullSync and changeFullToHalfSync parameters, this is difficult to test or think through together with the main theme of the PR. So I consider this a side hustle trying to sneak in with the main theme. An better would be in separate PR. But it seem to be not very big issue.

(This should have been in the PR description)

These functions are the main (only?) consumer of `getSyncState`; the PR includes multiple new tests for them both (see `test_fuel_schedule_init.cpp`). I guess it could be a separate PR, but enhancing the unit tests was one of the things that initiated this PR. Seems harmless to leave it it?

TBAMax:
> These functions are the main (only?) consumer of getSyncState;

Good observation! The sync state is not used in very many places. Of course the "main loop" that runs with this clause is the most important consumer!

> Seems harmless to leave it it?

  changeFullToHalfSync and back is generally quite sensitive topic. I remember it was quite struggle to get it working right when it was introduced. Also there were possible race conditions when outputs were reconfigured (some would stick "ON" for longer than nessesary), so it must be approached with caution. 
  I think it is ok to leave it in. no big trouble. Haven't looked closely what this is tested against thouhg. With those functions the main concerns was probably other kind than tests can detect -- that the outputs would not stick when callback pointers are changed mid-running schedule.

adbancroft:
> changeFullToHalfSync and back is generally quite sensitive topic. I remember it was quite struggle to get it working right when it was introduced. Also there were possible race conditions when outputs were reconfigured (some would stick "ON" for longer than nessesary), so it must be approached with caution.
> I think it is ok to leave it in. no big trouble. Haven't looked closely what this is tested against thouhg. With those functions the main concerns was probably other kind than tests can detect -- that the outputs would not stick when callback pointers are changed mid-running schedule.

There were minimal changes to the half<-->full sync functions, just enough to get tests to pass. They still have several potential bugs (E.g. changing ignition callbacks without checking the schedule is off), but fixing those is a future PR.

PR #1372, opened 2026-01-02
Add the fuel menu and re-organize some stuff

smclt30p:
* Move all the corrections (cranking, iat, clt, accel, baro) under the Fuel menu
* Move injector configuration under the Fuel menu
* Move lambda and flex to Fuel menu
* Rename "Tuning" to "Tables" to display proper icon in TS
* Rename "Startup/Idle" menu to "Idle" as it no longer handles crank enrichment
* Move "Engine protection" to Settings and rename it "Limits/protection"
* Rename "AFR protection" to "Lean power cut" as lambda is not a AFR ratio

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -1992,11 +1992,15 @@ menuDialog = main
 
    menu = "Settings"
       subMenu = engine_constants,   "Engine Constants"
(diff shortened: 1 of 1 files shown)


8SecSleeper:
I did this years ago and the result was no merge, stating that the users are already use the complete mess the menu's had :(.

PR #1373, opened 2026-01-03
Replace run-time shift with compile time constant.

adbancroft:
Small performance bump on AtMega2560


TBAMax:
@adbancroft Did you snoop out that it actually did not get calculated compiletime? Because of usage of fast-div libraries?
I would expect it to get precalculated at compiletime as it was also, because it was all constants...

adbancroft:
Good question...I did not. However, the `lshift()` call that was used originally used inline assembler, so it seems like a good assumption that it was **not** pre-calculated, Even if it was, the new code makes it explicitly compile time, which can only be a good thing.

PR #1445, opened 2026-03-30
Separate comms interfaces to allow for multiple connections

noisymime:
This is a PR raised for discussion and review and is not currently in a ready to pull status.

In the current master code there is a concept of primary and secondary serial interfaces that operate very differently. Handling of protocols is different between interfaces and you cannot have multiple connections at once using the TunerStudio protocol. 

This PR separates the protocol handling from the interface that is being used to sent/receive the data, meaning that any protocol can be handled over any interface. Separate interface instances are used to track in-progress transmissions with each interface maintaining its own payload buffer, state variables etc. 

This is working well on Teensy and stm32 however it does not currently work on AVR due to the C++ memory library (https://en.cppreference.com/w/cpp/header/memory.html) not being part of the Arduino library subset. This is needed due to the use of unique_ptr for safe dynamic arrays, though there are other ways of doing this. 

This is good functionality and I think it's important for reliable comms via the secondary serial interfaces, but it will likely need some work/discussion before implementation.


adbancroft:
> This is working well on Teensy and stm32 however it does not currently work on AVR due to the C++ memory library (https://en.cppreference.com/w/cpp/header/memory.html) not being part of the Arduino library subset.

I've [implemented](https://github.com/adbancroft/speeduino/tree/pr/comms_split_noptr) a fix for this by implementing a 'span' class which is injected into `commsInterface` (a span is a container that [doesn't own it's buffer:](https://en.cppreference.com/w/cpp/container/span.html)). 

Injecting the buffer into `commsInterface` is a better solution anyway - clearer separation of responsibilities.

adbancroft:
> I've [implemented](https://github.com/adbancroft/speeduino/tree/pr/comms_split_noptr) a fix for this by implementing a 'span' class 

My preference would be to consume an implementation of the C++ standard library. E.g. [https://github.com/modm-io/avr-libstdcpp](https://github.com/modm-io/avr-libstdcpp). Doing so means no more re-implementing of standard C++ idioms such as `span`, `<type_traits>` etc.

PR #1455, opened 2026-04-09
STM32F407 USB failure when SD and Secundary serial are enabled

Antonecu:
This fix prevents the permanent hang-up of the Native USB port on STM32F407 boards when SD card and secundary serial are initialized during boot.

### DIFF
--- a/speeduino/board_stm32_official.cpp
+++ b/speeduino/board_stm32_official.cpp
@@ -345,6 +345,13 @@ STM32RTC& rtc = STM32RTC::getInstance();
     #endif
 
     Serial.begin(baudRate);
+
+    #if defined(STM32F407xx)
+	  uint32_t usbStart = millis();
+      while (!Serial && (millis() - usbStart < 2000)) {
+        yield();
+      }
+	#endif
   }
 
   uint16_t freeRam()


speeduino/board_stm32_official.cpp
```diff
@@ -345,6 +345,13 @@ STM32RTC& rtc = STM32RTC::getInstance();
     #endif
 
     Serial.begin(baudRate);
+
+    #if defined(STM32F407xx)
+	  uint32_t usbStart = millis();
+      while (!Serial && (millis() - usbStart < 2000)) {
+        yield();
+      }
+	#endif
```
adbancroft, line 354: Could you move this code into it's own function? E.g. `static void waitForSerialToStart(SerialUSB &serial, uint32_t timeOut)`

adbancroft:
How was this tested and on what hardware combo?

Antonecu:
Tested on a Speeduino v0.4.4 board using a Black STM32F407VGT6 via adapter. The hardware setup includes SD logging via SPI3 (PC10, PC11, PC12), Native USB (PA11, PA12), and Secondary Serial (PA9, PA10 or PD5, PD6).

The Issue: With an SD card inserted and Secondary Serial enabled, the Native USB connection fails/breaks during boot. This fix addresses that conflict.

adbancroft (changes requested):

PR #1456, opened 2026-04-11
Implemented Non-linear TPS

ruubinsson:
Hi,

I implemented a non-linear tps signal that works wonders with ITBs. It can be adjusted from the analog filters menu. If set  < 1 then it will be more sensitive in the low range and if set > 1 then it will be more sensitive in the high range.
Screenshot 2026-04-11 at 9 11 21

The analog value for the tps range has been altered use the full 10-bit analog input. This because of the logartimic/exponential value needing more values to work with to be stable.

This has been tested and works wonders with my ITB setup.


theKraid:
This is cool!

ruubinsson:
The sweet spot I have used for two different ITBs is 0.35 curvefactor.
graph

adbancroft (changes requested):
Since the page sizes & contents have changed, you will need changes to `update.cpp` to migrate tunes existing tunes.

adbancroft:
Converting to draft until feedback is addressed

PR #1457, opened 2026-04-14
Allow integerPID::Compute() to accept negative inputs

adbancroft:
Remove the `>0` restriction on `integerPID::Compute()` input values and unit test.

### DIFF
--- a/speeduino/src/PID/integerPID.cpp
+++ b/speeduino/src/PID/integerPID.cpp
@@ -49,38 +49,35 @@ bool integerPID::Compute(unsigned long now, long FeedForwardTerm)
    {
       /*Compute all the working error variables*/
 	   long input = *myInput;
-      if(input > 0) //Fail safe, should never be 0
+      long error = *mySetpoint - input;
+      long dInput = (input - lastInput);
+      FeedForwardTerm <<= PID_SHIFTS;
+
+      if (ki != 0)
       {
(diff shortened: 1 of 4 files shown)


adbancroft:
@VitorBoss @pazi88 @theKraid Could you review this? Since all it does is remove the `>0` check, I'm suspicious the tests are missing something. 

Could you confirm the unit tests are realistic & valid?

|  Test |Start Point|Set Point|
| ------------ | ------------ |------------ |
|test_end_to_end_negative_negative_up|-235|-155|
|test_end_to_end_negative_negative_down|-155|-235|
|test_end_to_end_positive_negative|63|-55|
|test_end_to_end_negative_positive|-55|65|

You can get more detailed test output using the `-v` flag. E.g. `pio test -e megaatmega2560_sim_unittest --filter test_pid -v`

theKraid:
@adbancroft I went and tested it on my car, unfortunately it still puts VVT into an error state when the angle is negative. See the log excerpt here:

image

adbancroft:
> @adbancroft I went and tested it on my car, unfortunately it still puts VVT into an error state when the angle is negative. 

Wow, wasn't expecting that. 

This PR only fixes the PID controller. Another PR incoming....see #1458

adbancroft:
Closing...replaced by #1458

PR #1458, opened 2026-04-14
Allow negative VVT angles.

### LINKED ISSUE #1454: VVT closed loop will fail when VVT angle goes negative
This issue boils down to this line in `integerPID::Compute`:
```
      if(input > 0) //Fail safe, should never be 0
```

1. This means the function will do nothing when the input is negative.
2. For VVT control, the input is the VVT cam angle. 
3. So when the VVT angle is less than zero, closed loop VVT control will not apply any correction :-(

adbancroft:
Fixes #1454


adbancroft:
@theKraid Try this PR - you'll need to used the updated INI that's included and edit your minimum VVT angle

theKraid:
@adbancroft Good thing you mentioned the "minimum cam angle" setting - I forgot to consider it when testing the previous PR #1457
I just tested this PR #1458 and set the "minimum cam angle" to "-10" and it's looking good! VVT is no longer falling into an error state when dipping into the negatives and it also seems to apply duty cycle.
I approve! Good job and thank you very much

image

adbancroft:
Thanks for testing.

@VitorBoss @pazi88 I'd still like your feedback on this, since the "fix" was to just remove the `>0` check which seems naive.

PR #1460, opened 2026-04-20
integerPID_ideal: use Derivative on Measurement

adbancroft:
In `integerPID_ideal`:
1. Remove the pre-clamping of the integral term. It's not required **and** we are backing out the error from it anyway
2. Use Derivative on Measurement in place of Derivative on Error so the calculation matches the other PID classes.

Existing tests all pass. End to end tests show very, very close output:
image

image


adbancroft:
@Tjeerdie Could you review?

PR #1466, opened 2026-04-23
Scatter output channels

### LINKED ISSUE #1461: Optimised Output Channels - Higher data read rates
Whilst working on https://vetuner.co.uk, I've got to know the INI definition PDF https://www.tunerstudio.com/index.php/support/manuals/tsdevmanuals/137-ecu-definition-specification-aka-the-ini-document quite well.

One intriguing section is 11.2, 'Optimised OutputChannel Capture'  It is basically a mechanism for the tuning software to request only the output channels it needs with respect to the currently used expressions.  I've done some 'what if' calculations of potential increases in data rate using it on Speeduino

|  | Frame Size | Max Rate |
|--|--| -- |
| Baseline | 127B | 61Hz |
| Gauges only | 34B | 120Hz |
| Gauges + full 103 field log | 104B | 69Hz |
| Gauges + log minus aux/outputStatus/vvt fields | 74B | 85Hz |

The speed is governed by the number of bytes at 115200 baud and about 5ms overhead per request/response pair

The cost is a RAM buffer of approx 2 bytes per requested field.  There is a bit of a balance here as if the buffer is too small, TS/VETuner will fall back to the default full read.  

I'm happy to create a PR for the Speeduino side of the code, but if this is seen as a profligate waste of RAM then it's a non-starter.

scudderfish:
Implements Section 11.2, 'Optimised OutputChannel Capture'  of the TunerStudio INI spec https://www.tunerstudio.com/index.php/support/manuals/tsdevmanuals/137-ecu-definition-specification-aka-the-ini-document and fixes #1461 in the process.  

The size of the buffer is debatable.  The larger it is, the less of an impact this can have as the more fields you pull, the lower the read rate.  Running the default dashboard I've been getting around 200Hz read rate.  Logging to file with about 22 fields gave about a 150Hz logging rate (to MLG format)  If you utilise more fields than is possible to fit in the buffer  TS & VETuner will fall back to the current ochGetCommand and ochBlockSize read.

I am unsure about the repeated definition of 'SCATTER_ARRAY_SIZE' across the header files.  It could be board specific dependent on board RAM capacity, but that would also need #ifdef in the INI file.


adbancroft:
1. Could this be modeled as something *other* than a new config page? In my head, the pages store **the** tune - which this isn't part of (it's not written to EEPROM for example).
2. Could most of the code be isolated to a new "scatter logging" module/controller, which comms then calls into?

scudderfish:
Very likely.  I went with a pseudo-page mainly for accounting for the memory but I'm happy to try other approaches.

PR #1470, opened 2026-05-01
Fix incorrect comparison causing only one tooth to be logged

### LINKED ISSUE #1469: Tooth logging on logs the first tooth
The comparison for knowing if the tooth logger buffer is full is reversed.  This means only the first tooth read is ever logged.

scudderfish:
Fixes #1469 

Simple comparison change to allow full use of the tooth log buffer


adbancroft:
Weird - just spotted this as well!

PR #1485, opened 2026-05-25
Refactor pin number storage (use structs) & initial pin assignments

adbancroft:
1. Organize pin numbers into structs, so they can be passed as a parameter (rather than 100+ global uint8_t): this enables unit testing and lays the runway for a future reduction in coupling to global variables.
2. Refactor initial pin assignments into one function per CPU shield: reduces cognitive complexity.
3. Use a fluent builder pattern to construct `pinNumbers_t` instances: reduces duplicate code
4. Store `pinNumbers_t` instances in flash: moves runtime code to compile time
5. Unit test pin mappings

Overall reduction in flash usage, RAM usage and increase in test coverage. 😄


itszeekay:
low value slop? what problem does this solve?

adbancroft:
> low value slop? what problem does this solve?

@itszeekay Have the updates addressed your concern(s)?

itszeekay:
@adbancroft My main concern here is this is a _very large_ commit and fundamentally changes how pins are defined. Reading through it makes the pin definitions for any given boards a lot harder to understand as a human and possibly more difficult to add new board layouts in the future.

Optimizations are great, but is 1% flash worth the maintenance increase and has this been actually tested in detail? I'd be very weary of this.

adbancroft:
> @adbancroft My main concern here is this is a _very large_ commit and fundamentally changes how pins are defined. 

Fair point, I cannot disagree.

> Reading through it makes the pin definitions for any given boards a lot harder to understand as a human and possibly more difficult to add new board layouts in the future.

If you are referring to the use of the builder pattern, I can partially agree. However, wading through a 1300 line switch statement interspersed with conditional compilation (`#ifdef`) isn't understandable either.

> I'd be very *weary* of this.

Appropriate typo....lol

adbancroft:
Closing - replaced by #1495

PR #1490, opened 2026-05-28
Only run stepper idle control once per loop 

### LINKED ISSUE #1489: Idle stepper taper running twice as fast
The function ``idleControl()`` is called twice in the ``speediono.ino`` file, one at 10Hz time and every loop, there are checks for the 10Hz flag inside the idle code, that makes the taper run twice as fast for stepper mode.

adbancroft:
Fixes #1489


adbancroft:
@VitorBoss Please review, thx.

PR #1494, opened 2026-06-02
Add support for Fat Chicken Racing Micro F4

smclt30p:
Adds support for the FCR Micro F4. The upstream linker script is broken, so a specific has been added.

Web: https://www.fatchickenracing.com


adbancroft:
I recommend adding the `FCR_Micro_F4` environment to the CI build GitHub action (`build-firmware.yml`)

smclt30p:
> I recommend adding the `FCR_Micro_F4` environment to the CI build GitHub action (`build-firmware.yml`)

I will add it eventually. Just wanted a buildable target that people can build.

smclt30p:
Is there any issues I need to address? These bots generate lots of noise.

speeduino/sensors.cpp
```diff
@@ -90,7 +90,12 @@ static inline uint16_t readAnalogPin(uint8_t pin)
   int tmp = analogRead(pin);
   // max is a macro on some platforms - DO NOT place the call to analogRead as an inline parameter:
   // (you might end up calling it twice)
-  return max(0, tmp);
+  tmp = max(0, tmp);
+#if defined(BOARD_ANALOG_SCALE_NUM) && defined(BOARD_ANALOG_SCALE_DEN)
+  tmp = (int)(((uint32_t)tmp * BOARD_ANALOG_SCALE_NUM) / BOARD_ANALOG_SCALE_DEN);
+  if (tmp > 1023) { tmp = 1023; }
+#endif
+  return (uint16_t)tmp;
```
adbancroft: What is this for? Why is it needed?

smclt30p: My ADC on my board is not full range, 5V gets converted to 3.2V and not 3.3V

smclt30p:
Any news?

speeduino/sensors.cpp
adbancroft: Makes sense. Could you wrap the `#if` into a separate function? E.g. `int16_t postProcessAnalogRead(int16_t pinValue)` And then change `readAnalogPin` to `return clamp(postProcessAnalogRead(analogRead(pin)),0, 1023);`

adbancroft: And use appropriate fixed width types. I.e. `int16_t`

adbancroft (changes requested):

adbancroft:
> Any news?

Looks great apart from a couple of minor niggles (noted above).

adbancroft:
> > I recommend adding the `FCR_Micro_F4` environment to the CI build GitHub action (`build-firmware.yml`)
> 
> I will add it eventually. Just wanted a buildable target that people can build.

I'm trying to keep `platform.ini` as the single source of truth for CI. I.e. if it's in `platform.ini`, it should be in the CI builds. So once you are happy with this PR, please update the CI build.

smclt30p:
I updated the PR and CI files, please merge when possible. Br, Ognjen

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

smclt30p:
What's happening with this? This is taking ages. How is anything getting done?

adbancroft:
> What's happening with this? This is taking ages. How is anything getting done?

1. I'm on vacation
2. The unit tests are failing & I'm trying to figure out why: happy to have some help doing so.

smclt30p:
I let Claude loose on this, here is what he says:

Yes — now I've actually root-caused it, and it's not your branch's code.

The answer

The SIGBUS is a simavr 1.7 bug. I built the simavr 1.8 you downloaded and ran the unmodified branch under it:

157 Tests 0 Failures 52 Ignored.
OK.

No SIGBUS, accuracy tests pass. I confirmed it both ways — original branch and with my change both pass on 1.8. So the crash lives in simavr 1.7's peripheral model, not in the firmware.

Why it bit this branch / why I chased the wrong thing first

- PlatformIO's tool-simavr is pinned at 1.10700.241009 = simavr 1.7, and the registry has nothing newer — so CI runs the buggy simulator.
- On my x86 mac I couldn't run the pinned tool-simavr (it's an arm64 binary), and Homebrew's simavr 1.7 has a broken Timer0, so locally the same 1.7 bug surfaced as a serial-output hang instead of your SIGBUS. Different symptom, same root (simavr 1.7 mishandling the test's I/O).
- I also ruled out the two mechanical suspects: stack overflow (deepest stack 0x20C6, 425 bytes of headroom) and RAM pressure (.data/.bss byte-identical to master).

Fix options

1. Root fix — get CI onto simavr 1.8. Proper, but there's no PlatformIO 1.8 package, so the CI workflow would have to build simavr (like I just did) and point the test runner at it. This also fixes master, which is affected the same way.
2. My unity_config workaround (polled UDR0 output instead of the interrupt-driven Serial ring buffer). It made the buggy simavr 1.7 pass locally, and your SIGBUS happens during a Unity output print — the exact path it replaces — so it very likely kills the SIGBUS too. Self-contained, no CI infra change.

I can't run your arm64 tool-simavr to prove #2 against the SIGBUS directly. Could you apply the current branch (my test/unity_config.{h,cpp} + the one-line platformio.ini change) and re-run in the environment that SIGBUSed? That's the one definitive check. If it passes, #2 is the clean fix; if not, we go with #1 (simavr 1.8 in CI).

Two housekeeping notes: the simavr/ folder you added is untracked — don't commit it. And if we go with option 1, I'd revert the unity_config change since it'd no longer be needed.

Which way do you want to go?

adbancroft:
> and your SIGBUS happens during a Unity output print

That was enough of a clue to put a workaround in place, thx.

PR #1495, opened 2026-06-03
Refactor pin number storage (use structs) & initial pin assignments (v2)

adbancroft:
1. Group all pin numbers into a struct (`pinNumbers_t`), rather than 45+ global `uint8_t` variables: this enables unit testing and lays the runway for a future reduction in coupling to global variables.
2. Refactor initial pin assignments into one function per CPU shield: reduces cognitive complexity.
3. Unit test pin mappings


adbancroft:
@itszeekay Any feedback on this PR?

PR #1497, opened 2026-06-06
Refactor PID classes

adbancroft:
1. Decouple construction, PID settings, input & output. Currently, most (but not all!) settings, input & output are passed in during class construction, often as pointers. This tightly couples the PID object to it's input & output, which means the PID object lifetime and input/output types control those aspects of the consumer. I.e. it's a very leaky abstraction.
2. Step 1 allows removal of shadow variables. E.g. `statuses::longRPM`
3. Eliminate the unwritten ordering of calls to PID setting functions. E.g. `PID::SetOutputLimits` & `PID::SetMode`
4. Share common types & algorithms. See `PidTuningParameters` & `PidCore`
5. Remove the concept of controller direction: it's not required, since the PID algorithm works on positive and negative inputs.
6.  Inject current time into methods, rather than calling `millis` internally: simplifies code & tests.
7. Consistent method names that follow Speeduino convention
8. Original unit tests still pass :-)

Simpler code, lower RAM usage, lower flash usage, increased test coverage :-)


adbancroft:
@theKraid Any chance you could test this PR? There should be no change in behavior, but it does impact VVT code (and boost). Thx

theKraid:
@adbancroft I would love to help but I wont have a chance to test this on a real car with VVT any time soon. Sorry!

adbancroft:
> @adbancroft I would love to help but I wont have a chance to test this on a real car with VVT any time soon. Sorry!

NP, thx for replying.

PR #1504, opened 2026-06-16
Add per-cylinder ignition trim for sequential ignition

### LINKED ISSUE #143: ICIT Ignition Trim/Offset
Individual Cylinder Ignition Trim enhancement discussed long ago and perhaps just an offset correction value for now with possible table-driven solution following.  Highlighted in this forum thread:  [https://speeduino.com/forum/viewtopic.php?f=12&t=1880](https://speeduino.com/forum/viewtopic.php?f=12&t=1880)

AJ0070:
Adds individual cylinder ignition trim support for sequential ignition setups.

This introduces per-cylinder ignition offset values in degrees and applies them only when ignition is running in true full sequential mode. Shared-output modes such as wasted spark continue to use the untrimmed base advance.

## Changes

- Add `ignTrim[8]` storage to page 13 config data
- Apply per-cylinder ignition trims during non-rotary ignition angle calculation
- Restrict trim usage to full sequential ignition only
- Add TunerStudio ini fields, defaults, help text, and UI for cylinder trims
- Add unit tests covering:
  - trims applied in sequential mode
  - trims ignored in wasted spark mode

## Notes

- Positive trim values add advance
- Negative trim values retard timing
- This is implemented as a fixed per-cylinder offset, matching the issue request for an offset-based first step

fixes #143


scribaecho-dotcom:
I pulled `AJ0070:fix/143` locally and independently verified it on the simulator / firmware build:

- `platformio test -v -e megaatmega2560_sim_unittest -f test_schedules` -> 159 cases: 107 succeeded, 52 skipped, 0 failed
- `platformio run -e megaatmega2560` -> SUCCESS

I also checked that the implementation matches the offset-first scope from #143: fixed per-channel offsets, applied only in true sequential ignition, with wasted/shared-output ignition left untrimmed.

Looks like a good direction for the initial ICIT implementation.

adbancroft:
Do you need to set the offsets to zero for older tunes? (in `updates.cpp`)

speeduino/scheduler_ignition_controller.cpp
```diff
@@ -307,56 +317,65 @@ static inline void calculateRotaryIgnitionAngles(uint16_t dwellAngle, const stat
 #endif
 }
 
-static inline void calculateNonRotaryIgnitionAngles(uint16_t dwellAngle, const statuses &current)
+static inline bool hasIndividualIgnitionTrim(const config2 &page2, const config4 &page4, const statuses &current)
```
adbancroft: Could `isFullSequentialIgnition` be used instead of this function?

AJ0070:
@adbancroft Good catch. Yes, these bytes were previously unused in page 13, so older tunes could inherit non-zero values.

I've added an `updates.cpp` migration that bumps the EEPROM data version to 27 and explicitly zeros all `configPage13.ignTrim[]` values for existing tunes.

speeduino/scheduler_ignition_controller.cpp
```diff
@@ -23,6 +23,16 @@ IgnitionSchedule ignitionSchedule8(IGN8_COUNTER, IGN8_COMPARE); //cppcheck-suppr
...
 constexpr table2D_u8_u8_8 rotarySplitTable(&configPage10.rotarySplitBins, &configPage10.rotarySplitValues);
 
+static inline int8_t constrainAdvanceTrim(int16_t advance)
+{
+  return (int8_t)clamp(advance, (int16_t)INT8_MIN, (int16_t)INT8_MAX);
+}
+
+static inline int8_t getIgnitionTrimmedAdvance(int8_t baseAdvance, uint8_t channelIndex)
+{
+  return constrainAdvanceTrim((int16_t)baseAdvance + (int16_t)configPage13.ignTrim[channelIndex]);
```
adbancroft: Could you pass in the `config13` instance as a parameter instead of using the `configPage13` global? You'll have to pass it down the entire call stack, starting with `calculateIgnitionAngles`. This will also require unit test changes (I'm trying to move towards a more testable code base with less hidden coupling)

adbancroft (changes requested):
I like what I see, just a couple of small feedback items.

speeduino/scheduler_ignition_controller.cpp
AJ0070: done

speeduino/scheduler_ignition_controller.cpp
AJ0070: done

AJ0070:
@adbancroft I have addressed both comments

speeduino/updates.cpp
```diff
@@ -64,9 +64,22 @@ TESTABLE_STATIC void upgradeV25toV26(void) {
   }
 }
 
+TESTABLE_STATIC void upgradeV26toV27(void) {
+  if(loadEEPROMVersion() == 26U)
+  {
+    for (uint8_t i = 0; i < sizeof(configPage13.ignTrim); i++)
```
adbancroft: Possibly more robust to use `_countof` here instead of `sizeof`?

adbancroft (changes requested):
Looking good, a couple of minor comments

speeduino/updates.cpp
AJ0070: done

AJ0070:
@adbancroft CI is green

adbancroft:
Looking good. I'd like to leave this open for a couple of weeks for others to give further feedback, since it's a new feature (plus the original request has been open for ~8 years)

AJ0070:
@PSIG-dev since this is originally your requested feature please let me know if you have any feedback.

adbancroft:
Converting to draft until we have feedback from @PSIG-dev

PSIG-dev:
Bummer I missed all the work as my notifications were not getting through.  But thanks for the updates on this!  I need to be clear, that stepping up to full-time cylinder offset is helpful for establishing feature function; offset by table (like fuel trims) would be required for the feature to have best effect.  So IMO this is a good first step to validate function, with some limited operational improvements; but finalizing with tables is ultimately required.  The equivalent fuel trim function is the final goal.

In order to tie full function with desired effect, I will example.  Your 6-cylinder engine detonates on cylinder 3 from 2200 to 3000 rpm at full-throttle.  Currently, the timing is pulled for all cylinders in that range, in order to avoid det in Cyl3.  This limits performance in all cylinders, while necessarily avoiding det on Cyl3.  Setting full-time offset on Cy3 with the function as-is would increase performance of the other cylinders while in that range, however Cyl3 will then have incorrect timing and down on performance _everywhere_ except in the limited range it requires offset.

As-is the function is a tradeoff, of all cylinders altered in a specific range, against one cylinder altered all the time.  I hope that ties the ultimate goals to the function, and establishes the next step to gain full effects from the function.  While this first step may be useful in some situations, the function would be greatly improved with tabling added.  Is that where this effort is planned to go?  Sorry for the wall of text, but wanted to be clear on the usefulness.

adbancroft:
Sounds like this PR is good enough to merge.

@PSIG-dev Assuming you mean 1 table per cylinder (like fuel trims), EEPROM space will be the limiting factor for a table based solution.

PSIG-dev:
@adbancroft – Do you indicate tables would, would not, or be limited in size or number of tables?

While there is some utility in the current version it is very limited of course, and useful to some hobby engine applications.  I did not anticipate the feature forcing always-on timing shift at all rpm.  The more precise and output of the engine, the more per-cylinder tables and cells are required for effectiveness.  The utility of fuel trims is equally important for ignition, as they are co-dependent.  The applications for this are numerous, simply from the ability to extract maximum output per-cylinder, as "_an engine is only a powerful as the weakest cylinders_" applies.  Detonation is but one example of correcting handicap.

Moving forward and if limited in space for tables, a minimum of a couple tables that are assignable to specific IGN channels would cover a much greater number of conditions and applications more effectively.  For example two tables (presuming the worst of the cylinders), such as Cyl3 of the example above, with ICIT Table1 assignable to IGN3.  While still limited, at least the greater ability of tables could be made available for the much greater benefits, to the worst of the cylinders.  

Alternatively but still useful, instead of tables could rpm simply be assigned for a single correction range per-cylinder?  Perhaps **IGN Channel __ ; +Adv/-Ret ___° ; Low RPM _____ to High RPM _____** .  Could that free enough valuable space?  The ignition feature is an example of advanced function that takes Speeduino from 'hobby' to serious, and I look forward to exploring it.  Thank you for your time and effort to set the foundation!

adbancroft:
> Do you indicate tables would, would not, or be limited in size or number of tables?

Both: EEPROM usage = number of tables * table size.

* Each fuel trim table is 48 bytes
* 8 trim tables = 384 bytes
* Currently we have 393 bytes of unused space on the EEPROM
  * Largest block is 217 bytes, though we can move things to increase this.

Would 2d table(s) of RPM vs Advance Correction work? Or is load also required?

PSIG-dev:
Eh... 3D targets conditions far more effectively and would be preferred, such as adding interpolated advance for a cylinder with weak ring sealing, which decreases with load.  I think the issues are going to be a budget of bits, versus compromises to the function, in order to allow greatest benefit to the most users.  Limiting to 2D as I suggested earlier as a compromise certainly reduces the effectiveness of the feature substantially, but if compromise is required would be preferential to always-on timing shift for example.  Each function in the feature multiplies the effectiveness.

To be most helpful in assisting best compromises for useful function of the feature, we would need to know which options could fit into that budget, for best user balance of usefulness.  So thank you for the question of 2D vs 3D, as an example of that option.  Is that kind of option vs space budget info possible without too much effort, such as [example] using only 3 interpolated 3D tables vs 4, 6, or 8, but make them assignable to specific channels (compromise option of higher capability but only x-channels of greatest need)?  

If this type of options info was easily available, then the Mega2560 may have limitations, whereas the STM32 or similar might not, and simply expands coverage in those versions with greater EEPROM budget.  Depending on the effort in budgeting options, could it be more efficient to approach from the higher-capacity uC versions for max function, trimming for budget compromise toward lower-capacity uCs?

adbancroft:
> Is that kind of option vs space budget info possible without too much effort

Yes, we can estimate the space budget for multiple options.

adbancroft:
> then the Mega2560 may have limitations, whereas the STM32 or similar might not

I don't think that's the case. E.g. STM32 battery backed SRAM is also 4Kb.

adbancroft:
I think we would also need overwhelming community support for any feature that will consume the final bytes of EEPROM.

adbancroft:
Wild ideas:
1. shrink the size of existing EEPROM items? 
2. reuse some fuel trim tables for ignition? E.g. assign N tables to fuel trims for user chosen cylinders and M tables to ignition trims for user chosen cylinders (where N+M<=8).

PSIG-dev:
I would not add full-time timing offset, as the intent is to only alter a cylinder's timing when it needs it, not all the time.  If I can't use it to solve a passing issue (e.g., detonation on cylinder 3 from 2000-2800 rpm) without recovering timing where it isn't needed, then few would likely use it.  

IMO, it would be counter-productive to cripple a cylinder everywhere for a specific condition.  We may have that capability already in some cases as outlined in the link earlier, using oddfire settings for example. dontknow

TL;DR - if the function cannot be focused with 3D tables, I would wait until it could be.  Perhaps compromises for space @adbancroft suggested might make it possible, or for use on PCBs with expanded EEPROM?  Thanks!

AJ0070:
I ran the numbers against the actual EEPROM layout, from `print_eeprom_layout()` in `test_storage` on current master plus a `megaatmega2560` build.

@adbancroft's totals check out: 392 unallocated bytes, largest block 217. The catch is fragmentation. That 392 is spread over 24 gaps and 22 of them are 2 to 9 byte scraps of inter-entity padding. Usable today is 336 bytes in two pieces: 217 at 3457-3674 and 119 at 3686-3805.

Two ways to do better, neither of which moves user data:

`config15` is 176 bytes at 3281 but only 26 are used, the rest being `Unused15_106_255[150]`. That tail sits at 3307-3457, directly adjacent to the 217 byte gap. Shrinking the struct merges them into 367 contiguous bytes. It's the last page entity with only free space above it, so nothing relocates; the migration is just zeroing new fields.

The three calibration CRCs (4 bytes each at 3674/3678/3682) are the only thing splitting the 217 and 119 gaps, and they fit into dead scraps elsewhere: two in the 9 byte gap at 1701, one in the 5 byte gap at 2575. That gives 498 contiguous above 3307.

A `table3d6RpmLoad` is 48 bytes, same as a fuel trim table. `table3d4RpmLoad` (24 B) and `table3d8RpmLoad` (80 B) already exist as generated types, so none of these need new codegen. Sizes include the 2 byte padding between entities:

| Option | Res | Channels | EEPROM | Fits in |
|---|---|---|---|---|
| offset + RPM window | lo/hi | 8 | 16 | page 13 spare plus a few |
| 2D RPM to trim, 6 bins | 2D | 8 | 96 | 217 gap, today |
| 3D 4x4 | 3D | 8 | 206 | 217 gap, today |
| 3D 6x6, channel-assignable | 3D | 4 | 198 | 217 gap, today |
| 3D 6x6, channel-assignable | 3D | 7 | 348 | 367, needs the config15 reclaim |
| 3D 6x6 | 3D | 8 | 398 | 498, needs both |
| shared fuel/ign pool, N+M<=8 | 3D 6x6 | up to 8 | ~2 | page 13 spare |

So 8 full 6x6 tables are reachable with 100 bytes left over, and 3D on all 8 channels fits in currently free space today if 4x4 resolution is acceptable.

RAM is the tighter constraint on the Mega and hasn't come up yet. Tables are RAM resident and each carries a 7 byte interpolation cache, so a 6x6 costs 55 bytes and a 4x4 costs 31. The current build uses 6706 of 8192 with 1486 free, so eight 6x6 tables take 440 of that, about 30% of the headroom. Eight 4x4 take 248. Flash has roughly 50 KB free.

Happy to prototype whichever option looks best.

PR #1525, opened 2026-07-10
Fix micros() rollover in stepper IAC step timing

gdiciocco:
## Summary

In `checkForStepping()` the end of the STEP/COOLING phase was detected with:

```cpp
if(micros() > (idleStepper.stepStartTime + timeCheck) )
```

When a step starts within `timeCheck` microseconds of the 32-bit `micros()` wrap (every ~71.6 minutes), `stepStartTime + timeCheck` overflows to a small value and the condition becomes true immediately. This truncates the STEP pulse below the configured `iacStepTime`, which can cause the motor to miss the step and permanently desync `curIdleStep` from the real pintle position (never recovered until the next homing).

This PR replaces the comparison with the standard rollover-safe form:

```cpp
if( (micros() - idleStepper.stepStartTime) > timeCheck )
```

## Testing

- Compiled successfully for `megaatmega2560`.
- Unsigned-subtraction rollover behaviour verified by inspection: `micros() - stepStartTime` yields the correct elapsed time across the wrap.

### DIFF
--- a/speeduino/idle.cpp
+++ b/speeduino/idle.cpp
@@ -240,7 +240,7 @@ static inline uint8_t checkForStepping(void)
       timeCheck = iacCoolTime_uS;
     }
 
-    if(micros() > (idleStepper.stepStartTime + timeCheck) )
+    if( (micros() - idleStepper.stepStartTime) > timeCheck )
     {         
       if(idleStepper.stepperStatus == STEPPING)
       {


adbancroft:
This won't be the only places this can.should be used.

Any way that this could be abstracted into a function? E.g. `template <typename T> bool rolloverSafeCompare(T toCompare,  T baseline, T threshold)`

That function could then be unit tested.

gdiciocco:
Agreed — I audited the whole codebase for this pattern and you're right, this was not the only site. Done in #1528:

- The comparison is now abstracted in `elapsed_time.h` as `hasIntervalElapsed(now, start, interval)` plus a value-returning `timeElapsed(now, start)`, introduced in gdiciocco/speeduino@9f29c29f. `now` is injected instead of calling `micros()` internally, so both are unit tested with values on either side of the wrap (normal case, exact boundary, and both failure modes of the naive comparisons).
- I went with plain `uint32_t` functions rather than a template for now: every time base in the codebase (`micros()`, `millis()`, `runSecsX10`) is 32-bit unsigned. Trivial to generalise later if needed. Also note `elapsedTime` as a name was taken by a file-scope static in decoders.cpp.
- The audit found three more absolute-deadline comparisons with the same bug as this PR: overdwell protection (which can either cut a charging coil spuriously or leave it on past the limit at the wrap), the EEPROM write defer (whose main-loop rollover workaround becomes unnecessary and is removed), and the acceleration enrichment end time. All fixed in #1528, with wrap-crossing unit tests for the overdwell path.

This PR can either be merged as-is (#1528 rebases cleanly and migrates this site to the helper), or closed in favour of #1528 — whichever you prefer.

adbancroft:
Closing in favor of #1528

PR #1528, opened 2026-07-11
Add rollover-safe elapsed time helpers and fix remaining micros() wraparound bugs

gdiciocco:
## Summary

Follow-up to #1525, implementing @adbancroft's suggestion to abstract the rollover-safe comparison into a unit-testable function — and then using it to fix the remaining wraparound bugs an audit of the codebase turned up.

### 1. The helpers (`speeduino/elapsed_time.h`) — gdiciocco/speeduino@9f29c29f

```cpp
static inline uint32_t timeElapsed(uint32_t now, uint32_t start);
static inline bool hasIntervalElapsed(uint32_t now, uint32_t start, uint32_t interval);
```

Design notes:

- `now` is a parameter rather than calling `micros()` internally, so the functions can be unit tested with values on either side of the wrap point.
- The boundary is inclusive (`elapsed >= interval` counts as elapsed), documented in the header and pinned by a test.
- Plain `uint32_t` functions instead of a template: every time base in the codebase (`micros()`, `millis()`, `runSecsX10`) is 32-bit unsigned. Easy to template later if a different-width counter ever appears.
- Named `timeElapsed` rather than `elapsedTime` because decoders.cpp already has a file-scope static variable of that name.

Unit tests in `test/test_math/test_elapsed_time.cpp` cover the normal case, the exact boundary, and both historical failure modes (overflowing `start + interval` sum; deadline stored past the wrap never being reached).

### 2. Bug fixes (same class of bug as #1525)

Auditing all time comparisons found three more absolute-deadline sites:

- **Overdwell protection** (`scheduler_ignition_controller.cpp`): compared `schedule._startTime < micros() - dwellLimit`. Just after the `micros()` wrap the target underflows and running coils are cut spuriously; a dwell spanning the wrap could instead leave the coil charging — the exact failure this protection exists to prevent. Now takes `(now, dwellLimit_uS)` and compares elapsed time. Added wrap-crossing unit tests to `test_overdwell.cpp`.
- **EEPROM write defer** (`comms.cpp`): stored an absolute deadline (`micros() + delay`). Now stores start + interval, which also makes the manual rollover workaround in the main loop (`if(currentLoopTime > micros())`) unnecessary — removed.
- **Acceleration enrichment** (`corrections.cpp`): stored its end time in `statuses.AEEndTime`; near the wrap the enrichment could be cut short. Renamed to `AEStartTime`, timeout computed as elapsed >= aeTime.

The IAC stepper site from #1525 is also migrated to the helper.

### 3. Cleanup migration

Behaviour-preserving rewrite of the sites that already used the open-coded subtraction idiom (serial rx timeout, knock retard timing, VSS pulse age, flex pulse width, EEPROM reset button, VVT warm-up delay, `sharedEngineIsRunning()`), so the idiom is centralised and a bare `micros()`/`millis()` comparison stands out in review. The scheduler's signed elapsed calculation is intentionally left as-is (it needs a value, not a boolean).

## Testing

- Full unit test suite on simavr (`megaatmega2560_sim_unittest`): 19 suites, 1225 tests passed, 0 failed.
- New tests: 4 in `test_elapsed_time.cpp`, 2 wrap-crossing overdwell tests.

### DIFF
--- a/speeduino/comms.h
+++ b/speeduino/comms.h
@@ -37,9 +37,9 @@ bool isRxTimeout(void);
  * Serial comms can send data quicker than we can write it to permanent storage.
  * This is used to manually throttle the writes so that we don't stall the main loop. 
  * 
- * @param time Absolute time in µS 
+ * @param delay_uS How long from now (in µS) before storage writes can resume
  */
-void setStorageWriteTimeout(uint32_t time);
+void setStorageWriteTimeout(uint32_t delay_uS);
 
 /** @brief Test if the timeout set by @ref setStorageWriteTimeout has expired */
 bool storageWriteTimeoutExpired(void);
--- a/speeduino/comms.h
+++ b/speeduino/comms.h
@@ -31,17 +31,7 @@ void serialTransmit(void);
 */
 bool isRxTimeout(void);
 
-/**
- * @brief During serial comms, defer storage writes
- * 
- * Serial comms can send data quicker than we can write it to permanent storage.
- * This is used to manually throttle the writes so that we don't stall the main loop. 
- * 
- * @param delay_uS How long from now (in µS) before storage writes can resume
- */
-void setStorageWriteTimeout(uint32_t delay_uS);
-
-/** @brief Test if the timeout set by @ref setStorageWriteTimeout has expired */
+/** @brief Test if the current storage write defer interval has expired. */
 bool storageWriteTimeoutExpired(void);
 
 #endif // COMMS_H
--- a/speeduino/auxiliaries.cpp
+++ b/speeduino/auxiliaries.cpp
@@ -5,6 +5,7 @@ A full copy of the license may be found in the projects root directory
 */
 #include "auxiliaries.h"
 #include "globals.h"
+#include "elapsed_time.h"
 #include "maths.h"
 #include "src/PID/integerPID.h"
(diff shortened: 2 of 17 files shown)


speeduino/comms.h
```diff
@@ -37,9 +37,9 @@ bool isRxTimeout(void);
  * Serial comms can send data quicker than we can write it to permanent storage.
  * This is used to manually throttle the writes so that we don't stall the main loop. 
  * 
- * @param time Absolute time in µS 
+ * @param delay_uS How long from now (in µS) before storage writes can resume
  */
-void setStorageWriteTimeout(uint32_t time);
+void setStorageWriteTimeout(uint32_t delay_uS);
```
adbancroft: `setStorageWriteTimeout` is only used within `comms.cpp`? So it could be given internal linkage & removed from this header?

adbancroft (changes requested):

speeduino/comms.h
gdiciocco: Done in 0401a1df. `setStorageWriteTimeout()` now has internal linkage in `comms.cpp` and its declaration has been removed from `comms.h`. I also audited the other symbols introduced by this PR; their linkage is already appropriately scoped. The full simavr suite passes: 19 suites, 1,225 tests passed.

PR #1530, opened 2026-07-14
Make flood clear a latched fuel cut with ATDC ignition retard

### LINKED ISSUE #1054: Flood clear operation not reliable.

gdiciocco:
## Problem

Fixes #1054.

Flood clear currently deactivates the instant RPM exceeds the cranking threshold. A combustion kick on residual fuel can spike the decoder RPM past `crankRPM` (the crank/run hysteresis is only 15 RPM), flipping the ECU into the running state and resuming full-load fuel with the throttle held wide open — re-flooding the engine the driver is trying to clear, running it on one or two unflooded cylinders, and diluting the oil. It also makes flood clear unusable for its second common purpose: building oil pressure by cranking without fuel.

## Change 1: latch the fuel cut

Once triggered (cranking with TPS at or above the threshold), flood clear now **latches**:

- Fuel stays hard-cut (pulsewidths forced to zero) even if the engine spins past `crankRPM` on residual fuel, and the crank→run transition is blocked while latched, so the engine cannot enter the running state.
- The latch is released only once the TPS has been held below 1% for more than 3 seconds, or by an ECU restart. The release timer runs at the TPS read rate even while the engine is stopped, and the latch intentionally survives stall detection. This is slightly stricter than the "release throttle below the flood clear value" proposed in #1054, so that a partially-lifted throttle can't resume fuel mid-crank unexpectedly.

## Change 2: retard ignition to 5° ATDC while latched

With no fuel injected, the only combustion possible is from residual mixture. Firing it before TDC produces torque against or with the starter — enough to kick back or momentarily spin the engine on the fuel being purged. While the latch is set, ignition advance is overridden to a fixed **5° ATDC** (within the existing supported cranking angle range of −10..80): any residual burn happens with the piston already descending (no kickback, almost no torque) while the plugs keep firing to dry themselves and the late burn helps vaporise the excess fuel. The angle is deliberately not later than −5° so mixture doesn't burn into an open exhaust valve at cranking speed. Note: no effect when `ignCranklock` is enabled, as the decoder bypasses the advance value.

## Related fixes

- Priming and the fuel cut disagreed at `TPS == floodClear` (priming used `<=`, the cut used `>=`); flood clear now wins at equality in both places, and priming is also suppressed while the latch is set.
- `floodClear = 0` used to match any TPS reading, permanently cutting all cranking fuel (an undiagnosable no-start); 0 now disables the feature.
- `ANALOG_ISR` (AVR): `AnChannel` reads are now atomic (16-bit reads can tear on 8-bit AVR), and `initialiseTPS()` re-enables the ADC interrupt and waits one full conversion cycle so the boot-time TPS reading used for the flood clear/priming decision cannot be stale.
- Teensy now sets `analogReadResolution(10)` explicitly instead of relying on the core default; corrected a misleading "12 bit" comment in the STM32 branch.
- TunerStudio ini: the FLOOD CLEAR indicator now matches firmware semantics and the tooltip documents the latch, the retard and the 0 = disabled case.

## Testing

- 13 new/updated unit tests covering: trigger/no-trigger, exact-threshold boundary, threshold-0 disabled, latch holding through Running/Stopped states, release timing (3.0 s not enough, >3 s releases), release-timer reset on a throttle blip, priming suppression (at-threshold / while-latched / when-disabled), and the ignition override.
- Full suite passes on simavr (1278 tests): `pio test -e megaatmega2560_sim_unittest`
- Firmware builds clean for `megaatmega2560`, `black_F407VE` and `teensy41`.

### DIFF
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -2179,7 +2179,7 @@ menuDialog = main
   dwellErrCorrect   = "A basic closed loop adjustment will be made to the dwell time to account for variations due to accel/decel. This is generally only needed on lower resolution trigger arrangements"
 
   crankRPM          = "The cranking RPM threshold. When RPM is lower than this value (and above 0) the system will be considered to be cranking"
-  tpsflood          = "Keep throttle over this value to disable the priming pulse and cranking fuel. Used to prevent flood or clear already flooded engine, and to build oil pressure by cranking without fuel. Once triggered the fuel cut latches (the engine cannot start even if it spins past the cranking RPM) until the throttle is held below 1% for 3 seconds or the ECU is restarted. Set to 0 to disable flood clear entirely."
+  tpsflood          = "Keep throttle over this value to disable the priming pulse and cranking fuel. Used to prevent flood or clear already flooded engine, and to build oil pressure by cranking without fuel. Once triggered the fuel cut latches (the engine cannot start even if it spins past the cranking RPM) until the throttle is held below 1% for 3 seconds or the ECU is restarted. Whilst the cut is latched the ignition is retarded to 5 degrees ATDC so residual fuel keeps being burned off without producing torque. Set to 0 to disable flood clear entirely."
 
   fanInv            = ""
   fanHyster         = "The number of degrees of hysteresis to be used in controlling the fan. Recommended values are between 2 and 5"
--- a/reference/speeduino.ini
+++ b/reference/speeduino.ini
@@ -2179,7 +2179,7 @@ menuDialog = main
   dwellErrCorrect   = "A basic closed loop adjustment will be made to the dwell time to account for variations due to accel/decel. This is generally only needed on lower resolution trigger arrangements"
 
   crankRPM          = "The cranking RPM threshold. When RPM is lower than this value (and above 0) the system will be considered to be cranking"
-  tpsflood          = "Keep throttle over this value to disable the priming pulse and cranking fuel. Used to prevent flood or clear already flooded engine"
+  tpsflood          = "Keep throttle over this value to disable the priming pulse and cranking fuel. Used to prevent flood or clear already flooded engine, and to build oil pressure by cranking without fuel. Once triggered the fuel cut latches (the engine cannot start even if it spins past the cranking RPM) until the throttle is held below 1% for 3 seconds or the ECU is restarted. Set to 0 to disable flood clear entirely."
 
   fanInv            = ""
   fanHyster         = "The number of degrees of hysteresis to be used in controlling the fan. Recommended values are between 2 and 5"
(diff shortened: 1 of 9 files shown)


noisymime:
I think a better option here would be to have a temporary RPM limit when coming off flood clear. The number of people I've seen who have used flood clear as a means of starting the engine after flooding it is amusingly high and this change would mean that this (admittedly janky) method wouldn't ever work. It's simile to the latching approach, but have the temp RPM limit be in place until that TPS drops back to an idle type level.

gdiciocco:
After considering the use case you pointed out, together with the ambiguity in my original proposal, I think I’ve come to the conclusion that flood clear is really trying to solve two different problems, and it would be better to support them explicitly with two operating modes.

My original PR still allowed ignition during flood clear, meaning the engine could still produce combustion torque from residual fuel. That makes it a compromise: it isn’t a true workshop mode, but it also doesn’t intentionally support the driver use case you described.

Service / Racing mode

This mode is intended for workshop procedures such as oil-pressure priming, diagnostics or drying a flooded engine. Once flood clear is activated, it latches until TPS has been below 1% for more than three seconds. While latched, both fuel injection and ignition are disabled, guaranteeing that the engine cannot start or produce torque under any circumstances.

Driver mode

This mode is intended to recover a flooded engine as quickly as possible. It could begin with a short period of complete fuel cut to purge excess fuel, then transition to a heavily reduced cranking fuel quantity while keeping the normal cranking ignition strategy. This allows the engine to start intentionally while flood clear is still active, rather than relying on residual fuel to fire by chance.

Once the engine exceeds the cranking threshold, a temporary RPM limiter (for example around twice crankRPM) would prevent it from immediately accelerating into a high-load condition. The limiter would then expire automatically after only a few seconds of running.

I slightly prefer a short time-based release of the limiter over waiting for the engine to return to idle. In the situations where flood clear is typically used, the engine may have other issues that prevent a stable idle, and the driver may need to keep it running by manually modulating the throttle. A short timed limit still provides the desired protection without risking the engine remaining unnecessarily RPM-limited.

If you think this direction makes sense, I’d be happy to implement these two modes. I’d prefer to get your feedback on the overall approach and the behavioural details first, then I can rework the PR accordingly.


Edit: forgot to mention that in "driver mode" ignition advance should follow cranking settings and not consider tps

adbancroft:
See also #1072

PR #1573, opened 2026-07-25
Fix boundary/overflow bugs in serial & CAN input parsing

### LINKED ISSUE #1555: Tautological CAN start-byte check reads uninitialized memory in comms_secondary.cpp
### Summary
In `secondserial_Command()`'s `'G'` handler, the check meant to guard against a 2-byte CAN value starting at the last valid byte is tautological — it always evaluates `true` — so the guard never fires. This lets `Ghigh` be read from an uninitialized stack byte.

### Location
`speeduino/comms_secondary.cpp:79-105` (case `'G'`), specifically the check at line 93.

```cpp
uint8_t Gdata[9];
uint8_t Glow, Ghigh;

for (byte Gx = 0; Gx < 8; Gx++) // fills Gdata[0..7]; Gdata[8] is never written
{
  Gdata[Gx] = secondarySerial.read();
}
Glow = Gdata[(configPage9.caninput_source_start_byte[destcaninchannel]&7)];
if ((BIT_CHECK(configPage9.caninput_source_num_bytes,destcaninchannel) > 0))  //if true then num bytes is 2
{
  if ((configPage9.caninput_source_start_byte[destcaninchannel]&7) < 8)   //you can't have a 2 byte value starting at byte 7(8 on the list)
  {
    Ghigh = Gdata[((configPage9.caninput_source_start_byte[destcaninchannel]&7)+1)];
  }
  else { Ghigh = 0; }
}
```

### Details
`caninput_source_start_byte[...] & 7` is a value masked to the range 0–7, so the condition `(x & 7) < 8` is **always true** — it can never take the `else` branch. The trailing comment ("you can't have a 2 byte value starting at byte 7") shows the intent was to reject `start_byte == 7` for 2-byte values, which requires `< 7`, not `< 8`.

Because of this, when a TS-configured CAN input channel has `start_byte == 7` and is marked as a 2-byte value, the code executes:
```cpp
Ghigh = Gdata[((7)+1)]; // Gdata[8]
```
`Gdata` is declared `uint8_t Gdata[9]` but the preceding loop only fills indices `0..7` — `Gdata[8]` is never assigned. This reads an uninitialized stack byte and stores it into `currentStatus.canin[destcaninchannel]`, which is subsequently used for things like VSS speed calculation (`sensors.cpp`) and exposed over the logging/TunerStudio interface.

### Impact
Uninitialized-memory read reachable via a normal (if unusual) CAN-input configuration — no malicious input required, just `start_byte = 7` + 2-byte width on any of the 16 CAN input channels (`configPage9.caninput_source_start_byte[16]`). Produces a non-deterministic high byte in `currentStatus.canin[]`.

### Suggested fix
Change the comparison to match the comment's intent:
```cpp
if ((configPage9.caninput_source_start_byte[destcaninchannel]&7) < 7)
```

### Found via

### LINKED ISSUE #1556: Unvalidated CAN channel byte indexes fixed-size arrays in comms_secondary.cpp (OOB read/write)
### Summary
`secondserial_Command()`'s `'G'` handler uses a raw byte read directly off the secondary serial line to index two 16-element arrays, with no bounds check. A value of 8–255 causes an out-of-bounds read from config memory and an out-of-bounds write past `currentStatus.canin[16]`.

### Location
`speeduino/comms_secondary.cpp:73-105` (case `'G'`)

```cpp
case 'G': // this is the reply command sent by the Can interface
  ...
  destcaninchannel = secondarySerial.read();  // the input channel that requested the data value
  if (cmdSuccessful != 0)
  {
    ...
    Glow = Gdata[(configPage9.caninput_source_start_byte[destcaninchannel]&7)];
    ...
    currentStatus.canin[destcaninchannel] = (Ghigh<<8) | Glow;
  }
```

### Details
- `destcaninchannel` is declared `byte` and assigned directly from `secondarySerial.read()` — any value `0..255`, entirely controlled by whatever device is attached to the secondary serial port (a CAN-interface adapter such as a TouCAN, or anything else sending bytes on that UART).
- It is used unchecked to index `configPage9.caninput_source_start_byte[16]` (defined in `config_pages.h:585`) and `currentStatus.canin[16]` (defined in `statuses.h:210`).
- For `destcaninchannel` in `8..255`, the read from `caninput_source_start_byte[]` pulls adjacent config-page memory, and the write to `canin[destcaninchannel]` (a `uint16_t`) writes 2 bytes somewhere past the end of the `canin` array, into whatever field of `currentStatus` (or beyond) happens to follow it in memory — corrupting live engine-state data.

### Impact
A misbehaving or noisy secondary-serial CAN interface (this is a physical UART link to an external adapter, not internet-facing, but is genuinely a separate, less-trusted device on the bus) can corrupt arbitrary adjacent memory in the `statuses` struct — which holds live RPM, fueling, and ignition state.

### Suggested fix
Reject the command if the channel is out of range, mirroring the pattern already used for CAN input channels elsewhere in the codebase:
```cpp
destcaninchannel = secondarySerial.read();
if (destcaninchannel < 16U)
{
  ... existing body ...
}
```

### Found via

### LINKED ISSUE #1557: 'p' page-read command: wire-controlled length overflows serialPayload buffer
### Summary
The `'p'` (page read) serial command derives a buffer length directly from the wire with no check against the actual size of the destination buffer, allowing a crafted command to overflow the global `serialPayload` array.

### Location
`speeduino/comms.cpp:337-351` (`loadPageValuesToBuffer`) and its call site around line 740-745.

```cpp
static void loadPageValuesToBuffer(uint8_t pageNum, uint16_t offset, byte *buffer, uint16_t length)
{
  for(uint16_t i = 0; i < length; i++)
  {
    buffer[i] = getPageValue(pageNum, offset + i);
  }
}
```
```cpp
uint16_t length = word(serialPayload[6], serialPayload[5]);
...
loadPageValuesToBuffer(serialPayload[2], word(serialPayload[4], serialPayload[3]), &serialPayload[1], length);
```

### Details
`length` is a `uint16_t` (0–65535) built directly from two attacker-controlled bytes on the wire (`serialPayload[5]`/`serialPayload[6]`). `loadPageValuesToBuffer` writes `buffer[i]` for `i` in `[0, length)` with **no check that `length` fits inside the destination buffer**.

The destination is `&serialPayload[1]`, and `serialPayload` is declared as:
```cpp
// comms.cpp
#define TS_SERIAL_BUFFER_SIZE (2048 + 3)
static uint8_t serialPayload[TS_SERIAL_BUFFER_SIZE];
```
i.e. 2051 bytes. Any `length` greater than ~2050 overflows this global array. The follow-up send (`sendSerialPayloadNonBlocking(length+1)`, further down the same handler) then also reads `length+1` bytes back out of `serialPayload`, so an oversized length additionally leaks whatever adjacent RAM follows the buffer back over serial.

### Impact
Global buffer overflow (write) and adjacent-RAM disclosure (read), triggerable by a single malformed `'p'` command over the primary or secondary serial link — no authentication exists on this protocol.

### Suggested fix
Validate `length` against the remaining buffer capacity before calling `loadPageValuesToBuffer`, e.g.:
```cpp
uint16_t length = word(serialPayload[6], serialPayload[5]);
if (length <= (sizeof(serialPayload) - 1U))
{
  loadPageValuesToBuffer(serialPayload[2], word(serialPayload[4], serialPayload[3]), &serialPayload[1], length);
  ...
}
else
{
  // reject / send error response
}
```

### Found via

### LINKED ISSUE #1558: SEND_OUTPUT_CHANNELS ('r') command has the same unbounded serialPayload write as 'p'
### Summary
The `SEND_OUTPUT_CHANNELS` (`'r'`) serial command has the same unbounded-write pattern as `'p'` (issue to be cross-referenced): a wire-controlled length drives a write loop into `serialPayload` with no bounds check.

### Location
`speeduino/comms.cpp:359-374` (`generateLiveValues`) and its call site around line 757-766.

```cpp
static void generateLiveValues(uint16_t offset, uint16_t packetLength)
{  
  ...
  serialPayload[0] = SERIAL_RC_OK;
  for(uint16_t x=0; x<packetLength; x++)
  {
    serialPayload[x+1U] = getTSLogEntry(offset+x); 
  }
  ...
}
```
```cpp
uint16_t length = word(serialPayload[6], serialPayload[5]);
...
generateLiveValues(offset, length);
```

### Details
`packetLength` comes from `length = word(serialPayload[6], serialPayload[5])` — a fully wire-controlled `uint16_t` (0–65535) — and is used directly as the loop bound with no check against `sizeof(serialPayload)` (2051 bytes, see `comms.cpp:97`). A crafted `'r'` command with a large length field overflows the global `serialPayload` buffer.

### Impact
Same class as the `'p'`-command overflow: a single malformed serial command can overflow a global buffer, corrupting adjacent globals.

### Suggested fix
Bound `packetLength`/`length` against the buffer's remaining capacity before calling `generateLiveValues`, consistent with whatever check is added for the `'p'` command handler.

### Found via

### LINKED ISSUE #1559: 'M' page-write command: integer overflow bypasses page bounds check
### Summary
The `'M'` (burn/write page) serial command's bounds check can be bypassed by 16-bit integer overflow: `offset + length` is computed at a width where it can wrap around, letting a crafted pair of values pass a check that was meant to keep the write inside the page.

### Location
`speeduino/comms.cpp:312-335` (`updatePageValues`) and its call site around line 699.

```cpp
static bool updatePageValues(uint8_t pageNum, uint16_t offset, const byte *buffer, uint16_t length)
{
  if ( (offset + length) <= getPageSize(pageNum) )
  {
    for(uint16_t i = 0; i < length; i++)
    {
      setPageValue(pageNum, (offset + i), buffer[i]);
    }
    ...
    return true;
  }
  return false;
}
```
```cpp
if (updatePageValues(serialPayload[2], word(serialPayload[4], serialPayload[3]), &serialPayload[7], word(serialPayload[6], serialPayload[5])))
```

### Details
Both `offset` and `length` are `uint16_t` values built directly from wire bytes (up to 65535 each). Per the C++ usual-arithmetic-conversion rules, adding two `uint16_t` values on a platform where `int` is 16 bits (e.g. AVR/avr-gcc) promotes them to `unsigned int`, which is *also* 16 bits there — so `offset + length` is computed modulo 65536.

This means an attacker/caller can choose `offset` and `length` such that the true sum is large (e.g. `offset = 65500`, `length = 100`, true sum 65600) but the wrapped sum (`65600 - 65536 = 64`) passes the `<= getPageSize(pageNum)` check for any page whose size is `>= 64`. The subsequent loop then calls `setPageValue(pageNum, offset + i, buffer[i])` with the *original*, out-of-range `offset` (e.g. `65500 + i`), which is far larger than any real page's size — an out-of-bounds write into whatever `getPageValue`/`setPageValue`'s internal addressing indexes into.

### Impact
Integer-overflow bypass of a bounds check (CWE-190), reachable via a single crafted `'M'` command, resulting in an out-of-bounds config-memory write.

### Suggested fix
Perform the bounds check in a wider type so it can't wrap:
```cpp
if ( ((uint32_t)offset + (uint32_t)length) <= (uint32_t)getPageSize(pageNum) )
```

### Found via

### LINKED ISSUE #1560: Off-by-one read past CAN payload in readAuxCanBus() (comms_CAN.cpp)
### Summary
`readAuxCanBus()` reads one byte past the end of the incoming CAN message's 8-byte payload when a 2-byte CAN input is configured with `start_byte == 7`.

### Location
`speeduino/comms_CAN.cpp:779-807`

```cpp
void readAuxCanBus()
{
  for (int i = 0; i < 16; i++)
  {
    uint16_t channelAddress = (configPage9.caninput_source_can_address[i] + TS_CAN_OFFSET);
    if (inMsg.id == channelAddress )
    {
      if (!BIT_CHECK(configPage9.caninput_source_num_bytes, i))
      {
        currentStatus.canin[i] = inMsg.buf[configPage9.caninput_source_start_byte[i]];
      }
      else
      {
        if (configPage9.caninputEndianess == 1)
        {
          currentStatus.canin[i] = ((inMsg.buf[configPage9.caninput_source_start_byte[i]]) | (inMsg.buf[configPage9.caninput_source_start_byte[i] + 1] << 8));
        }
        else
        {
          currentStatus.canin[i] = ((inMsg.buf[configPage9.caninput_source_start_byte[i]] << 8) | (inMsg.buf[configPage9.caninput_source_start_byte[i] + 1]));
        }
      }
    }
  } 
}
```

### Details
`configPage9.caninput_source_start_byte[16]` is documented as holding "the start byte number (value of 0-7)" (`config_pages.h:585`). `inMsg.buf` is a fixed 8-byte CAN payload (`CAN_message_t::buf[8]`, `src/STM32_CAN/STM32_CAN.h:34`).

For a 2-byte input (`caninput_source_num_bytes` bit set), the code always reads `buf[start_byte + 1]` with no check that `start_byte <= 6`. If a user configures `start_byte == 7` for a 2-byte input (a value the field's own documented range explicitly allows), `buf[start_byte + 1]` becomes `buf[8]` — one past the end of the array.

This is the same inclusive/exclusive boundary mistake as the recently-fixed stepper-motor hysteresis bug in `idle.cpp` (PR #1289): a bound that should exclude the last valid value doesn't.

### Impact
Out-of-bounds struct read reachable via ordinary (if edge-case) CAN-input configuration through TunerStudio; the value stored into `currentStatus.canin[i]` becomes an adjacent field's byte instead of the intended CAN data.

### Suggested fix
Guard the 2-byte path against `start_byte == 7`, e.g. treat it as a configuration error and skip the update, or clamp/ignore the high byte:
```cpp
if (configPage9.caninput_source_start_byte[i] <= 6U)
{
  // existing 2-byte read
}
```

### Found via

bahram-azimbeyk:
## Summary
Found while doing a static review of the core firmware. Fixes six related boundary/overflow bugs in the serial and CAN input parsing code:

- `comms_secondary.cpp`: the `'G'` handler's start-byte check was tautological (`(x&7)<8` is always true), letting a `start_byte==7` 2-byte read pull an uninitialized `Gdata[8]` byte into `currentStatus.canin`. Also, `destcaninchannel` came straight off the wire with no bounds check before indexing `caninput_source_start_byte[16]`/`canin[16]`.
- `comms_CAN.cpp`: `readAuxCanBus()` had the same off-by-one — a 2-byte CAN input configured with `start_byte==7` read `inMsg.buf[8]`, one past the 8-byte CAN payload.
- `comms.cpp`: `updatePageValues()` (`'M'` command) checked `(offset+length)<=getPageSize()` using 16-bit arithmetic, which can wrap and bypass the check; widened to 32-bit. `loadPageValuesToBuffer` (`'p'` command) and `generateLiveValues` (`SEND_OUTPUT_CHANNELS`) took a wire-controlled length with no check against `serialPayload`'s actual size, allowing a crafted length to overflow the global buffer; both now reject out-of-range lengths with `SERIAL_RC_RANGE_ERR`.

Each of these is filed as its own issue with full detail/repro reasoning.

Fixes #1555, Fixes #1556, Fixes #1557, Fixes #1558, Fixes #1559, Fixes #1560

## Test plan
- [ ] CI build/test suite
- Manually traced each fix against the current call sites; none of the changes alter behavior for well-formed/in-range input, only reject previously-unchecked out-of-range input.
- No local hardware available to me to test against a real TunerStudio/CAN session — would appreciate a maintainer sanity-check on the `'p'`/`'r'`/`'M'` serial protocol paths in particular.

---

### Note on `codecov/patch`

This check is red (0% patch coverage) and I don't think it can be reasonably satisfied here — flagging for maintainer discretion rather than leaving it unexplained:

- The changed lines are **defensive bounds-guards on serial/CAN input parsing** across `comms.cpp`, `comms_secondary.cpp` and `comms_CAN.cpp`. They are the fix itself, so (unlike a formatting-only change) they can't be reduced to a non-coverable diff.
- Exercising them requires driving `secondserial_Command()` / `readAuxCanBus()` / the `'p'`/`'r'`/`'M'` command handlers, which depend on `secondarySerial`, the CAN `inMsg` struct and the static `serialPayload` buffer. There is currently **no `test_comms` suite / serial+CAN mocking harness** in the repository, and codecov's 90% patch target would require covering ~all of these hardware-dependent branches, not just a subset.
- Every functional check on this PR (all-board compiles, unit tests, SonarCloud, `codecov/project`) is green; only the new-code `codecov/patch` gate is affected.

Happy to add a `test_comms` harness in a follow-up if the maintainers would like the coverage, or to adjust the approach.

### DIFF
--- a/speeduino/comms.cpp
+++ b/speeduino/comms.cpp
@@ -329,7 +329,7 @@ static void sendReturnCodeMsg(byte returnCode)
  */
 static bool updatePageValues(uint8_t pageNum, uint16_t offset, const byte *buffer, uint16_t length)
 {
-  if ( (offset + length) <= getPageSize(pageNum) )
+  if ( ((uint32_t)offset + (uint32_t)length) <= (uint32_t)getPageSize(pageNum) )
   {
     for(uint16_t i = 0; i < length; i++)
     {
@@ -747,10 +747,17 @@ void processSerialCommand(void)
       //2 - Length
       uint16_t length = word(serialPayload[6], serialPayload[5]);
 
-      //Setup the transmit buffer
-      serialPayload[0] = SERIAL_RC_OK;
-      loadPageValuesToBuffer(serialPayload[2], word(serialPayload[4], serialPayload[3]), &serialPayload[1], length);
-      sendSerialPayloadNonBlocking(length + 1U);
+      if (length <= (sizeof(serialPayload) - 1U))
+      {
+        //Setup the transmit buffer
+        serialPayload[0] = SERIAL_RC_OK;
+        loadPageValuesToBuffer(serialPayload[2], word(serialPayload[4], serialPayload[3]), &serialPayload[1], length);
+        sendSerialPayloadNonBlocking(length + 1U);
+      }
+      else
+      {
+        sendReturnCodeMsg(SERIAL_RC_RANGE_ERR);
+      }
       break;
     }
 
@@ -771,8 +778,15 @@ void processSerialCommand(void)
 
       if(cmd == SEND_OUTPUT_CHANNELS) //Send output channels command 0x30 is 48dec
       {
-        generateLiveValues(offset, length);
-        sendSerialPayloadNonBlocking(length + 1U);
+        if (length <= (sizeof(serialPayload) - 1U))
+        {
+          generateLiveValues(offset, length);
+          sendSerialPayloadNonBlocking(length + 1U);
+        }
+        else
+        {
+          sendReturnCodeMsg(SERIAL_RC_RANGE_ERR);
+        }
       }
       else if(cmd == 0x0fU)
       {
--- a/speeduino/comms.cpp
+++ b/speeduino/comms.cpp
@@ -314,8 +314,19 @@ static void sendReturnCodeMsg(byte returnCode)
 
 // ====================================== Command/Action Support =============================
 
-// The functions in this section are abstracted out to prevent enlarging callers stack frame, 
+// The functions in this section are abstracted out to prevent enlarging callers stack frame,
 // which in turn throws off free ram reporting.
+
+/** @brief Add two 16-bit values in 32-bit arithmetic so the sum cannot wrap.
+ *
+ * On AVR `int` is 16-bit, so `a + b` for two uint16_t operands is evaluated at
+ * 16 bits and can overflow. Widening to uint32_t first guarantees an exact sum,
+ * which matters when the result is used in a bounds check against attacker-supplied
+ * offset/length values.
+ */
+static inline uint32_t addWithoutOverflow(uint16_t a, uint16_t b) {
+  return (uint32_t)a + (uint32_t)b;
+}
 
 /**
  * @brief Update a pages contents from a buffer
@@ -329,7 +340,7 @@ static void sendReturnCodeMsg(byte returnCode)
  */
 static bool updatePageValues(uint8_t pageNum, uint16_t offset, const byte *buffer, uint16_t length)
 {
-  if ( ((uint32_t)offset + (uint32_t)length) <= (uint32_t)getPageSize(pageNum) )
+  if ( addWithoutOverflow(offset, length) <= getPageSize(pageNum) )
   {
     for(uint16_t i = 0; i < length; i++)
     {
@@ -747,7 +758,7 @@ void processSerialCommand(void)
       //2 - Length
       uint16_t length = word(serialPayload[6], serialPayload[5]);
 
-      if (length <= (sizeof(serialPayload) - 1U))
+      if (length < _countof(serialPayload))
       {
         //Setup the transmit buffer
         serialPayload[0] = SERIAL_RC_OK;
@@ -778,7 +789,7 @@ void processSerialCommand(void)
 
       if(cmd == SEND_OUTPUT_CHANNELS) //Send output channels command 0x30 is 48dec
       {
-        if (length <= (sizeof(serialPayload) - 1U))
+        if (length < _countof(serialPayload))
         {
           generateLiveValues(offset, length);
           sendSerialPayloadNonBlocking(length + 1U);
--- a/speeduino/comms_secondary.cpp
+++ b/speeduino/comms_secondary.cpp
@@ -29,11 +29,45 @@ sendcancommand is called when a command is to be sent either to serial3
 uint8_t currentSecondaryCommand;
 SECONDARY_SERIAL_T* pSecondarySerial;
 
+// A classic CAN 2.0 data frame carries at most 8 data bytes.
+static constexpr uint8_t CAN_FRAME_DATA_BYTES = 8U;
+
 #if defined(CORE_AVR)
 #pragma GCC push_options
 // This minimizes RAM usage at no performance cost
-#pragma GCC optimize ("Os") 
+#pragma GCC optimize ("Os")
 #endif
+
+// Handle a 'G' reply from the CAN interface: <success><channel><8 data bytes>.
+// Assumes the caller has confirmed enough bytes are available to read.
+static void processSecondaryCanReply(void)
+{
+  const uint8_t cmdSuccessful = secondarySerial.read();      // 0 == fail, 1 == good
+  const uint8_t destcaninchannel = secondarySerial.read();   // the input channel that requested the data value
+
+  // Nothing further is sent for a failed request.
+  if (cmdSuccessful == 0U) { return; }
+
+  uint8_t Gdata[CAN_FRAME_DATA_BYTES + 1U] = { 0U };
+  for (uint8_t Gx = 0U; Gx < CAN_FRAME_DATA_BYTES; Gx++)
+  {
+    Gdata[Gx] = secondarySerial.read();
+  }
+
+  // Ignore out-of-range channels: destcaninchannel indexes both
+  // caninput_source_start_byte[] and canin[].
+  if (destcaninchannel >= _countof(currentStatus.canin)) { return; }
+
+  // Mask to a valid data-byte index (0..CAN_FRAME_DATA_BYTES-1)
+  const uint8_t startByte = configPage9.caninput_source_start_byte[destcaninchannel] & (CAN_FRAME_DATA_BYTES - 1U);
+  uint8_t Ghigh = 0U;
+  // A 2-byte value needs startByte and startByte+1, so it can't start at the last data byte.
+  if (BIT_CHECK(configPage9.caninput_source_num_bytes, destcaninchannel) && (startByte < (CAN_FRAME_DATA_BYTES - 1U)))
+  {
+    Ghigh = Gdata[startByte + 1U];
+  }
+  currentStatus.canin[destcaninchannel] = ((uint16_t)Ghigh << 8) | Gdata[startByte];
+}
 
 void secondserial_Command(void)
 {
@@ -72,43 +106,10 @@ void secondserial_Command(void)
 
     case 'G': // this is the reply command sent by the Can interface
       serialSecondaryStatusFlag = SERIAL_COMMAND_INPROGRESS_LEGACY;
-      byte destcaninchannel;
       if (secondarySerial.available() >= 9)
       {
         serialSecondaryStatusFlag = SERIAL_INACTIVE;
-        uint8_t cmdSuccessful = secondarySerial.read();        //0 == fail,  1 == good.
-        destcaninchannel = secondarySerial.read();  // the input channel that requested the data value
-        if (cmdSuccessful != 0)
-        {                                 // read all 8 bytes of data.
-          uint8_t Gdata[9];
-          uint8_t Glow, Ghigh;
-
-          for (byte Gx = 0; Gx < 8; Gx++) // first two are the can address the data is from. next two are the can address the data is for.then next 1 or two bytes of data
-          {
-            Gdata[Gx] = secondarySerial.read();
-          }
-        if (destcaninchannel < 16U) // destcaninchannel indexes caninput_source_start_byte[16] / canin[16]; discard out-of-range channels
-        {
-          Glow = Gdata[(configPage9.caninput_source_start_byte[destcaninchannel]&7)];
-          if ((BIT_CHECK(configPage9.caninput_source_num_bytes,destcaninchannel) > 0))  //if true then num bytes is 2
-          {
-            if ((configPage9.caninput_source_start_byte[destcaninchannel]&7) < 7)   //you can't have a 2 byte value starting at byte 7(8 on the list)
-            {
-              Ghigh = Gdata[((configPage9.caninput_source_start_byte[destcaninchannel]&7)+1)];
-            }
-            else { Ghigh = 0; }
-          }
-        else
-        {
-          Ghigh = 0;
-        }
-
-        currentStatus.canin[destcaninchannel] = (Ghigh<<8) | Glow;
-      }
-      }
-
-        else{}  //continue as command request failed and/or data/device was not available
-
+        processSecondaryCanReply();
       }
       break;
 
--- a/speeduino/comms_secondary.cpp
+++ b/speeduino/comms_secondary.cpp
@@ -87,10 +87,12 @@ void secondserial_Command(void)
           {
             Gdata[Gx] = secondarySerial.read();
           }
+        if (destcaninchannel < 16U) // destcaninchannel indexes caninput_source_start_byte[16] / canin[16]; discard out-of-range channels
+        {
           Glow = Gdata[(configPage9.caninput_source_start_byte[destcaninchannel]&7)];
           if ((BIT_CHECK(configPage9.caninput_source_num_bytes,destcaninchannel) > 0))  //if true then num bytes is 2
           {
-            if ((configPage9.caninput_source_start_byte[destcaninchannel]&7) < 8)   //you can't have a 2 byte value starting at byte 7(8 on the list)
+            if ((configPage9.caninput_source_start_byte[destcaninchannel]&7) < 7)   //you can't have a 2 byte value starting at byte 7(8 on the list)
             {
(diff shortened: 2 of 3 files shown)


speeduino/comms_secondary.cpp
```diff
@@ -87,10 +87,12 @@ void secondserial_Command(void)
           {
             Gdata[Gx] = secondarySerial.read();
           }
+        if (destcaninchannel < 16U) // destcaninchannel indexes caninput_source_start_byte[16] / canin[16]; discard out-of-range channels
```
adbancroft: Can we use _countof() instead of a hard coded magic number?

speeduino/comms_secondary.cpp
```diff
@@ -87,10 +87,12 @@ void secondserial_Command(void)
...
+        {
           Glow = Gdata[(configPage9.caninput_source_start_byte[destcaninchannel]&7)];
           if ((BIT_CHECK(configPage9.caninput_source_num_bytes,destcaninchannel) > 0))  //if true then num bytes is 2
           {
-            if ((configPage9.caninput_source_start_byte[destcaninchannel]&7) < 8)   //you can't have a 2 byte value starting at byte 7(8 on the list)
```
adbancroft, line 93: Can we use something other than a hard coded magic number?

speeduino/comms.cpp
```diff
@@ -771,8 +778,15 @@ void processSerialCommand(void)
 
       if(cmd == SEND_OUTPUT_CHANNELS) //Send output channels command 0x30 is 48dec
       {
-        generateLiveValues(offset, length);
-        sendSerialPayloadNonBlocking(length + 1U);
+        if (length <= (sizeof(serialPayload) - 1U))
```
adbancroft: Can we use _countof() instead of sizeof()?

speeduino/comms.cpp
```diff
@@ -747,10 +747,17 @@ void processSerialCommand(void)
       //2 - Length
       uint16_t length = word(serialPayload[6], serialPayload[5]);
 
-      //Setup the transmit buffer
-      serialPayload[0] = SERIAL_RC_OK;
-      loadPageValuesToBuffer(serialPayload[2], word(serialPayload[4], serialPayload[3]), &serialPayload[1], length);
-      sendSerialPayloadNonBlocking(length + 1U);
+      if (length <= (sizeof(serialPayload) - 1U))
+      {
```
adbancroft, line 762: Can we use _countof() instead of sizeof()?

speeduino/comms.cpp
```diff
@@ -329,7 +329,7 @@ static void sendReturnCodeMsg(byte returnCode)
  */
 static bool updatePageValues(uint8_t pageNum, uint16_t offset, const byte *buffer, uint16_t length)
 {
-  if ( (offset + length) <= getPageSize(pageNum) )
+  if ( ((uint32_t)offset + (uint32_t)length) <= (uint32_t)getPageSize(pageNum) )
   {
```
adbancroft, line 344: Can you add a comment on why the casts are required? Better yet, encapsulate into a function (E.g. uint32_t addOverflowSafe(uint16_t, uint16_t)?

adbancroft (changes requested):

bahram-azimbeyk:
Thanks for the review @adbancroft — all addressed in the latest push:

- **`comms_secondary.cpp`** – the reply handler now bounds `destcaninchannel` against `_countof(currentStatus.canin)`, and I introduced `CAN_FRAME_DATA_BYTES` (8) so the data-byte mask and the "a 2-byte value can't start at the last byte" check derive from that constant instead of literal `7`/`8`. Also tidied the stray brace nesting in that block.
- **`comms.cpp`** – the `'p'`/`'r'` length bounds now use `_countof(serialPayload)`, and the widening overflow-safe add is encapsulated in `addWithoutOverflow(uint16_t, uint16_t)` with a comment explaining why 16-bit `offset + length` needs widening.
- **`comms_CAN.cpp`** – expressed the same 2-byte start-byte guard as `_countof(inMsg.buf) - 1U` for consistency.

speeduino/comms.cpp
```diff
@@ -747,10 +758,17 @@ void processSerialCommand(void)
       //2 - Length
       uint16_t length = word(serialPayload[6], serialPayload[5]);
 
-      //Setup the transmit buffer
-      serialPayload[0] = SERIAL_RC_OK;
-      loadPageValuesToBuffer(serialPayload[2], word(serialPayload[4], serialPayload[3]), &serialPayload[1], length);
-      sendSerialPayloadNonBlocking(length + 1U);
+      if (length <= (_countof(serialPayload) - 1U))
+      {
```
adbancroft, line 762: What's the thinking behind '<=' instead of the more traditional `<'? I.e. 
```
if (length < _countof(serialPayload) 
```

speeduino/comms.cpp
```diff
@@ -771,8 +789,15 @@ void processSerialCommand(void)
 
       if(cmd == SEND_OUTPUT_CHANNELS) //Send output channels command 0x30 is 48dec
       {
-        generateLiveValues(offset, length);
-        sendSerialPayloadNonBlocking(length + 1U);
+        if (length <= (_countof(serialPayload) - 1U))
```
adbancroft: Same as above. Why not `if (length < _countof(serialPayload))`?

bahram-azimbeyk:
Thanks for the review! Addressed in the latest push:
- `comms_secondary.cpp`: index against `_countof(currentStatus.canin)` and introduced `CAN_FRAME_DATA_BYTES` so the data-byte mask / 2-byte-start guard are no longer literal 7s. Also extracted the reply parsing into `processSecondaryCanReply()` to flatten the nesting SonarCloud flagged.
- `comms.cpp`: `_countof(serialPayload)` for the 'p'/'r' length checks (switched to the cleaner `length < _countof(serialPayload)` per your note), and encapsulated the widening add in `addWithoutOverflow()` with a comment.
- `comms_CAN.cpp`: same 2-byte guard expressed as `_countof(inMsg.buf) - 1`.

Note the `codecov/patch` check remains red — these are defensive guards on the serial/CAN paths and there's no `test_comms` harness in the repo to exercise them; happy to add one in a follow-up if you'd like the coverage.

PR #1574, opened 2026-07-25
Fix off-by-one bounds checks in Programmable I/O rule-reuse indexing

### LINKED ISSUE #1561: Off-by-one bounds checks in Programmable I/O rule-reuse indexing (programmableIOControl.cpp)
### Summary
Two off-by-one bounds checks in `checkProgrammableIO()`'s rule-reuse indexing admit an out-of-range bit index into the 8-bit `currentRuleStatus` bitfield (valid range 0–7).

### Location
`speeduino/programmableIOControl.cpp:48-105`

```cpp
dataRequested = page13.firstDataIn[y];
if ( dataRequested > 239U )
{
  dataRequested -= REUSE_RULES;                                    // REUSE_RULES == 240
  if ( dataRequested <= sizeof(page13.outputPin) ) { data = BIT_CHECK(currentRuleStatus, dataRequested); }  // line 64
  else { data = 0; }
}
...
dataRequested = page13.secondDataIn[y];
if ( dataRequested <= (REUSE_RULES + sizeof(page13.outputPin)) ) //Failsafe check                            // line 82
{
  if ( dataRequested > 239U )
  {
    dataRequested -= REUSE_RULES;
    data = BIT_CHECK(currentRuleStatus, dataRequested);
  }
  ...
}
```

### Details
`page13.outputPin` has 8 elements (`config_pages.h:853`), and `currentRuleStatus` is an 8-bit variable — valid "reuse rule" indices are `0..7`. Both checks use `<=` where `<` is needed:

- **Line 64**: `dataRequested <= sizeof(page13.outputPin)` (i.e. `<= 8`) admits `dataRequested == 8`. In practice `BIT_CHECK(currentRuleStatus, 8)` shifts the bit position out of the 8-bit value and evaluates to `0` on this codebase's toolchains, so this instance is currently behaviorally inert — but it's still incorrect and fragile (relies on shift-out-of-range behavior rather than an explicit, portable guard).
- **Line 82**: `dataRequested <= (REUSE_RULES + sizeof(page13.outputPin))` (i.e. `<= 248`) admits `dataRequested == 248`, which (after `-= REUSE_RULES`) becomes index `8`. Unlike line 64, this **does** change observable behavior: when this check passes, the whole bitwise-combination block executes and `secondCheck` is computed from a comparator against `data = 0`; when it correctly fails, `secondCheck` stays at its initialized value of `false`. For comparator settings such as `COMPARATOR_NOT_EQUAL` with a non-zero target, these two paths produce different results — a real, reachable behavioral bug for `secondDataIn[y] == 248`.

### Impact
A Programmable I/O rule configured (via TunerStudio) to reuse "rule 9" (`dataRequested == 248` on the second input) can silently evaluate differently than intended, depending on comparator/target settings.

### Suggested fix
```cpp
if ( dataRequested < sizeof(page13.outputPin) ) { data = BIT_CHECK(currentRuleStatus, dataRequested); }
else { data = 0; }
...
if ( dataRequested < (REUSE_RULES + sizeof(page13.outputPin)) ) //Failsafe check
```

### Found via

bahram-azimbeyk:
## Summary
`checkProgrammableIO()` used `<=` where `<` was needed when validating a reused-rule index against `page13.outputPin`'s 8 elements / the 8-bit `currentRuleStatus` bitfield, admitting index 8 (one past the valid 0-7 range). The second check (`secondDataIn` path) has an observable effect: comparator/target combinations at the boundary value (`secondDataIn[y] == 248`) produce different results depending on whether the invalid index is let through or rejected.

No existing test in `test_programmableio` exercises this exact boundary (tests use rule indices 0, 2-4, and an out-of-bounds value of 10; not 8), so this shouldn't affect existing coverage.

Fixes #1561

## Test plan
- [ ] CI build/test suite
- Manually traced both call sites against `test/test_programmableio/test_programmableIOControl.cpp` to confirm no existing test asserts on the old (buggy) boundary behavior.

### DIFF
--- a/speeduino/programmableIOControl.cpp
+++ b/speeduino/programmableIOControl.cpp
@@ -61,7 +61,7 @@ TESTABLE_STATIC void checkProgrammableIO(statuses& current, const config13& page
       if ( dataRequested > 239U ) //Somehow using 239 uses 9 bytes of RAM, why??
       {
         dataRequested -= REUSE_RULES;
-        if ( dataRequested <= sizeof(page13.outputPin) ) { data = BIT_CHECK(currentRuleStatus, dataRequested); }
+        if ( dataRequested < sizeof(page13.outputPin) ) { data = BIT_CHECK(currentRuleStatus, dataRequested); }
         else { data = 0; }
       }
       else { data = getData(dataRequested); }
@@ -79,7 +79,7 @@ TESTABLE_STATIC void checkProgrammableIO(statuses& current, const config13& page
       if (page13.operation[y].bitwise != BITWISE_DISABLED)
       {
         dataRequested = page13.secondDataIn[y];
-        if ( dataRequested <= (REUSE_RULES + sizeof(page13.outputPin)) ) //Failsafe check
+        if ( dataRequested < (REUSE_RULES + sizeof(page13.outputPin)) ) //Failsafe check
         {
           if ( dataRequested > 239U ) //Somehow using 239 uses 9 bytes of RAM, why??
           {


adbancroft:
Thanks for finding & documenting this. I've fixed it in #1467

PR #1579, opened 2026-07-25
Avoid implementation-defined narrowing in correctionLaunch()

### LINKED ISSUE #1571: correctionLaunch() relies on implementation-defined narrowing conversion
### Summary (portability nit, not currently a functional bug)
`correctionLaunch()` accumulates a value that can exceed `int8_t`'s range in an `int8_t` variable, relying on an implementation-defined narrowing conversion (which happens to round-trip correctly today, but isn't guaranteed by the standard the way the code seems to assume).

### Location
`speeduino/corrections.cpp:568-575`

```cpp
TESTABLE_INLINE_STATIC uint8_t correctionLaunch(void)
{
  int8_t correction = (int8_t)NO_FUEL_CORRECTION;   // 100
  if (currentStatus.launchingHard || currentStatus.launchingSoft) {
    correction = correction + configPage6.lnchFuelAdd;
  } 
  return (uint8_t)correction;
}
```
`lnchFuelAdd` is `int8_t` (`config_pages.h:505`, "Note signed"), tune-editable from -40 to +80 (`reference/speeduino.ini`).

### Details
`NO_FUEL_CORRECTION` is 100. With the maximum tune value (`+80`), `correction + lnchFuelAdd` computes `180` as an `int` (both operands promote to `int` first, so the addition itself doesn't overflow). The overflow happens on the *assignment back* into `int8_t correction` — converting `180` into a type whose range tops out at `127` is implementation-defined behavior in C++ (well-defined as two's-complement wraparound as of C++20, and in practice on every compiler this project targets). That wraparound produces `-76`, which is then cast to `uint8_t` on return — a signed-to-unsigned conversion that *is* well-defined and reinterprets the bit pattern back to `180`.

So in practice, on the two's-complement/GCC-family toolchains this project builds for, the function returns the intended `180` — this is **not currently causing wrong behavior**. But it's relying on implementation-defined conversion behavior for a "correct" result rather than computing it directly, which is exactly the class of implicit-conversion issue MISRA-C flags (and the kind the project's "Boards: MISRA fixes" commits targeted elsewhere) — worth tightening up so correctness doesn't depend on conversion semantics.

### Suggested fix
Accumulate in a type wide enough to hold the intermediate result before narrowing, e.g.:
```cpp
uint8_t correctionLaunch(void)
{
  int16_t correction = (int16_t)NO_FUEL_CORRECTION;
  if (currentStatus.launchingHard || currentStatus.launchingSoft) {
    correction += configPage6.lnchFuelAdd;
  }
  return (uint8_t)correction;
}
```

### Found via

bahram-azimbeyk:
## Summary
`correctionLaunch()` accumulated `NO_FUEL_CORRECTION` (100) + `lnchFuelAdd` (tune-editable -40..+80) in an `int8_t`, which can hold a computed value of 180 only via an implementation-defined conversion on assignment (well-defined as two's-complement wraparound on this project's GCC-based toolchains, but not something the standard guarantees generally). Widened the accumulator to `int16_t` so the value is correct by construction rather than by relying on a specific narrowing-conversion behavior; the final `(uint8_t)` cast is now a plain, always-safe conversion since the result is always in `[60,180]`.

Fixes #1571

## Test plan
- [ ] CI build/test suite
- Checked `test/test_fuel/test_corrections.cpp` - existing tests use `lnchFuelAdd` of +/-25 (never near the +/-40/+80 extremes that trigger the original narrowing), and produce identical results (100, 125, 75) under this change.

### DIFF
--- a/speeduino/corrections.cpp
+++ b/speeduino/corrections.cpp
@@ -568,6 +568,9 @@ This simple check applies the extra fuel if we're currently launching
 */
 TESTABLE_INLINE_STATIC uint8_t correctionLaunch(void)
 {
+  // Accumulate in int16_t: NO_FUEL_CORRECTION (100) + lnchFuelAdd (tune range
+  // -40..+80) can reach 180, which overflows int8_t. The result is always within
+  // [60, 180], so the final narrowing to uint8_t is safe.
   int16_t correction = (int16_t)NO_FUEL_CORRECTION;
   if (currentStatus.launchingHard || currentStatus.launchingSoft) {
     correction = correction + configPage6.lnchFuelAdd;
--- a/speeduino/corrections.cpp
+++ b/speeduino/corrections.cpp
@@ -568,10 +568,10 @@ This simple check applies the extra fuel if we're currently launching
 */
 TESTABLE_INLINE_STATIC uint8_t correctionLaunch(void)
 {
-  int8_t correction = (int8_t)NO_FUEL_CORRECTION;
+  int16_t correction = (int16_t)NO_FUEL_CORRECTION;
   if (currentStatus.launchingHard || currentStatus.launchingSoft) {
     correction = correction + configPage6.lnchFuelAdd;
-  } 
+  }
   return (uint8_t)correction;
 }
 
--- a/test/test_fuel/test_corrections.cpp
+++ b/test/test_fuel/test_corrections.cpp
@@ -746,6 +746,26 @@ static void test_corrections_launch_removeFuel(void) {
   TEST_ASSERT_EQUAL(75U, correctionLaunch() );
 }
 
+static void test_corrections_launch_maxAdd(void) {
+  //lnchFuelAdd max is +80 (see speeduino.ini). 100 + 80 = 180, which exceeds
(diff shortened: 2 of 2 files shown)


speeduino/corrections.cpp
```diff
@@ -568,10 +568,10 @@ This simple check applies the extra fuel if we're currently launching
 */
 TESTABLE_INLINE_STATIC uint8_t correctionLaunch(void)
 {
-  int8_t correction = (int8_t)NO_FUEL_CORRECTION;
+  int16_t correction = (int16_t)NO_FUEL_CORRECTION;
```
adbancroft, line 574: Please add a comment explaining the use of `int16_t`. Thanks.

adbancroft (changes requested):

bahram-azimbeyk:
Thanks @adbancroft — added a comment above the accumulator explaining that `NO_FUEL_CORRECTION` (100) + `lnchFuelAdd` (tune range −40..+80) can reach 180, which overflows `int8_t`, and that the result stays within [60, 180] so the final `uint8_t` narrowing is safe.

Issue #1629, opened 2026-09-09
Tracking: expand configuration, migration and resource validation


bahram-azimbeyk:
### Extend supported-configuration and migration tests — P2 improvement

Recent five-cylinder fixes show the value of checking complete configuration matrices. Cover supported cylinder counts, two/four-stroke, sequential/semi-sequential, staging, ignition modes and available hardware channels. Extend current table/page/storage tests with realistic old-tune fixtures, upgrade/save/reload comparisons and interrupted-save recovery scenarios. Preserve EEPROM layout and serialized values when refactoring containers.

This is a coverage proposal, not a claim that all combinations or migrations currently fail. Decoder RPM boundary issues [#1612](https://github.com/speeduino/speeduino/issues/1612) and [#1613](https://github.com/speeduino/speeduino/issues/1613) already exist; do not open duplicates.

### Measure stack and timing before further memory reductions — P2 improvement

Retain existing base/PR size comparisons for Mega, Teensy and STM32. Add stack high-water measurements on Mega and worst-case ISR/main-loop timing on a test rig, especially with logging, secondary serial and CAN activity. Map hot paths before changing integer widths or moving state from static storage onto the stack.

Pin dependency versions or commit references where reproducible comparisons require them, and update them deliberately. The current configuration includes unpinned platform/library Git references. Do not interpret toolchain drift as a firmware memory regression.

### Continue controller extraction without duplicating active work — P3 improvement

The remaining idle/sensor/protection/global dependencies are reasonable future refactor targets, but behavior tests should precede substantial movement. Keep ownership and hardware access explicit and avoid virtual/dynamic abstractions merely for style.

Active upstream work already covers boost [#1605](https://github.com/speeduino/speeduino/pull/1605), crank-angle/atomic-copy helpers [#1610](https://github.com/speeduino/speeduino/pull/1610), ignition initialization [#1616](https://github.com/speeduino/speeduino/pull/1616), and fuel-pump control [#1617](https://github.com/speeduino/speeduino/pull/1617).

The maintainer has explicitly said a pin-priority solution is in progress in [#1604](https://github.com/speeduino/speeduino/issues/1604#issuecomment-5398015635). Contribute reproduction cases and tests there rather than competing with that redesign. Broad pin-conflict UI and a second fan stage should be separate feature proposals with defined priority/interlock behavior and measured AVR costs.

### Baseline and scope
[...]

PR #1640, opened 2026-09-09
CI: report occupied memory and preserve build diagnostics

### LINKED ISSUE #1637: PR memory report uses free Teensy bytes and substitutes zero for missing metrics
### Summary
The PR size-report parser treats Teensy **free** RAM/flash as if it were **used** memory. A reduction in free space would therefore produce a negative delta and be marked as an optimization. For other targets, an unrecognized/missing measurement is silently replaced by zero, so missing evidence can appear as a green 0-byte delta.

### Evidence
At master `cc9eb445`, `.github/workflows/pr-memory-deltas.py` matches `free for local variables` and `free for files` in Teensy output, stores those numbers under `ram` / `flash`, then computes `PR - base` with negative considered good.

The generated report on #1634 displays 7,877,636 bytes as Teensy flash and 0-byte STM32 RAM/flash values: https://github.com/speeduino/speeduino/pull/1634#issuecomment-5607530157 . Zero in that report does not establish zero memory usage. The root cause of each missing target measurement should be diagnosed from its build log.

### Expected behavior
- Report used regions rather than free capacity. For Teensy 4.x, define the RAM metric explicitly (RAM1 static occupancy including code/padding plus RAM2 variables) and sum FLASH code/data/headers.
- Keep support for ordinary PlatformIO used-byte reports, including Teensy targets that use that format.
- Represent unavailable/unparseable metrics as N/A, never fabricate zero or a delta against an absent environment.
- Add fixture-based Python tests for both report formats, growth/shrinkage, missing metrics and a missing environment.

### Scope
This is a concrete follow-up to resource-validation item #1629. No firmware behavior or memory allocation needs to change. A parser fix should be a separate PR from firmware optimizations, because otherwise it can reverse the interpretation of their resource deltas.

bahram-azimbeyk:
The memory report treated Teensy free space as occupied memory, omitted RAM2, and replaced missing measurements with zero. Report occupied FLASH and RAM regions, prefer standard PlatformIO used-byte metrics, and distinguish missing measurements from a measured zero.

Keep both build logs and the partial report when a build fails, then fail the job after uploading diagnostics. Compare the tested merge commit against its actual first parent: pull-request metadata can retain an older base SHA after master advances, otherwise attributing unrelated upstream firmware growth to the PR.

Fixes #1637. Refs #1629.

### Scope

Only the memory-report parser, its seven regression tests, and the size build/report workflows change. There are no firmware, calibration-layout, or `platformio.ini` changes. PlatformIO resolves the platform configured by each checkout; this workflow does not preinstall or force a particular STM32 version.

The independently observed fresh-build STM32 core/RTC incompatibility is tracked in #1649 for specialist review. When a completed build fails and has uploaded the report artifact, the reporting workflow also publishes that available partial report with a Build failed banner and a link to the source run, instead of leaving a previous success table in place. This PR preserves diagnostics and reports build failures; it does not fix that dependency conflict or substitute zero measurements for failed builds.

### Validation

- All seven parser tests pass; six failed against the original parser. Coverage includes the supported formats, ANSI escapes, real zero values, incomplete/missing measurements, and delta direction.
- Actual Teensy 4.1 output was checked against occupied-region sums: RAM 278,560 bytes and flash 248,828 bytes.
- Prior live size run [34491323597](https://github.com/speeduino/speeduino/actions/runs/34491323597) validated the actual-merge-parent comparison and reported zero RAM/flash delta on all three boards at revision `aa8554de`. That run included the former STM32 workaround, which is no longer part of this PR; it is not a passing build claim for the latest revision.
- After the review updates: `python -m unittest discover -s .github/workflows/tests -v`, actionlint (external shellcheck/pyflakes disabled locally), and `git diff --check` pass.
- New GitHub checks run on the updated revision. Fresh STM32 builds may still encounter the independently tracked #1649 conflict until the supported dependency combination is agreed.

### Commit structure

Tests, occupied-memory parsing, build-failure propagation, diagnostic retention, and actual merge-base selection are separate commits. The review follow-ups separately remove the workflow-wide STM32 installation and revert the platform pin, leaving both out of the final diff.

### Latest review follow-up

Run [34534504103](https://github.com/speeduino/speeduino/actions/runs/34534504103) failed on the STM32/RTC callback conflict tracked in #1649; the other nine PR checks passed at 6cdeb026. Both `memory-build-logs` and `sketches-report-all` artifacts were uploaded. The existing reporting workflow runs only on successful source builds, explaining why the older all-green table remained visible.

Commit 1b897220 enables reporting for completed success/failure runs and labels reports from failed builds explicitly. All seven parser tests, actionlint for both workflows, and diff checks pass. The exact embedded annotation script was also executed locally: it preserves the original report and prepends the failure status and source-run link.

GitHub loads this `workflow_run` reporting workflow from the default branch, so the changed comment behavior can take effect after merge; it cannot be demonstrated by the PR's own reporter run before then. Missing artifacts from an earlier infrastructure failure are not fabricated. The STM32 build failure remains separate and is not suppressed.

### DIFF
--- a/.github/workflows/pr-memory-deltas.yml
+++ b/.github/workflows/pr-memory-deltas.yml
@@ -45,6 +45,9 @@ jobs:
       run: python -m unittest discover -s "${{ env.PR_FOLDER }}/.github/workflows/tests"
 
     - name: Build PR Envs
+      id: build_pr
+      # Preserve measurements and logs before failing the final build check.
+      continue-on-error: true
       run: |
         set -o pipefail
         source PIO/bin/activate
@@ -53,10 +56,13 @@ jobs:
     - name: Checkout Base Branch
       uses: actions/checkout@v4
       with:
-        ref: master
+        ref: ${{ github.event.pull_request.base.sha }}
         path: ${{ env.BASE_FOLDER }}
 
     - name: Build Base Envs
+      id: build_base
+      # Preserve measurements and logs before failing the final build check.
+      continue-on-error: true
       run: |
         set -o pipefail
         source PIO/bin/activate
@@ -73,11 +79,13 @@ jobs:
         fi
 
     - name: Upload build logs
-      if: ${{ env.ACT }}
+      if: ${{ always() }}
       uses: actions/upload-artifact@v4
       with:
+        name: memory-build-logs
         path: |
-            **/*.log
+          ${{ env.PR_LOG_FILE }}
+          ${{ env.BASE_LOG_FILE }}
 
       # This step is needed to pass the size data to the report job.
     - name: Upload sketches report to workflow artifact
@@ -87,3 +95,9 @@ jobs:
         path: |
           ${{ env.SIZE_REPORT_FILE }}
           ${{ env.PR_NUM_FILE }}
+
+    - name: Check build results
+      if: ${{ always() && (steps.build_pr.outcome == 'failure' || steps.build_base.outcome == 'failure') }}
+      run: |
+        echo "At least one firmware build failed; see memory-build-logs and the partial size report."
+        exit 1
--- a/.github/workflows/pr-memory-deltas.yml
+++ b/.github/workflows/pr-memory-deltas.yml
@@ -30,6 +30,12 @@ jobs:
         source PIO/bin/activate
         pip install platformio
 
+    # Use the same compatible STM32 toolchain for the PR and an unpinned base.
+    - name: Install compatible STM32 platform
+      run: |
+        source PIO/bin/activate
+        platformio pkg install --global --platform "platformio/ststm32@19.5.0"
+
     - name: Build command line
       if: ${{ env.PIO_ENVS }}
       id: build_command_line
--- a/.github/workflows/pr-memory-deltas.yml
+++ b/.github/workflows/pr-memory-deltas.yml
@@ -30,12 +30,6 @@ jobs:
         source PIO/bin/activate
         pip install platformio
 
-    # Use the same compatible STM32 toolchain for the PR and an unpinned base.
-    - name: Install compatible STM32 platform
-      run: |
-        source PIO/bin/activate
-        platformio pkg install --global --platform "platformio/ststm32@19.5.0"
-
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-memory-deltas.yml
```diff
@@ -30,6 +30,12 @@ jobs:
         source PIO/bin/activate
         pip install platformio
 
+    # Use the same compatible STM32 toolchain for the PR and an unpinned base.
+    - name: Install compatible STM32 platform
+      run: |
+        source PIO/bin/activate
+        platformio pkg install --global --platform "platformio/ststm32@19.5.0"
+
```
adbancroft: Is this required? PIO should use the platform in the INI. I'm not concerned about a one time mismatch/failure that will resolve post-merge.

platformio.ini
```diff
@@ -102,7 +102,8 @@ extra_scripts = post:post_extra_script.py
 
 ;STM32 Official core
 [env:black_F407VE]
-platform = ststm32
+; Core 3.0 conflicts with STM32duino RTC 1.2.0 callback typedefs.
+platform = ststm32@19.5.0
```
adbancroft: Has this been verified by one of the STM32 experts? E.g. VitorBoss, Pazi

.github/workflows/pr-memory-deltas.yml
bahram-azimbeyk: Removed in 743198d6. The size workflow now lets PlatformIO resolve the configuration in each checkout, with no global STM32 platform override. The separate core/RTC compatibility evidence is preserved in #1649.

platformio.ini
bahram-azimbeyk: No STM32 specialist or hardware validation was obtained; the evidence was limited to successful compilation with 19.5.0/core 2.12.0 versus the core 3.0.0 callback conflict. I reverted the platform pin in 6cdeb026, so platformio.ini is no longer changed by this PR. #1649 records the logs and the dependency/RTC questions for specialist review.

PR #1642, opened 2026-09-10
Refactor launch and flat-shift condition handling

bahram-azimbeyk:
Launch and flat-shift control currently mix clutch input sampling, engagement-RPM capture, eligibility checks and cut activation in one function. The rolling-cut RPM adjustment is also implemented separately for the two modes. Extract these responsibilities into small internal helpers so the conditions and their shared arithmetic can be reviewed independently.

`updateClutchState()` retains the enabled-only input read, polarity handling, previous-state tracking and rising-edge RPM capture. `isLaunchArmed()` names the launch eligibility conditions. `withinLaunchSpeedLimit()` uses `VSS_MODE_OFF`; its result is combined with `aboveLaunchRpmLimit()` using short-circuit evaluation, while retaining the outer launch/flat-shift priority. `getHardCutRpmLimit()` applies the rolling-cut adjustment for both launch and flat shift, using the existing `RPM_COARSE` and `SIGNED_RPM_MEDIUM` conversions instead of local scale factors. The clutch polarity and flat-shift flag use direct boolean comparisons. A documented final else branch explicitly retains the cleared flags when neither function is armed, addressing Sonar rule cpp:M23_112. The public interface and stored status/configuration layout are unchanged.

The existing behavior is retained:

- Engagement RPM is captured when the clutch is pressed and held until the next engagement.
- Launch requires engagement RPM below the arming threshold and TPS at or above its threshold.
- Equality at the arming RPM belongs to flat shift.
- With VSS enabled, launch requires speed strictly below the configured limit. Disabled VSS imposes no speed restriction.
- Full cut ignores the configured rolling delta; rolling cut applies the signed RPM/10 delta.
- Both cuts require RPM strictly above the resulting limit; flags clear when RPM returns to equality.

Six new characterization tests cover those boundaries and clutch transitions through the public controller function, alongside the four existing tests. Tests and implementation are in separate commits. The test fixture holds its own configuration/status objects and adds no firmware state.

Validation:

- All 10 `test_launchcontrol` native tests pass on upstream before the refactor and on the final implementation (4 injector / 4 ignition channels).
- Mega2560 firmware builds pass before and after, using upstream `d023d25d` and the same local toolchain/settings.
- `git diff --check` passes.

| Mega2560 occupied memory | Before | After | Delta |
| --- | ---: | ---: | ---: |
| RAM | 6,354 bytes | 6,354 bytes | 0 |
| Flash | 187,948 bytes | 187,950 bytes | +2 bytes |

Commands: `platformio test -e native_code_coverage -f test_launchcontrol` and `platformio run -e megaatmega2560`. Local Windows native tests used `-Wa,-mbig-obj -DINJ_CHANNELS=4 -DIGN_CHANNELS=4` through `PLATFORMIO_BUILD_FLAGS`. The AVR size comparison used a separate build directory to avoid interference from native test cleanup. Hardware/engine testing was not performed.

Review follow-ups are separate commits: 95cdef5e simplifies the boolean assignments; 7c41b986 adds the explicit inactive branch. All 10 native tests and the Mega build were rerun successfully on 7c41b986; the memory table above describes that revision. Remote CI/Sonar results for the update are separate from the local checks.

The subsequent review updates use VSS_MODE_OFF (2d2d4492) and extract the speed/RPM predicates (b45820d1). All 10 native tests and the Mega2560 build passed again on b45820d1, with the same memory figures shown above.

### DIFF
--- a/speeduino/src/controllers/launch/launchController.cpp
+++ b/speeduino/src/controllers/launch/launchController.cpp
@@ -1,54 +1,70 @@
 #include "launchController.h"
+#include "../../../units.h"
+
+static void updateClutchState(statuses &current, uint8_t launchPin, const config6 &page6)
+{
+  current.previousClutchTrigger = current.clutchTrigger;
+  // Only read the shared clutch input when a function using it is enabled.
+  if (page6.flatSEnable || page6.launchEnabled)
+  {
+    if (page6.launchHiLo > 0) { current.clutchTrigger = digitalRead(launchPin); }
+    else { current.clutchTrigger = !digitalRead(launchPin); }
+
+    current.clutchTriggerActive = current.clutchTrigger; // TunerStudio indication
+  }
+
+  // Capture RPM on engagement, then retain it while the clutch is held or released.
+  if (current.clutchTrigger && !current.previousClutchTrigger)
+  {
+    current.clutchEngagedRPM = current.RPM;
+  }
+}
+
+static bool isLaunchArmed(const statuses &current, const config6 &page6, const config10 &page10)
+{
+  return page6.launchEnabled
+      && current.clutchTrigger
+      && (current.clutchEngagedRPM < RPM_COARSE.toUser(page6.flatSArm))
+      && (current.TPS >= page10.lnchCtrlTPS);
+}
+
+static uint16_t getHardCutRpmLimit(uint16_t baseRpm, const config2 &page2, const config15 &page15)
+{
+  if (page2.hardCutType == HARD_CUT_ROLLING)
+  {
+    baseRpm += SIGNED_RPM_MEDIUM.toUser(page15.rollingProtRPMDelta[0]);
+  }
+  return baseRpm;
+}
 
 void checkLaunchAndFlatShift(statuses &current, uint8_t launchPin, const config2 &page2, const config6 &page6, const config10 &page10, const config15 &page15)
 {
-  //Check for launching/flat shift (clutch) based on the current and previous clutch states
-  current.previousClutchTrigger = current.clutchTrigger;
-  //Only check for pinLaunch if any function using it is enabled. Else pins might break starting a board
-  if(page6.flatSEnable || page6.launchEnabled)
-  {
-    if(page6.launchHiLo > 0) { current.clutchTrigger = digitalRead(launchPin); }
-    else { current.clutchTrigger = !digitalRead(launchPin); }
+  updateClutchState(current, launchPin, page6);
 
-    current.clutchTriggerActive = current.clutchTrigger; //Stores the value to send to TunerStudio
-  }
-  if(current.clutchTrigger && (current.previousClutchTrigger != current.clutchTrigger) ) { current.clutchEngagedRPM = current.RPM; } //Check whether the clutch has been engaged or disengaged and store the current RPM if so
-
-  //Default flags to off
-  current.launchingHard = false; 
+  current.launchingHard = false;
   current.hardLaunchActive = false;
   current.flatShiftingHard = false;
 
-  if (page6.launchEnabled && current.clutchTrigger && (current.clutchEngagedRPM < ((unsigned int)(page6.flatSArm) * 100)) && (current.TPS >= page10.lnchCtrlTPS) ) 
-  { 
-    //Only enable if VSS is not used or if it is, make sure we're not above the speed limit
-    if( (page2.vssMode == 0) || ((page2.vssMode > 0) && (current.vss < page10.lnchCtrlVss)) )
+  if (isLaunchArmed(current, page6, page10))
+  {
+    // A configured vehicle speed limit applies only to launch control.
+    if ((page2.vssMode == 0) || (current.vss < page10.lnchCtrlVss))
     {
-      //Check whether RPM is above the launch limit
-      uint16_t launchRPMLimit = (page6.lnchHardLim * 100);
-      if( (page2.hardCutType == HARD_CUT_ROLLING) ) { launchRPMLimit += (page15.rollingProtRPMDelta[0] * 10); } //Add the rolling cut delta if enabled (Delta is a negative value)
-
-      if(current.RPM > launchRPMLimit)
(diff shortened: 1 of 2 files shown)


speeduino/src/controllers/launch/launchController.cpp
```diff
@@ -1,54 +1,70 @@
...
 {
-  //Check for launching/flat shift (clutch) based on the current and previous clutch states
   current.previousClutchTrigger = current.clutchTrigger;
-  //Only check for pinLaunch if any function using it is enabled. Else pins might break starting a board
-  if(page6.flatSEnable || page6.launchEnabled)
+  // Only read the shared clutch input when a function using it is enabled.
+  if (page6.flatSEnable || page6.launchEnabled)
   {
-    if(page6.launchHiLo > 0) { current.clutchTrigger = digitalRead(launchPin); }
+    if (page6.launchHiLo > 0) { current.clutchTrigger = digitalRead(launchPin); }
```
adbancroft: Is this equivalent to `current.clutchTrigger = page6.launchHiLo==digitalRead(launchPin);`?

speeduino/src/controllers/launch/launchController.cpp
```diff
@@ -1,54 +1,70 @@
...
-        current.flatShiftingHard = true;
-      }
+  }
+  else if (page6.flatSEnable && current.clutchTrigger
+        && (current.clutchEngagedRPM >= RPM_COARSE.toUser(page6.flatSArm)))
+  {
+    const uint16_t flatRpmLimit = getHardCutRpmLimit(current.clutchEngagedRPM, page2, page15);
+    if (current.RPM > flatRpmLimit)
+    {
+      current.flatShiftingHard = true;
```
adbancroft: Equivalent to `current.flatShiftingHard = current.RPM > flatRpmLimit`?

speeduino/src/controllers/launch/launchController.cpp
```diff
@@ -1,54 +1,70 @@
...
-      if( (page2.hardCutType == HARD_CUT_ROLLING) ) { flatRPMLimit += (page15.rollingProtRPMDelta[0] * 10); } //Add the rolling cut delta if enabled (Delta is a negative value)
-
-      if(current.RPM > flatRPMLimit)
-      {
-        //Flat shift rev limit
-        current.flatShiftingHard = true;
-      }
+  }
+  else if (page6.flatSEnable && current.clutchTrigger
+        && (current.clutchEngagedRPM >= RPM_COARSE.toUser(page6.flatSArm)))
```
adbancroft, line 61: Could the Sonar issue be fixed?

speeduino/src/controllers/launch/launchController.cpp
bahram-azimbeyk: Yes: launchHiLo is a one-bit field and digitalRead returns LOW/HIGH (0/1), so the equality gives the same polarity truth table. Simplified in 95cdef5e, while retaining the enabled-only input read and previous-state/RPM-capture behavior. All 10 launch-control tests pass, including normal and inverted input cases.

speeduino/src/controllers/launch/launchController.cpp
bahram-azimbeyk: Yes. Replaced the conditional assignment with current.flatShiftingHard = (current.RPM > flatRpmLimit) in 95cdef5e. The existing tests exercise below/equal/above limits and clearing an already-active cut at equality for both full and rolling cut; all 10 launch-control tests pass.

speeduino/src/controllers/launch/launchController.cpp
bahram-azimbeyk: The reported rule is cpp:M23_112: the if/else-if chain requires a terminating else. Added a documented final else in 7c41b986 that retains the already-cleared flags when neither function is armed. Local tests and the Mega2560 build pass; the new remote Sonar scan will confirm the finding is cleared. The PR memory table is also updated: RAM unchanged, flash +2 bytes versus upstream (187,950 versus 187,948).

speeduino/src/controllers/launch/launchController.cpp
```diff
@@ -1,54 +1,70 @@
...
   current.flatShiftingHard = false;
 
-  if (page6.launchEnabled && current.clutchTrigger && (current.clutchEngagedRPM < ((unsigned int)(page6.flatSArm) * 100)) && (current.TPS >= page10.lnchCtrlTPS) ) 
-  { 
-    //Only enable if VSS is not used or if it is, make sure we're not above the speed limit
-    if( (page2.vssMode == 0) || ((page2.vssMode > 0) && (current.vss < page10.lnchCtrlVss)) )
+  if (isLaunchArmed(current, page6, page10))
+  {
+    // A configured vehicle speed limit applies only to launch control.
+    if ((page2.vssMode == 0) || (current.vss < page10.lnchCtrlVss))
```
adbancroft, line 50: Could you use `VSS_MODE_OFF` here?

adbancroft: Or extract the boolean check into a separate function? E.g. `withinLaunchSpeedLimit`?

adbancroft: And do the same for the inner if? So it reads `if (withinLaunchSpeedLimit(...) && withinLauncRpmLimit(...))`

bahram-azimbeyk: Updated in 2d2d4492 / b45820d1: the speed predicate uses VSS_MODE_OFF and is extracted as withinLaunchSpeedLimit(). The RPM predicate is extracted as aboveLaunchRpmLimit(), and the inner check now combines them with &&. I used "above" for RPM because a true result means RPM is strictly above the hard-cut threshold, rather than within the permitted range.

The outer launch-arming branch and its else-if flat-shift priority are unchanged. All 10 native launch-control tests pass, including VSS modes/boundaries, RPM equality and rolling-cut thresholds. The Mega2560 build passes with the same footprint as the previous revision: RAM 6,354 B, flash 187,950 B (+2 B versus upstream).

